PR #57, opened 2024-10-01
Implement Hedging (Retry Slow Parts) for Upload

waahm7:
*Description of changes:*
- Implements Hedging (retry slow parts) for upload. Based on my experiments, I didn't see any difference for Download, so this PR only adds it for Upload. We can easily add it to Download stack in future if needed. 

Here are the benchmarks with and without hedging for Upload:
```
60GB Upload with 250 concurrency:
With hedging:
Run:1 Secs:16.674105 Gb/s:30.909969
Run:2 Secs:16.062010 Gb/s:32.087895
Run:3 Secs:16.787110 Gb/s:30.701895
Run:4 Secs:15.197997 Gb/s:33.912106
Run:5 Secs:16.218516 Gb/s:31.778250
Run:6 Secs:16.623848 Gb/s:31.003417
Run:7 Secs:16.910915 Gb/s:30.477126
Run:8 Secs:15.788846 Gb/s:32.643048
Run:9 Secs:15.869860 Gb/s:32.476409
Run:10 Secs:16.757620 Gb/s:30.755923
Without Hedging:
Run:1 Secs:17.601898 Gb/s:29.280711
Run:2 Secs:16.084260 Gb/s:32.043505
Run:3 Secs:17.486960 Gb/s:29.473167
Run:4 Secs:17.730397 Gb/s:29.068502
Run:5 Secs:17.425133 Gb/s:29.577741
Run:6 Secs:16.279045 Gb/s:31.660093
Run:7 Secs:16.883302 Gb/s:30.526971
Run:8 Secs:17.831837 Gb/s:28.903140
Run:9 Secs:17.302441 Gb/s:29.787478
Run:10 Secs:17.429089 Gb/s:29.571028

30 GB with 250 concurrency:
With Hedging:
Run:1 Secs:9.298489 Gb/s:27.713968
Run:2 Secs:9.620810 Gb/s:26.785483
Run:3 Secs:8.474216 Gb/s:30.409661
Run:4 Secs:8.710808 Gb/s:29.583712
Run:5 Secs:8.909990 Gb/s:28.922372
Run:6 Secs:8.086036 Gb/s:31.869516
Run:7 Secs:7.887725 Gb/s:32.670770
Run:8 Secs:9.721999 Gb/s:26.506692
Run:9 Secs:8.053054 Gb/s:32.000038
Run:10 Secs:8.440625 Gb/s:30.530680

Without Hedging:
Run:1 Secs:11.358797 Gb/s:22.687090
Run:2 Secs:10.803229 Gb/s:23.853797
Run:3 Secs:11.495138 Gb/s:22.418003
Run:4 Secs:9.150468 Gb/s:28.162280
Run:5 Secs:10.896210 Gb/s:23.650245
Run:6 Secs:10.624905 Gb/s:24.254150
Run:7 Secs:10.698735 Gb/s:24.086777
Run:8 Secs:9.340982 Gb/s:27.587897
Run:9 Secs:11.398385 Gb/s:22.608294
Run:10 Secs:10.350546 Gb/s:24.897047

```

### DIFF
--- a/aws-s3-transfer-manager/src/middleware/hedge.rs
+++ b/aws-s3-transfer-manager/src/middleware/hedge.rs
@@ -18,41 +18,48 @@ use tower::{
 * Based on our experiments, this makes a significant difference for multipart upload use-cases and
 * does not have a noticeable impact for the Download.
 */
-pub(crate) struct HedgeBuilder<P> {
+
+/// S3 recommends retrying the slowest 5% of the requests.
+const LATENCY_PERCENTILE: f32 = 95.0;
+/// This value was chosen randomly and empirically tested.
+const MIN_DATA_POINTS: u64 = 20;
+/// The Hedge layer maintains two rotating histograms, i.e., ReadHistogram and WriteHistogram. They
+/// are switched every period. This value was chosen randomly while taking into consideration that most 8-16 MB part
+/// requests take on average 0.2 seconds, and we should retry it if it takes more than a second.
+const PERIOD: Duration = Duration::new(2,0);
+
+pub(crate) struct Builder<P> {
     policy: P,
     latency_percentile: f32,
     min_data_points: u64,
     period: Duration,
 }
 
-impl<P> HedgeBuilder<P> {
-    pub(crate) fn new(policy: P) -> Self {
+#[derive(Debug, Clone, Default)]
+pub(crate) struct DefaultPolicy;
+
+impl<T: Clone> Policy<T> for DefaultPolicy {
+    fn clone_request(&self, req: &T) -> Option<T> {
+        Some(req.clone())
+    }
+
+    fn can_retry(&self, _req: &T) -> bool {
+        true
+    }
+}
+
+impl Default for Builder<DefaultPolicy> {
+    fn default() -> Self {
         Self {
-            policy,
-            latency_percentile: 95.0,
-            min_data_points: 20,
-            period: Duration::new(2, 0),
+            policy: DefaultPolicy,
+            latency_percentile: LATENCY_PERCENTILE,
+            min_data_points: MIN_DATA_POINTS,
+            period: PERIOD,
         }
     }
+}
 
-    #[allow(dead_code)]
-    pub(crate) fn with_percentile(mut self, percentile: f32) -> Self {
-        self.latency_percentile = percentile;
-        self
-    }
-
-    #[allow(dead_code)]
-    pub(crate) fn with_min_data_points(mut self, min_data_points: u64) -> Self {
-        self.min_data_points = min_data_points;
-        self
-    }
-
-    #[allow(dead_code)]
-    pub(crate) fn with_duration(mut self, duration: Duration) -> Self {
-        self.period = duration;
-        self
-    }
-
+impl<P> Builder<P> {
     /// Converts the `Hedge` into a `Layer` that can be used in a service stack.
     pub(crate) fn into_layer<Request, S>(self) -> impl Layer<S, Service = Hedge<S, P>> + Clone
     where
--- a/aws-s3-transfer-manager/src/middleware/hedge.rs
+++ b/aws-s3-transfer-manager/src/middleware/hedge.rs
@@ -12,7 +12,7 @@ use tower::{
 
 /// S3 recommends retrying the slowest 5% of the requests.
 const LATENCY_PERCENTILE: f32 = 95.0;
-/// This value was chosen randomly and empirically tested.
+/// This value was chosen randomly, but seems to work.
 const MIN_DATA_POINTS: u64 = 20;
 /// The Hedge layer maintains two rotating histograms, i.e., ReadHistogram and WriteHistogram. They
 /// are switched every period. This value was chosen randomly but with consideration given to the fact that most 8-16 MB part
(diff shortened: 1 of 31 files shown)


aws-s3-transfer-manager/src/middleware/hedge.rs
```diff
@@ -0,0 +1,84 @@
...
+    period: Duration,
+}
+
+#[derive(Debug, Clone, Default)]
+pub(crate) struct DefaultPolicy;
+
+impl<T: Clone> Policy<T> for DefaultPolicy {
+    fn clone_request(&self, req: &T) -> Option<T> {
+        Some(req.clone())
+    }
```
ysaito1001, line 43: Since this is not a public API we can change it later, but here's my two cents.

If an API takes a reference and internally clones it to do something, it's more honest to the callers of the API if the API takes ownership of the passed-in argument (sort of documented [here](https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control)).

Something like
```
    fn clone_request(&self, req: T) -> Option<T> {
        Some(req)
    }
```
At this point, we may or may not need the trait bound `Clone` on `T`.

Curious what you think, though.

waahm7: I don't understand it. The link states that "If a function does not require ownership of an argument, it should take a shared or exclusive borrow of the argument rather than taking ownership and dropping the argument." Since we don't need ownership of `T` here, should we take a reference? 

Also, this trait is not defined by us; it is defined by Tower at https://github.com/tower-rs/tower/blob/master/tower/src/hedge/mod.rs#L58.
```
fn clone_request(&self, req: &Request) -> Option<Request>;
```
So we are getting a reference to Request instead of the ownership.

waahm7: I think what this API is doing is that since the first request is taking too long, we need to send another one. So we clone the request and send the second request. Both requests will stay alive until one of them completes, so we can't drop the original request yet.

ysaito1001: > this trait is not defined by us

Oh my bad, I don't know where I was looking at, and this is specifically a method for `clone` (for some reason I didn't pay attention to the method name).

The link applies to many cases, but `clone_request` is an exception because of what it does.

aws-s3-transfer-manager/src/middleware/hedge.rs
```diff
@@ -0,0 +1,84 @@
...
+};
+
+/// S3 recommends retrying the slowest 5% of the requests.
+const LATENCY_PERCENTILE: f32 = 95.0;
+/// This value was chosen randomly and empirically tested.
+const MIN_DATA_POINTS: u64 = 20;
+/// The Hedge layer maintains two rotating histograms, i.e., ReadHistogram and WriteHistogram. They
+/// are switched every period. This value was chosen randomly but with consideration given to the fact that most 8-16 MB part
+/// requests take on average 0.2 seconds, and we should retry it if it takes more than a second.
+const PERIOD: Duration = Duration::new(2, 0);
```
graebm, line 20: utterly trivial: `from_secs()` is a bit more explicit than `new()`

aws-s3-transfer-manager/src/middleware/hedge.rs
```diff
@@ -0,0 +1,84 @@
...
+use std::time::Duration;
+use tower::{
+    hedge::{Hedge, Policy},
+    layer::layer_fn,
+    BoxError, Layer, Service,
+};
+
+/// S3 recommends retrying the slowest 5% of the requests.
+const LATENCY_PERCENTILE: f32 = 95.0;
+/// This value was chosen randomly and empirically tested.
```
graebm, line 15: utterly trivial: "empirically tested" sounds like there is actually some science behind it
```suggestion
/// This value was chosen randomly, but seems to work.
```

aws-s3-transfer-manager/src/operation/upload/service.rs
```diff
@@ -69,7 +70,19 @@ pub(super) fn upload_part_service(
        + Send {
     let svc = service_fn(upload_part_handler);
     let concurrency_limit = ConcurrencyLimitLayer::new(ctx.handle.scheduler.clone());
-    ServiceBuilder::new().layer(concurrency_limit).service(svc)
+
+    let svc = ServiceBuilder::new()
+        .layer(concurrency_limit)
```
graebm, line 75: A "hedged" request should count towards to concurrency limit. If this is hard to do, let's do it in a followup task.

Even better would be if hedged requests were lower priority, and only got a ticket when there's no other work to do. Since they really only help us if we're done with a workload and all that's left is the slow requests. Otherwise they're just stealing from another request that would be more useful.

waahm7: I have looked into this, and it seems like the hedged requests can bypass the concurrency_limit layer. Reorganizing the layers had no effect. I have added a TODO to fix this for now since I want to get this merged and see if there is any difference in benchmarks.

Another interesting thing I noticed is that we are hedging very few requests, like for a 30GB upload with 125 concurrency, we only hedged less 5-15 requests. We can revisit the numbers once we have more tracing/metrics in place.

graebm (approved):

Issue #60, opened 2024-10-04
Improve Upload Memory Usage


waahm7:
We can blow our memory during upload since we don't hook our read_body phase with the scheduler. We need to improve our scheduler so that we acquire a permit before reading the body and maybe release it after uploading, or have different types of permits.

PR #61, opened 2024-10-04
Improve Download Memory Usage

waahm7:
*Description of changes:*
`Tokio.spawn` doesn't respect the spawn order, which can result in us downloading the first `num_concurrency` parts in random order. For a workload of 5GB * 100 files, this can lead to very high memory usage, as seen in the diagram below. This PR refactors the exact part to be determined only once the task has been scheduled.

Uploads can also have a similar issue where we read too many parts into memory. To fix that, we will need to refactor our scheduler to be smarter so that we only read the part when we have the permit. (Created: https://github.com/awslabs/aws-s3-transfer-manager-rs/issues/60)

memory_usage_comparison

### DIFF
--- a/aws-s3-transfer-manager/src/operation/download.rs
+++ b/aws-s3-transfer-manager/src/operation/download.rs
@@ -28,6 +28,7 @@ use aws_smithy_types::byte_stream::ByteStream;
 use body::Body;
 use discovery::discover_obj;
 use service::{distribute_work, ChunkResponse};
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 use tokio::sync::mpsc;
 use tokio::task::JoinSet;
@@ -96,7 +97,6 @@ fn handle_discovery_chunk(
     completed: &mpsc::Sender<Result<ChunkResponse, crate::error::Error>>,
     permit: OwnedWorkPermit,
 ) -> u64 {
-    let mut start_seq = 0;
     if let Some(stream) = initial_chunk {
         let seq = handle.ctx.next_seq();
         let completed = completed.clone();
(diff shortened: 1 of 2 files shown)


aws-s3-transfer-manager/src/operation/download.rs
```diff
@@ -129,18 +129,30 @@ fn handle_discovery_chunk(
 
 /// Download operation specific state
 #[derive(Debug)]
-pub(crate) struct DownloadState {}
+pub(crate) struct DownloadState {
+    current_seq: u64,
```
aajtodd, line 133: Can you not get away with an atomic here?

aws-s3-transfer-manager/src/operation/download.rs
```diff
@@ -97,8 +97,8 @@ fn handle_discovery_chunk(
     permit: OwnedWorkPermit,
 ) -> u64 {
     let mut start_seq = 0;
-
     if let Some(stream) = initial_chunk {
+        let seq = handle.ctx.next_seq();
```
aajtodd, line 101: `start_seq` and `seq` aren't connected anywhere here, should probably set `start_seq` to `seq` instead of hard coded to `1`

aajtodd (commented):
What's this do to throughput for download? In particular the warmup/first few runs?

aws-s3-transfer-manager/src/operation/download.rs
waahm7: Thanks, updated to AtomicU64.

aws-s3-transfer-manager/src/operation/download.rs
waahm7: Thanks, we would have to set start_seq = seq+1. I have added a current_seq() function which will return 0 or 1 depending upon we called next_seq or not.

waahm7:
> What's this do to throughput for download? In particular the warmup/first few runs?
It didn’t help much since we were only doing 125 out of 3840 parts out of order.

aajtodd (approved):

ysaito1001 (approved):

PR #66, opened 2024-10-18
Curate tracing spans

graebm:
*Description of changes:*
- Each major operation has a "parent" span it assigns to the tasks it spawns. Otherwise, spans from spawned tasks have no parent, which makes them a disorganized mess in visualizations.
- Ensure there's a span for each HTTP request, because these are often the bottleneck.
- Ensure spans have important key/value pairs (e.g. seq, part_number, bucket, key), so when a bunch of spans with the same name are together, we tell what each one is up to.
- **debatable:** gave names that stand on their own. For example: the function `aws_s3_transfer_manager::operation::upload::builders::send()` is named "initiate-download", instead of just "send", so a visualization doesn't need to show the full (very very wide) namespace to explain what's going on.

### DIFF
--- a/aws-s3-transfer-manager/src/operation/download.rs
+++ b/aws-s3-transfer-manager/src/operation/download.rs
@@ -81,16 +81,28 @@ impl Download {
             // spawn all work into the same JoinSet such that when the set is dropped all tasks are cancelled.
             tasks: JoinSet::new(),
             ctx,
-            parent_span_for_tasks,
         };
 
         // spawn a task (if necessary) to handle the discovery chunk. This returns immediately so
         // that we can begin concurrently downloading any reamining chunks/parts ASAP
-        let start_seq = handle_discovery_chunk(&mut handle, initial_chunk, &comp_tx, permit);
+        let start_seq = handle_discovery_chunk(
+            &mut handle,
+            initial_chunk,
+            &comp_tx,
+            permit,
+            parent_span_for_tasks.clone(),
+        );
 
         if !discovery.remaining.is_empty() {
             let remaining = discovery.remaining.clone();
-            distribute_work(&mut handle, remaining, input, start_seq, comp_tx)
+            distribute_work(
+                &mut handle,
+                remaining,
+                input,
+                start_seq,
+                comp_tx,
+                parent_span_for_tasks,
+            )
         }
 
         Ok(handle)
@@ -109,6 +121,7 @@ fn handle_discovery_chunk(
     initial_chunk: Option<ByteStream>,
     completed: &mpsc::Sender<Result<ChunkResponse, crate::error::Error>>,
     permit: OwnedWorkPermit,
+    parent_span_for_tasks: tracing::Span,
 ) -> u64 {
     if let Some(stream) = initial_chunk {
         let seq = handle.ctx.next_seq();
@@ -134,7 +147,7 @@ fn handle_discovery_chunk(
                     &DisplayErrorContext(send_err)
                 );
             }
-        }.instrument(tracing::debug_span!(parent: handle.parent_span_for_tasks.clone(), "collect-body-from-discovery", seq)));
+        }.instrument(tracing::debug_span!(parent: parent_span_for_tasks.clone(), "collect-body-from-discovery", seq)));
     }
     handle.ctx.current_seq()
 }
--- a/aws-s3-transfer-manager/src/operation/download.rs
+++ b/aws-s3-transfer-manager/src/operation/download.rs
@@ -42,10 +42,20 @@ pub(crate) struct Download;
 
 impl Download {
     /// Execute a single `Download` transfer operation
+    ///
+    /// If `use_current_span_as_parent_for_tasks` is true, spawned tasks will
+    /// be under the current span, and that span will not end until all tasks
+    /// complete. Useful for download-objects, which calls this from a task
+    /// whose span should last until all individual object downloads complete.
+    ///
+    /// If `use_current_span_as_parent_for_tasks` is false, spawned tasks will
+    /// "follow from" the current span, but be under their own root of the trace tree.
+    /// Use this for `TransferManager.download().send()`, where the spawned tasks
+    /// should NOT extend the life of the current `send()` span.
     pub(crate) async fn orchestrate(
         handle: Arc<crate::client::Handle>,
         input: crate::operation::download::DownloadInput,
-        existing_span_for_tasks: Option<tracing::span::Id>,
+        use_current_span_as_parent_for_tasks: bool,
     ) -> Result<DownloadHandle, error::Error> {
         // if there is a part number then just send the default request
         if input.part_number().is_some() {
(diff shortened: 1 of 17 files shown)


aws-s3-transfer-manager/src/operation/download.rs
```diff
@@ -44,6 +45,7 @@ impl Download {
     pub(crate) async fn orchestrate(
         handle: Arc<crate::client::Handle>,
         input: crate::operation::download::DownloadInput,
+        existing_span_for_tasks: Option<tracing::span::Id>,
```
aajtodd, line 48: Why do we need to pass this in? Span relationships should generally be automatic.

graebm: It's not automatic for spawned tasks, which start with no parent span. Previously, the "download-chunk" spans had no parent. If your main thread did 2 downloads, where each download had 2 chunks, the span tree would look like:
```
- main
  - initiate-download
  - join-download
  - initiate-download
  - join-download
- download-chunk
- download-chunk
- download-chunk 
- download-chunk <-- which "download operation" do these chunks belong to?
```

So I started creating "download-tasks" span that's used to group a download's spawned tasks together. So now it looks like:
```
- main
  - initiate-download
  - join-download
  - initiate-download
  - join-download
- download-tasks <-- yay a common parent per download operation
  - download-chunk
  - download-chunk
- download-tasks
  - download-chunk
  - download-chunk
```

But `download-objects` is going to do many downloads, which will lead to many different "download-tasks" spans, and I thought it would be nice to group them under a common parent.under another span. So when `download-objects` calls this, it passes a span named "download-objects-tasks" and the span tree looks like:
```
- main
  - initiate-download-objects
  - join-download-objects
- download-objects-tasks <-- 
  - download-tasks
    - download-chunk
    - download-chunk
  - download-tasks
    - download-chunk
    - download-chunk
```

aws-s3-transfer-manager/src/operation/download/service.rs
```diff
@@ -64,7 +75,7 @@ async fn download_chunk_handler(
 
     let bytes = body
         .collect()
-        .instrument(tracing::debug_span!("collect-body", seq = seq))
+        .instrument(tracing::debug_span!("download-chunk-collect-body", seq))
```
waahm7, line 78: We are not instrumenting the send-download-chunk few lines above?

aws-s3-transfer-manager/src/operation/upload/handle.rs
```diff
@@ -12,15 +12,20 @@ use aws_sdk_s3::error::DisplayErrorContext;
...
 #[non_exhaustive]
 pub struct UploadHandle {
     /// All child multipart upload tasks spawned for this upload
     pub(crate) upload_tasks: Arc<Mutex<task::JoinSet<Result<CompletedPart, crate::error::Error>>>>,
+    /// Parent tracing span for spawned upload tasks.
+    pub(crate) parent_span_for_upload_tasks: tracing::Span,
     /// All child read body tasks spawned for this upload
     pub(crate) read_tasks: task::JoinSet<Result<(), crate::error::Error>>,
+    /// Parent tracing span for spawned read body tasks.
+    pub(crate) parent_span_for_read_tasks: tracing::Span,
```
waahm7, line 28: Do we need to store these within the handle? Can we just create them within the distribute_work function and use them there?

waahm7: Similar for download, instead of storing it in the handle, can we just use it in the distribute_work function?

aws-s3-transfer-manager/src/operation/download/service.rs
graebm: I thought it was more useful to wrap this function in a parent span, so send + collect are grouped together

The HTTP requests where send() doesn't do all the work are weird ... spanwise

aws-s3-transfer-manager/src/operation/upload/handle.rs
graebm: oh hey, you're right

aws-s3-transfer-manager/src/operation/download.rs
aajtodd: Shouldn't calling `.instrument()` on the future passed to `spawn()` be sufficient?

aws-s3-transfer-manager/src/operation/download_objects.rs
```diff
@@ -43,6 +43,16 @@ impl DownloadObjects {
         let destination = input.destination().expect("destination set");
         validate_destination(destination).await?;
 
+        // create span to serve as parent of spawned child tasks
+        let parent_span_for_tasks = tracing::debug_span!(
+            parent: None,
```
aajtodd, line 48: Don't we want the parent to be whatever the current span is? Whatever span (if any) is in the current context  when calling `orchestrate` is the parent span. 

If the parent child relationship is correct we shouldn't need to add the `follows_from` for this span (also `[follows_from](https://docs.rs/tracing/latest/tracing/struct.Span.html#method.follows_from)` implies a indirect causal relationship, is that correct? Seems like a normal parent/child relationship would be appropriate. See [docs for example scenario](https://github.com/tokio-rs/tracing/blob/tracing-0.1.40/tracing/src/span.rs#L177-L193)). I could see possibly the upload worker tasks using this though depending.

aws-s3-transfer-manager/src/operation/download_objects/worker.rs
```diff
@@ -142,8 +142,12 @@ async fn download_single_obj(
     let delim = ctx.state.input.delimiter();
 
     let key_path = local_key_path(root_dir, key.as_str(), prefix, delim)?;
-    let mut handle =
-        crate::operation::download::Download::orchestrate(ctx.handle.clone(), input).await?;
+    let mut handle = crate::operation::download::Download::orchestrate(
+        ctx.handle.clone(),
+        input,
+        tracing::Span::current().id(),
```
aajtodd, line 148: I don't think you need this, the current span is always used as the parent by default. I think you just need to call `.instrument()` on this future before calling `.await`. Or in this case if you want the parent to be the parent you can use [`in_current_span()`](https://docs.rs/tracing-futures/0.2.5/tracing_futures/trait.Instrument.html#method.in_current_span) for that.

```rs
    let mut handle = crate::operation::download::Download::orchestrate(
        ctx.handle.clone(),
        input)
     .in_current_span()
     .await?;
```

aajtodd (commented):
It's totally possible I'm wrong but I _don't_ think we need to be doing as much passing around of spans/span ID's. A trace is usually a simple tree like structure with parent/child relationships that are usually automatically established by the structure of the code. async/futures takes slightly more care to ensure it's setup properly but it should still generally be automatic. 

Happy to discuss offline and do a deeper dive to either see why it has to be this way or to try to figure out a way that doesn't require manually tracking parents. 

In particular setting parent to `None` and establishing a `follows_from` relationship may break customer expectations when looking at traces from their own code.

e.g.

```rs
// src/main.rs

async fn customer_app(tm: aws_s3_transfer_manager::Client) -> Result<()> {
     let span = tracing::debug_span!("customer parent span");
     // all transfer manager download spans from this operation should generally be 
     // a child of the customer span
     let handle = tm.download()
         .bucket(...)
         .key(...)
         .send()
         .instrument(span)
         .await?;
}

```

aws-s3-transfer-manager/src/operation/download_objects.rs
graebm: A span doesn't "end" until all its children end. So if the spawned-tasks are children of the initial `tm.download().send()` function, it would make that `send()` span look waaaaay longer than the length of time `send().await` actually takes.

If API was just a single async `send()` function, I would make the spawned tasks be its children. But since our API requires the user to call `send()`, then `join()` / `collect()`, it's not clear which span the spawned-tasks should be children of.

aws-s3-transfer-manager/src/operation/download_objects/worker.rs
graebm: You're right that I don't need to be passing the current span's ID. I changed this argument to a bool: `use_current_span_as_parent_for_tasks`

But we talked. I do still want something to control whether the tasks for single-object-download go under the current span, or becomes a new root span. I added some detailed comments about this...

aajtodd (approved):
Only thing we may want to consider at some point is an integration test where we capture the logs and verify the span hierarchy is expected and we don't regress at some point (easy to lose span context in futures).

Issue #68, opened 2024-10-30
Make return types consistent: use `Option<&str>` instead of `&Option<String>`


ysaito1001:
For struct getters that return `&Option<String>` throughout the codebase, replace them with `Option<&str>`.

There are structs in the AWS SDK crates that return `&Option<string>` but this is due to the `smithy-rs` code generator mechanically determining the return type as `&Option<T>` given a struct field of type `T`. If we can afford to handwrite the Rust code, we should prefer more idiomatic `Option<&str>`.

ysaito1001:
Implemented in #74

PR #74, opened 2024-11-11
Address `TODO`s to handle chores

ysaito1001:
*Issue #, if available:*
- https://github.com/awslabs/aws-s3-transfer-manager-rs/issues/68
- Make return types use `Option<&Path>` instead of `&Option<PathBuf>`, similar to the PR above
- Address this rename [TODO](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/c88e07f7c92a3fe8bd391c624876538743c5314b/aws-s3-transfer-manager/src/types.rs#L225-L226)


aajtodd (approved):

graebm (commented):
👍

Issue #65, opened 2024-10-17
Figure out `join()` and `Drop` behavior for `DownloadHandle`


graebm:
Currently, DownloadHandle has a [join()](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/3dd20c8aa0872352100cf456beee02bfc53c73d1/aws-s3-transfer-manager/src/operation/download/handle.rs#L45) function, but neither [cp.rs](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/3dd20c8aa0872352100cf456beee02bfc53c73d1/aws-s3-transfer-manager/examples/cp.rs#L174) or the [benchmark runner](https://github.com/awslabs/aws-crt-s3-benchmarks/blob/71be44d88dff5975fca2caa8a838bc0e6c45cbb7/runners/s3-benchrunner-rust/src/transfer_manager.rs#L94) bother to call it. It was added because `UploadHandle` has a [join()](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/3dd20c8aa0872352100cf456beee02bfc53c73d1/aws-s3-transfer-manager/src/operation/upload/handle.rs#L55), and we thought the APIs should be similar.

But `DownloadHandle::join()` cancels the download, while `UploadHandle::join()` waits for the upload to successfully complete, and has a separate `UploadHandle::abort()` function. Should we rename ~DownloadHandle::join()~ to `DownloadHandle::abort()`?

Also, what's the `Drop` behavior for `DownloadHandle`? We should probably ensure its spawned tasks get cancelled ASAP. Likewise for `UploadHandle/UploadObjectsHandle/DownloadObjectsHandle`

- [x] Drop and cancel for `UploadHandle/UploadObjectsHandle` ([PR](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/75))
- [ ] Drop and cancel for `DownloadHandle` ([PR](https://docs.rs/tokio/latest/tokio/task/struct.JoinError.html#method.is_cancelled))
- [x] Drop and cancel for `DownloadObjectsHandle` ([PR](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/76))

ysaito1001:
This issue can be closed once all subtasks have been completed. As a result of the subtasks, here is a summary:
--- | Join | Abort | Drop
--- | --- | --- | ---
[...]

PR #75, opened 2024-11-18
Add mechanism to cancel the operation of uploading directory

ysaito1001:
*Issue #, if available:*
Resolves a subtask in #65 

*Description of changes:*
This PR introduces a mechanism to cancel the upload of multiple objects within a directory. It distinguishes between aborting and dropping a handle for both `UploadHandle` and `UploadObjectsHandle`. Note that this PR does not modify the `UploadHandle` functionality, except for adding some comments to the struct.

For both of the handles, the basic idea is
- Dropping a handle abruptly terminates in-progress tasks. This occurs either when the handle is destroyed via [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) or when `.drop()` is called explicitly. The tasks will be canceled at the points where their futures are waiting, and we don't have much control over the cleanup process, which is left to the underlying async runtime.
- Aborting a handle (via calling `.abort()` on it) allows tasks to complete their current work while ignoring any future tasks. The word "complete" has different meanings depending on whether the handle is an `UploadHandle` or an `UploadObjectsHandle`.
  - For an `UploadHandle`, it means that the control flow attempts to abort ongoing read and upload tasks, and it will invoke `AbortMultipartUpload` in the case of MPU.
  - For an `UploadObjectsHandle`, it means that each upload task either finishes uploading the current object or invokes `.abort()` on its respective single upload operation down the line.

For `UploadObjectsHandle`, this PR introduces a watch channel that signals when the operation enters cancellation mode and needs to be terminated gracefully. There are two ways to trigger cancellation:
- A user explicitly calls `.abort()` on the `UploadObjectsHandle`.
- A child operation fails while running with the abort policy.

Both cases are covered by newly added integration tests.

**Design Discussions:**
- I once considered making the behavior of `Drop` the same as calling `.abort()`—for example, having fn `drop(&mut self)` call `.abort()`. However, since `.drop` is a synchronous method and `.abort()` is asynchronous, this led me to implement separate behaviors for the two.
- Once the operation enters cancellation mode, `.abort()` (for both `UploadHandle` and `UploadObjectsHandle`) will only return the error that triggered the cancellation and will log any additional errors encountered along the way (e.g., errors from AbortMultipartUpload). We have [TODO](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/main/aws-s3-transfer-manager/src/operation/upload/handle.rs#L163) and I also left [TODO](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/55f8ce77f50467066dfacad22cee629952a83c18/aws-s3-transfer-manager/src/operation/upload_objects/handle.rs#L55) to assess this behavior, but for now, the code after this PR "swallows" and logs any new errors encountered during the cancellation process.

### DIFF
--- a/aws-s3-transfer-manager/src/error.rs
+++ b/aws-s3-transfer-manager/src/error.rs
@@ -44,6 +44,11 @@ pub enum ErrorKind {
 
     /// child operation failed (e.g. download of a single object as part of downloading all objects from a bucket)
     ChildOperationFailed,
+
+    /// The operation is being canceled because the user explicitly called `.abort` on the handle,
+    /// or a child operation failed with the abort policy. The operation is now transitioning into
+    /// a graceful shutdown mode.
+    OperationCancelled,
 }
 
 impl Error {
@@ -75,6 +80,7 @@ impl fmt::Display for Error {
             ErrorKind::ChunkFailed => write!(f, "failed to process chunk"),
             ErrorKind::NotFound => write!(f, "resource not found"),
             ErrorKind::ChildOperationFailed => write!(f, "child operation failed"),
+            ErrorKind::OperationCancelled => write!(f, "operation cancelled"),
         }
     }
 }
--- a/aws-s3-transfer-manager/src/error.rs
+++ b/aws-s3-transfer-manager/src/error.rs
@@ -46,8 +46,7 @@ pub enum ErrorKind {
     ChildOperationFailed,
 
     /// The operation is being canceled because the user explicitly called `.abort` on the handle,
-    /// or a child operation failed with the abort policy. The operation is now transitioning into
-    /// a graceful shutdown mode.
+    /// or a child operation failed with the abort policy.
     OperationCancelled,
 }
 
--- a/aws-s3-transfer-manager/src/operation/upload_objects.rs
+++ b/aws-s3-transfer-manager/src/operation/upload_objects.rs
@@ -16,7 +16,10 @@ pub use handle::UploadObjectsHandle;
 
 mod output;
 pub use output::{UploadObjectsOutput, UploadObjectsOutputBuilder};
-use tokio::task::JoinSet;
+use tokio::{
+    sync::watch::{self, Sender},
+    task::JoinSet,
+};
 use tracing::Instrument;
 
 mod worker;
@@ -52,9 +55,9 @@ impl UploadObjects {
                 return Err(e);
             }
         };
-
+        let (cancel_tx, cancel_rx) = watch::channel(false);
         let concurrency = handle.num_workers();
-        let ctx = UploadObjectsContext::new(handle.clone(), input);
+        let ctx = UploadObjectsContext::new(handle.clone(), input, cancel_tx);
 
         // spawn all work into the same JoinSet such that when the set is dropped all tasks are cancelled.
         let mut tasks = JoinSet::new();
@@ -62,13 +65,15 @@ impl UploadObjects {
 
         // spawn worker to discover/distribute work
         tasks.spawn(worker::list_directory_contents(
-            ctx.state.input.clone(),
+            ctx.state.clone(),
             list_directory_tx,
+            cancel_rx.clone(),
         ));
 
         for i in 0..concurrency {
-            let worker = worker::upload_objects(ctx.clone(), list_directory_rx.clone())
-                .instrument(tracing::debug_span!("object-uploader", worker = i));
+            let worker =
+                worker::upload_objects(ctx.clone(), list_directory_rx.clone(), cancel_rx.clone())
+                    .instrument(tracing::debug_span!("object-uploader", worker = i));
             tasks.spawn(worker);
         }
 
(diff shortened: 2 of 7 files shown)


aws-s3-transfer-manager/tests/upload_objects_test.rs
```diff
@@ -401,3 +407,105 @@ async fn test_error_when_custom_delimiter_appears_in_filename() {
     assert!(format!("{}", DisplayErrorContext(err))
         .contains("a custom delimiter `-` should not appear"));
 }
+
+#[tokio::test]
+async fn test_abort_on_handle_should_terminate_tasks_gracefully() {
```
aajtodd, line 412: do we need a test for `drop(handle)`?

aws-s3-transfer-manager/src/operation/upload_objects/worker.rs
```diff
@@ -163,42 +187,56 @@ fn derive_object_key<'a>(
...
     }
 
-    tracing::trace!("req channel closed, worker finished");
     Ok(())
 }
 
 async fn upload_single_obj(
     ctx: &UploadObjectsContext,
     job: UploadObjectJob,
+    cancel_rx: watch::Receiver<bool>,
```
aajtodd, line 239: We're passing around `cancel_rx` everywhere, seems like it belongs in the context which is meant to hold the state for the operation. I realize we are storing the `tx` side of this in the handle so maybe that causes issues but if it doesn't I think it would be cleaner to just thread this into the context so that we don't have to have a separate parameter everywhere.

aws-s3-transfer-manager/src/operation/upload_objects/worker.rs
```diff
@@ -217,11 +255,19 @@ async fn upload_single_obj(
...
         .expect("valid input");
 
-    let handle = crate::operation::upload::Upload::orchestrate(ctx.handle.clone(), input).await?;
-
-    handle.join().await?;
+    let mut handle =
+        crate::operation::upload::Upload::orchestrate(ctx.handle.clone(), input).await?;
 
-    Ok(bytes_transferred)
+    if cancel_rx.has_changed().unwrap() {
```
aajtodd, line 261: Can we wrap this into a utility future like `MaybeCancel::new(fut)`? that checks cancelation on each poll? Feels like it would be useful.

aws-s3-transfer-manager/src/error.rs
```diff
@@ -44,6 +44,11 @@ pub enum ErrorKind {
 
     /// child operation failed (e.g. download of a single object as part of downloading all objects from a bucket)
     ChildOperationFailed,
+
+    /// The operation is being canceled because the user explicitly called `.abort` on the handle,
+    /// or a child operation failed with the abort policy. The operation is now transitioning into
```
aajtodd, line 49: this wording ("the operation is now transition into...") implies the user may still need to do something.

aajtodd (commented):
Overall looks good, couple questions/comments.

aws-s3-transfer-manager/tests/upload_objects_test.rs
ysaito1001: Added in 6c85ac5

aws-s3-transfer-manager/src/error.rs
ysaito1001: Ah got it, removed the sentence in a6fbdcf

aws-s3-transfer-manager/src/operation/upload_objects/worker.rs
ysaito1001: This particular `if` statement is technically optional (note `has_changed` is a sync method), and things would work just fine without it. It's just here so that we can potentially save a single upload operation that would otherwise be wasted if the world would already be in graceful shutdown mode. So a utility future may not be as useful as you might think (or I guess that depends on what utility future you were think of)?

aajtodd: Fair enough, we can wait and abstract later as use cases arise (if they arise).

aws-s3-transfer-manager/src/operation/upload_objects.rs
```diff
@@ -57,7 +57,7 @@ impl UploadObjects {
         };
         let (cancel_tx, cancel_rx) = watch::channel(false);
         let concurrency = handle.num_workers();
-        let ctx = UploadObjectsContext::new(handle.clone(), input, cancel_tx);
+        let ctx = UploadObjectsContext::new(handle.clone(), input, cancel_tx, cancel_rx);
```
aajtodd, line 60: **nit**: Since you pass both in you could just have `new()` create the watch channel for you.

aajtodd (approved):

aws-s3-transfer-manager/src/operation/upload_objects.rs
ysaito1001: Yeah, I usually prefer dependency injection but in this case, creating receiver ends internally in `UploadObjectsContext` feels the right thing to do (i.e., if a user passed two unrelated receiver ends to `new`, then cancellation mechanism wouldn't work).

aws-s3-transfer-manager/src/operation/upload_objects/worker.rs
ysaito1001: Yes. For now, at least I added a comment in e655aeb based on my response above.

aws-s3-transfer-manager/src/operation/upload_objects/handle.rs
```diff
@@ -18,13 +42,61 @@ pub struct UploadObjectsHandle {
 
 impl UploadObjectsHandle {
     /// Consume the handle and wait for the upload to complete
+    ///
+    /// When the `FailedTransferPolicy` is set to [`FailedTransferPolicy::Abort`], this method
+    /// will return an error if any of the spawned tasks encounter one. The other tasks will
```
waahm7, line 47:
```suggestion
    /// will return the first error if any of the spawned tasks encounter one. The other tasks will

File: aws-s3-transfer-manager/src/operation/upload_objects/worker.rs [hunk 7]
```diff
@@ -617,4 +703,35 @@ mod unit {
             );
         }
     }
+
+    #[tokio::test]
```
waahm7, line 707: I think we should add a test that tries to do MPU but fails and calls abortMultipart instead of completeMultipart.

ysaito1001: Good point. As discussed offline, a good place to add that test is probably [an integration test file for uploading single object](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/main/aws-s3-transfer-manager/tests/upload_test.rs).

PR #76, opened 2024-11-27
Add mechanism to cancel the operation of downloading objects

ysaito1001:
*Issue #, if available:*
Resolves the third subtask in #65 

*Description of changes:*
The structure of these code changes mirrors that of #75, so most of the modifications should be expected.
This PR does NOT update the cancellation of single object download, and that's being worked on by a separate task.

### DIFF
--- a/aws-s3-transfer-manager/src/operation/download_objects/handle.rs
+++ b/aws-s3-transfer-manager/src/operation/download_objects/handle.rs
@@ -28,7 +28,7 @@ impl DownloadObjectsHandle {
     /// they will be logged as errors, instead.
     ///
     /// If the `FailedTransferPolicy` is set to [`FailedTransferPolicy::Continue`], the
-    /// [`DownloadObjectsOutput`] will include a detailed breakdown, such as the number of
+    /// [`DownloadObjectsOutput`] will include a detailed breakdown, including the number of
     /// successful downloads and the number of failed ones.
     ///
     // TODO(aws-sdk-rust#1159) - Consider if we want to return other all errors encountered during cancellation.
--- a/aws-s3-transfer-manager/src/operation/download_objects/handle.rs
+++ b/aws-s3-transfer-manager/src/operation/download_objects/handle.rs
@@ -65,7 +65,7 @@ impl DownloadObjectsHandle {
     pub async fn abort(&mut self) -> Result<(), crate::error::Error> {
         if self.ctx.state.input.failure_policy() == &FailedTransferPolicy::Abort {
             if self.ctx.state.cancel_tx.send(true).is_err() {
-                tracing::warn!(
+                tracing::debug!(
                     "all receiver ends have been dropped, unable to send a cancellation signal"
                 );
             }
--- a/aws-s3-transfer-manager/src/operation/download_objects/handle.rs
+++ b/aws-s3-transfer-manager/src/operation/download_objects/handle.rs
@@ -5,6 +5,8 @@
 
 use tokio::task;
 
+use crate::{error::ErrorKind, types::FailedTransferPolicy};
+
 use super::{DownloadObjectsContext, DownloadObjectsOutput};
 
 /// Handle for `DownloadObjects` transfer operation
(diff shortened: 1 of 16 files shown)


waahm7 (approved):

aws-s3-transfer-manager/src/operation/download_objects/handle.rs
```diff
@@ -19,14 +21,57 @@ pub struct DownloadObjectsHandle {
...
 impl DownloadObjectsHandle {
     /// Consume the handle and wait for download transfer to complete
+    ///
+    /// When the `FailedTransferPolicy` is set to [`FailedTransferPolicy::Abort`], this method
+    /// will return the first error if any of the spawned tasks encounter one. The other tasks
+    /// will be canceled, but their cancellations will not be reported as errors by this method;
+    /// they will be logged as errors, instead.
+    ///
+    /// If the `FailedTransferPolicy` is set to [`FailedTransferPolicy::Continue`], the
+    /// [`DownloadObjectsOutput`] will include a detailed breakdown, such as the number of
```
aajtodd, line 31: such as -> including

aws-s3-transfer-manager/src/operation/download_objects/handle.rs
```diff
@@ -19,14 +21,57 @@ pub struct DownloadObjectsHandle {
...
+        } else {
+            Ok(DownloadObjectsOutput::from(self.ctx.state.as_ref()))
+        }
+    }
+
+    /// Aborts all tasks owned by the handle.
+    pub async fn abort(&mut self) -> Result<(), crate::error::Error> {
+        if self.ctx.state.input.failure_policy() == &FailedTransferPolicy::Abort {
+            if self.ctx.state.cancel_tx.send(true).is_err() {
+                tracing::warn!(
```
aajtodd, line 68: should this be a warning? or is this more debug information?

aws-s3-transfer-manager/tests/download_objects_test.rs
```diff
@@ -403,3 +422,168 @@ async fn test_destination_dir_not_valid() {
...
+async fn test_abort_on_handle_should_terminate_tasks_gracefully() {
+    let (_guard, rx) = capture_test_logs();
+
+    let bucket = MockBucket::builder()
+        .key_with_size("key1", 12)
+        .key_with_error("key2")
+        .key_with_size("key3", 7)
+        .build();
+
+    let client = mock_client!(aws_sdk_s3, RuleMode::MatchAny, bucket.rules().as_slice());
```
aajtodd, line 436: Don't forget we need to configure an infallible HTTP client anywhere we are using mock client (and likely need to address this upstream but until then...)

aajtodd (approved):
Meant to approve it with some minor feedback/suggestions.

aws-s3-transfer-manager/src/operation/download_objects/handle.rs
```diff
@@ -19,14 +21,57 @@ pub struct DownloadObjectsHandle {
...
+            if let Err(e) = result {
+                match self.ctx.state.input.failure_policy() {
+                    FailedTransferPolicy::Abort
+                        if first_error_to_report.is_none()
+                            && e.kind() != &ErrorKind::OperationCancelled =>
+                    {
+                        first_error_to_report = Some(e);
+                    }
+                    FailedTransferPolicy::Continue => {
+                        tracing::warn!("encountered but dismissed error when the failure policy is `Continue`: {e}")
```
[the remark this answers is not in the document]
ysaito1001: Looks like we already do to some degree, like [here](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/06c087a5d53676bb048f6c512b8eb1fda63f03d5/aws-s3-transfer-manager/src/operation/download_objects/worker.rs#L102-L106) and [here](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/06c087a5d53676bb048f6c512b8eb1fda63f03d5/aws-s3-transfer-manager/src/operation/upload_objects/worker.rs#L218) regardless of the failure policy. The above location at line 50 is a bit of an edge case because with `Continue`, any error is supposed to be dismissed and the execution continues, so by the time the execution flow reaches line 50, we probably do not have access to an error instance that has useful info like the object key.

That said, maybe you could point me to an example location that could benefit additional logging. Happy to add it in a subsequent PR.

aws-s3-transfer-manager/src/operation/download_objects/handle.rs
ysaito1001: Updated in 8e041c3

aws-s3-transfer-manager/src/operation/download_objects/handle.rs
ysaito1001: Given the low likelihood of `is_err` being true (I believe we can almost use `send(true).expect(...)`), `debug` seems to make more sense. Updated in afb09ba.

aws-s3-transfer-manager/tests/download_objects_test.rs
ysaito1001: Yeah, good point. I created a thin wrapper macro in `test_common` and updated usages of `mock_client` throughout the codebase in d0159a3 and e78ecd1.

PR #78, opened 2024-11-29
Follow up on PR#75

ysaito1001:
*Description of changes:*
Addresses [post-merge code review feedback](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/75#pullrequestreview-2453017550)

### DIFF
--- a/aws-s3-transfer-manager/src/operation/upload_objects/worker.rs
+++ b/aws-s3-transfer-manager/src/operation/upload_objects/worker.rs
@@ -323,9 +323,12 @@ fn handle_failed_upload(
 
 #[cfg(test)]
 mod tests {
+    use std::sync::{Arc, Barrier};
+
     use aws_sdk_s3::operation::{
         abort_multipart_upload::AbortMultipartUploadOutput,
         create_multipart_upload::CreateMultipartUploadOutput, put_object::PutObjectOutput,
+        upload_part::UploadPartOutput,
     };
     use aws_smithy_mocks_experimental::{mock, RuleMode};
     use bytes::Bytes;
@@ -736,20 +739,37 @@ mod tests {
         assert_eq!(&crate::error::ErrorKind::OperationCancelled, err.kind());
     }
 
-    #[tokio::test]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn test_cancel_single_upload_via_multipart_upload() {
         let bucket = "test-bucket";
         let key = "test-key";
         let upload_id: String = "test-upload-id".to_owned();
 
+        let wait_till_create_mpu = Arc::new(Barrier::new(2));
+        let (resume_upload_single_obj_tx, resume_upload_single_obj_rx) =
+            tokio::sync::watch::channel(());
+        let resume_upload_single_obj_tx = Arc::new(resume_upload_single_obj_tx);
+
         let create_mpu = mock!(aws_sdk_s3::Client::create_multipart_upload).then_output({
+            let c = wait_till_create_mpu.clone();
             let upload_id = upload_id.clone();
             move || {
+                // This ensures that a cancellation signal won't be sent until `create_multipart_upload`.
+                c.wait();
+
+                // This increases the reliability of the test, ensuring that the cancellation signal has been sent
+                // and that `upload_single_obj` can now resume.
+                while !resume_upload_single_obj_rx.has_changed().unwrap() {
+                    std::thread::sleep(std::time::Duration::from_millis(100));
+                }
+
                 CreateMultipartUploadOutput::builder()
                     .upload_id(upload_id.clone())
                     .build()
             }
         });
+        let upload_part = mock!(aws_sdk_s3::Client::upload_part)
+            .then_output(|| UploadPartOutput::builder().build());
         let abort_mpu = mock!(aws_sdk_s3::Client::abort_multipart_upload)
             .match_requests({
                 let upload_id = upload_id.clone();
@@ -762,8 +782,8 @@ mod tests {
             .then_output(|| AbortMultipartUploadOutput::builder().build());
         let s3_client = mock_client_with_stubbed_http_client!(
             aws_sdk_s3,
-            RuleMode::Sequential,
-            &[create_mpu, abort_mpu]
+            RuleMode::MatchAny,
+            &[create_mpu, upload_part, abort_mpu]
         );
         let config = crate::Config::builder()
             .set_multipart_threshold(PartSize::Target(MIN_MULTIPART_PART_SIZE_BYTES))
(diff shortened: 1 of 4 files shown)


aws-s3-transfer-manager/src/operation/upload_objects/worker.rs
```diff
@@ -734,4 +739,58 @@ mod tests {
...
+        let upload_id: String = "test-upload-id".to_owned();
+
+        let create_mpu = mock!(aws_sdk_s3::Client::create_multipart_upload).then_output({
+            let upload_id = upload_id.clone();
+            move || {
+                CreateMultipartUploadOutput::builder()
+                    .upload_id(upload_id.clone())
+                    .build()
+            }
+        });
```
waahm7, line 756: If the upload is cancelled before invoking send(), I would expect nothing to happen, neither create nor abort. 

We will also need to consider what happens when we make the createMPU async instead of sync by having `initiate()` instead of send(). We need to ensure that we don't send abort without create. No need to fix that in this PR, just mentioning it.

waahm7: > Since you mentioned No need to fix that in this PR, just mentioning it. I can add TODO to incorporate the suggestion in the future, if that makes sense?

I was talking about the expected behavior of the test. This test relies on the internal detail of `orchestrate` that it will send a createMPU sync, which we will soon change. Is there a way to update this test to wait for the createMPU call to happen before sending the cancellation signal?

ysaito1001: > which we will soon change

Ah, this is the key information I was missing. Sounds like we want to write a test in a way that works both before and after the change in question. Let me look into that.

ysaito1001: > Is there a way to update this test to wait for the createMPU call to happen before sending the cancellation signal?

Looking at the current production code again, the answer is probably no, because the mock setup is inherently tied to the implementation details, i.e., what awaiting `awaiting `Upload::orchestrate` future entails. If awaiting `Upload::orchestrate` future entails `create_multipart_upload` like today then that's what shows up in s3 client's mock setup. If a future code change removes the need for `create_multipart_upload` as part of awaiting `orchestrate` future, then s3 client's mock setup won't need to mention `CreateMultipartUploadOutput`.

So I think the option is either
- Go with the current test, merge the PR, let it break when we change the behavior of awaiting `Upload::orchestrate`, and update the test that time
or
- Hold off on merging this PR, let a future PR of updating what is entailed in awaiting `Upload::orchestrate` be merged to the main first, and then merge that main branch to this PR

ysaito1001: > Is there a way to update this test to wait for the createMPU call to happen before sending the cancellation signal?

Looking at the current production code again, the answer is probably no, because the mock setup (in general) is inherently tied to the implementation details, i.e., what awaiting awaiting `Upload::orchestrate` future entails. If awaiting `Upload::orchestrate` future entails `create_multipart_upload` like today then that's what shows up in s3 client's mock setup. If a future code change removes the need for `create_multipart_upload` as part of awaiting `orchestrate` future, then s3 client's mock setup won't need to mention `CreateMultipartUploadOutput`.

So I think the option is either
- Go with the current test, merge the PR, let it break when we change the behavior of awaiting `Upload::orchestrate`, and update the test that time
or
- Hold off on merging this PR, let a future PR of updating what is entailed in awaiting `Upload::orchestrate` be merged to the main first, and then merge that main branch to this PR

ysaito1001: > Is there a way to update this test to wait for the createMPU call to happen before sending the cancellation signal?

Looking at the current production code again, the answer is probably no, because the mock setup (in general) is inherently tied to the implementation details, i.e., what awaiting `Upload::orchestrate` future entails. If awaiting `Upload::orchestrate` future entails `create_multipart_upload` like today then that's what shows up in s3 client's mock setup. If a future code change removes the need for `create_multipart_upload` as part of awaiting `orchestrate` future, then s3 client's mock setup won't need to mention `CreateMultipartUploadOutput`.

So I think the option is either
- Go with the current test, merge the PR, let it break when we change the behavior of awaiting `Upload::orchestrate`, and update the test that time
or
- Hold off on merging this PR, let a future PR of updating what is entailed in awaiting `Upload::orchestrate` be merged to the main first, and then merge that main branch to this PR

waahm7: > If a future code change removes the need for create_multipart_upload as part of awaiting orchestrate future, then s3 client's mock setup won't need to mention CreateMultipartUploadOutput.

The problem is that there might not be any abort if there is no create. So this test of `abort -> orchestrate -> test if abortMultipartUpload` is called won't work. Ideally, we should have this test with something like `orchestrate -> wait for the createMPU to happen -> abort -> make sure abortMultipartUpload is called`.

waahm7: I have tried to write a similar test for download in #80. Not sure if Rust testing frameworks are not advanced enough to test this reliably and efficiently, or if it is due to my lack of Rust knowledge.

ysaito1001: > we should have this test with something like orchestrate -> wait for the createMPU to happen -> abort -> make sure abortMultipartUpload is called

Hmm, if that's what we want to test, writing a test at this level will probably not give us enough control. It probably should be in `src/operation/upload` instead of `src/operation/uploads`.

ysaito1001: > we should have this test with something like orchestrate -> wait for the createMPU to happen -> abort -> make sure abortMultipartUpload is called

Hmm, if that's what we want to test, writing a test at this level will probably not give us enough control because we're subject to the predefined program control of [upload_single_obj](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/main/aws-s3-transfer-manager/src/operation/upload_objects/worker.rs#L235-L279). It probably should be in `src/operation/upload` instead of `src/operation/uploads`.

ysaito1001: It may sound silly but to reliably adhere to the order of `wait for the createMPU to happen -> abort -> make sure abortMultipartUpload` we should be writing a unit test against `UploadHandle`. Specifically, the system under test should start with a return value of `try_start_mpu_upload`, i.e. it's going to look like
```
let mut sut = try_start_mpu_upload(...); // ... will contain `ctx` in which mock S3 client is configured accordingly
sut.abort()
// test passes if mock expectations are satisfied 
```
What do you think?

ysaito1001: Disregard my comments above. Commit bde83ea should simulate the ping-pong execution flow of `CreateMPU` -> upload cancellation -> `AbortMPU` with the use of a barrier and a wait channel.

This ping-pong execution flow in the test should be reproduced regardless of whether `CreateMPU` is called synchronously (as in the current codebase) or asynchronously (which I assume to be implemented with future changes).

waahm7 (approved):

Issue #34, opened 2024-07-19
Customize SDK User Agent Header


aajtodd:
Set `User-Agent` header value for SEP, e.g. `ft/hll#s3-transfer`. This should be done using the SDK options for customizing the header as we want to retain the rest of the `User-Agent` header that ties this to the Rust SDK.

PR #79, opened 2024-12-04
Add user agent

TingDaoK:
*Issue #, if available:*

- https://github.com/awslabs/aws-s3-transfer-manager-rs/issues/34

*Description of changes:*

- Add an interceptor when we load the config to add the transfer manager metric for user agent
- Add a framework metadata for the transfer manager config and pass it down to the sdk s3 client as the framework name in the user agent.
- Only added for the `ConfigLoader`, we can decide what to do when user pass in their own client via config directly later.

### DIFF
--- a/aws-s3-transfer-manager/examples/cp.rs
+++ b/aws-s3-transfer-manager/examples/cp.rs
@@ -153,9 +153,6 @@ async fn do_download(args: Args) -> Result<(), BoxError> {
     let tm_config = aws_s3_transfer_manager::from_env()
         .concurrency(ConcurrencySetting::Explicit(args.concurrency))
         .part_size(PartSize::Target(args.part_size))
-        .frame_metadata(Some(
-            FrameworkMetadata::new("some-framework", Some(Cow::Borrowed("1.3"))).unwrap(),
-        ))
         .load()
         .await;
 
@@ -233,9 +230,6 @@ async fn do_upload(args: Args) -> Result<(), BoxError> {
     let tm_config = aws_s3_transfer_manager::from_env()
         .concurrency(ConcurrencySetting::Explicit(args.concurrency))
         .part_size(PartSize::Target(args.part_size))
-        .frame_metadata(Some(
-            FrameworkMetadata::new("some-framework", Some(Cow::Borrowed("1.3"))).unwrap(),
-        ))
         .load()
         .await;
 
--- a/aws-s3-transfer-manager/examples/cp.rs
+++ b/aws-s3-transfer-manager/examples/cp.rs
@@ -1,4 +1,3 @@
-use std::borrow::Cow;
 /*
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0
@@ -8,7 +7,6 @@ use std::path::PathBuf;
 use std::str::FromStr;
 use std::time;
 
-use aws_runtime::user_agent::FrameworkMetadata;
 use aws_s3_transfer_manager::io::InputStream;
 use aws_s3_transfer_manager::metrics::unit::ByteUnit;
 use aws_s3_transfer_manager::metrics::Throughput;
@@ -280,7 +278,6 @@ async fn main() -> Result<(), BoxError> {
         tracing_subscriber::fmt()
             .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
             .with_thread_ids(true)
-            .with_ansi(false) // Disable ANSI colors
             .init();
     }
 
--- a/aws-s3-transfer-manager/examples/cp.rs
+++ b/aws-s3-transfer-manager/examples/cp.rs
@@ -152,7 +152,7 @@ async fn do_download(args: Args) -> Result<(), BoxError> {
     let tm_config = aws_s3_transfer_manager::from_env()
         .concurrency(ConcurrencySetting::Explicit(args.concurrency))
         .part_size(PartSize::Target(args.part_size))
-        .app_name(Some(AppName::new("rust_ts_cp").unwrap()))
+        .app_name(Some(AppName::new("transfer_manager_example_cp").unwrap()))
         .load()
         .await;
 
@@ -230,7 +230,7 @@ async fn do_upload(args: Args) -> Result<(), BoxError> {
     let tm_config = aws_s3_transfer_manager::from_env()
         .concurrency(ConcurrencySetting::Explicit(args.concurrency))
         .part_size(PartSize::Target(args.part_size))
-        .app_name(Some(AppName::new("rust_ts_cp").unwrap()))
+        .app_name(Some(AppName::new("transfer_manager_example_cp").unwrap()))
         .load()
         .await;
 
--- a/aws-s3-transfer-manager/examples/cp.rs
+++ b/aws-s3-transfer-manager/examples/cp.rs
@@ -7,6 +7,7 @@ use std::path::PathBuf;
 use std::str::FromStr;
 use std::time;
 
+use aws_config::AppName;
 use aws_s3_transfer_manager::io::InputStream;
 use aws_s3_transfer_manager::metrics::unit::ByteUnit;
 use aws_s3_transfer_manager::metrics::Throughput;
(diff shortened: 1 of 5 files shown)


aws-s3-transfer-manager/examples/cp.rs
```diff
@@ -151,6 +152,7 @@ async fn do_download(args: Args) -> Result<(), BoxError> {
     let tm_config = aws_s3_transfer_manager::from_env()
         .concurrency(ConcurrencySetting::Explicit(args.concurrency))
         .part_size(PartSize::Target(args.part_size))
+        .app_name(Some(AppName::new("rust_ts_cp").unwrap()))
```
waahm7, line 155: What is ts? I think it would be better to just spell it `transfer_manager_sample_cp`.

aws-s3-transfer-manager/src/config/loader.rs
```diff
@@ -3,13 +3,37 @@
...
     types::{ConcurrencySetting, PartSize},
     Config,
 };
 
+#[derive(Debug)]
+struct TransferManagerFeatureInterceptor;
+
+impl Intercept for TransferManagerFeatureInterceptor {
+    fn name(&self) -> &'static str {
+        "TransferManagerFeature"
```
waahm7, line 23: Do we want to make it more unique like "AwsTransferManager"? 🤔

aws-s3-transfer-manager/examples/cp.rs
```diff
@@ -278,6 +281,7 @@ async fn main() -> Result<(), BoxError> {
         tracing_subscriber::fmt()
             .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
             .with_thread_ids(true)
+            .with_ansi(false) // Disable ANSI colors
```
aajtodd, line 284: question: why?

aws-s3-transfer-manager/src/config.rs
```diff
@@ -43,6 +46,11 @@ impl Config {
         &self.concurrency
     }
 
+    /// Returns the name of the app that is using the transfer manager, if it was provided.
+    pub fn app_name(&self) -> Option<&AppName> {
```
aajtodd, line 50: **fix**: I don't think we want to expose app name. (1) we allow taking an S3 client (today anyway) that can already have app name configured, (2) I don't think this is what mountpoint will be looking for out of a UA header.

Is there something I'm missing here about why we are exposing this?

aajtodd (changes requested):

aws-s3-transfer-manager/examples/cp.rs
TingDaoK: with the ansi color, it's unreadable from pure text. I think it will be easier to read and parse.

aws-s3-transfer-manager/src/config.rs
TingDaoK: From my reading of the sdk, the app name is just the string for the SDK to add in the user agent with some format check.

1. I think if we take the s3 client and want to respect the setting from transfer manager, we will be adding the user agent setting from transfer manager to the passed in client as the app name.

2. what do you think mountpoint will be looking for? 
I assume they just want a string, and I thought the app name is the wrap of string for SDK to be provided in the user agent.

aajtodd: > From my reading of the sdk, the app name is just the string for the SDK to add in the user agent with some format check.

App name (`appId`) comes from the UA SEP. It's meant for end user applications (though I've yet to see someone use it 🤷‍♂️  in any SDK). Mountpoint _could_ go through that but they are more like framework metadata. 


> I think if we take the s3 client and want to respect the setting from transfer manager, we will be adding the user agent setting from transfer manager to the passed in client as the app name.

The issue is the user _could_ have set app name on the client. I think maybe we just leave this off for now and gather requirements from mountpoint to guide us. I _think_ [this](https://github.com/awslabs/mountpoint-s3/blob/993b0d42c0b172fb82fea39ac3964a1b3c74d4cf/mountpoint-s3-client/src/user_agent.rs) is their user agent implementation though. From the looks of it we capture a lot of the md/* fields they are gathering I think so maybe they just need `lib/mountpoint-s3-client#<version>`.

TingDaoK: the framework metadata is supposed to be used as frameworks **owned** by the SDKs.

> For frameworks owned by the SDKs, framework metadata MUST be included in the header following this format: "lib" / name ["#" version]. Additional metadata MAY be appended.

It seems to be an implementation detail for SDKs to me. And mountpoint is more like the application user, but yeah, they are more like a framework for the end user?
 I guess we can chat with them and see what are they expecting?

aws-s3-transfer-manager/src/config/loader.rs
```diff
@@ -3,13 +3,37 @@
...
+    }
+
+    fn read_before_execution(
+        &self,
+        _ctx: &aws_sdk_s3::config::interceptors::BeforeSerializationInterceptorContextRef<'_>,
+        cfg: &mut aws_sdk_s3::config::ConfigBag,
+    ) -> Result<(), aws_sdk_s3::error::BoxError> {
+        cfg.interceptor_state()
+            .store_append::<AwsSdkFeature>(AwsSdkFeature::S3Transfer);
```
ysaito1001, line 32: We can probably remove `::<AwsSdkFeature>` since it'll be deduced.

aws-s3-transfer-manager/src/config/loader.rs
ysaito1001: Yeah, or maybe `S3TransferManager` (sorry [my naming there](https://github.com/smithy-lang/smithy-rs/blob/cd4d6504397a616b5dee6bc60e62975c924f0f5d/aws/sdk/integration-tests/s3/tests/business_metrics.rs#L20) was based on the assumption that it was defined in s3 integration test so no s3 in the naming)? Also, it can help avoid small code duplication if we create a module private constant string for the interceptor name and refer it from the unit test below.

aws-s3-transfer-manager/src/config/loader.rs
```diff
@@ -52,17 +76,60 @@ impl ConfigLoader {
...
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::types::{ConcurrencySetting, PartSize};
+    use aws_sdk_s3::config::Intercept;
+    use aws_types::app_name::AppName;
+
+    #[tokio::test]
+    async fn load_with_app_name_and_interceptor() {
```
ysaito1001, line 116: Nice!

aws-s3-transfer-manager/examples/cp.rs
```diff
@@ -152,7 +153,9 @@ async fn do_download(args: Args) -> Result<(), BoxError> {
     let tm_config = aws_s3_transfer_manager::from_env()
         .concurrency(ConcurrencySetting::Explicit(args.concurrency))
         .part_size(PartSize::Target(args.part_size))
-        .app_name(Some(AppName::new("transfer_manager_example_cp").unwrap()))
+        .frame_metadata(Some(
```
aajtodd, line 156: Probably don't need this on the `cp` example application

aws-s3-transfer-manager/src/config.rs
```diff
@@ -135,8 +135,8 @@ impl Builder {
     ///
     /// This _optional_ name is used to identify the application in the user agent that
     /// gets sent along with requests.
-    pub fn app_name(mut self, app_name: Option<AppName>) -> Self {
-        self.app_name = app_name;
+    pub fn frame_metadata(mut self, frame_metadata: Option<FrameworkMetadata>) -> Self {
```
aajtodd, line 138: **suggestion**: Spell this out, `framework_metadata`.

**suggestion**: Maybe doc hide these config functions as most people shouldn't need to care or know about them.

Applies here and elsewhere

aajtodd (approved):
Fix and ship

aws-s3-transfer-manager/src/config.rs
```diff
@@ -43,6 +46,12 @@ impl Config {
         &self.concurrency
     }
 
+    #[doc(hidden)]
```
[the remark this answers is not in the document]
TingDaoK: I actually saw that in rust-sdk we put the attributes above the comments. https://github.com/awslabs/aws-sdk-rust/blob/main/sdk/aws-runtime/src/user_agent.rs#L172-L191

what kind of style rules we have? Are we put multiple attributes above comments?

PR #80, opened 2024-12-04
WIP | Abort Download Object

waahm7:
*Issue #, if available:*

*Description of changes:*


aws-s3-transfer-manager/src/operation/download.rs
```diff
@@ -9,10 +9,10 @@ use aws_sdk_s3::error::DisplayErrorContext;
 /// Request type for dowloading a single object from Amazon S3
 pub use input::{DownloadInput, DownloadInputBuilder};
 
-/// Abstractions for response bodies and consuming data streams.
-pub mod body;
 /// Operation builders
 pub mod builders;
+/// Abstractions for responses and consuming data streams.
+pub mod output;
```
aajtodd, line 15: **fix**: Be consistent across operations. e.g. [upload output module is not public](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/main/aws-s3-transfer-manager/src/operation/upload.rs#L9) but it [re-exports the output type](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/main/aws-s3-transfer-manager/src/operation/upload.rs#L23)

aws-s3-transfer-manager/src/operation/download/output.rs
```diff
@@ -12,17 +12,17 @@ use crate::io::AggregatedBytes;
...
 
-/// Stream of binary data representing an Amazon S3 Object's contents.
+/// Stream of [ChunkOutput] representing an Amazon S3 Object's contents and metadata.
 ///
 /// Wraps potentially multiple streams of binary data into a single coherent stream.
 /// The data on this stream is sequenced into the correct order.
 #[derive(Debug)]
-pub struct Body {
-    inner: UnorderedBody,
+pub struct Output {
```
aajtodd, line 20: **suggestion/discuss**: Should this be consistent with the other operations output types (e.g. `DownloadOutput`)?

aws-s3-transfer-manager/src/operation/download/output.rs
```diff
@@ -42,35 +42,35 @@ pub struct ChunkOutput {
 // TODO: Do we want to expose something to yield multiple chunks in a single call, like
 // recv_many/collect, etc.? We can benchmark to see if we get a significant performance boost once
 // we have a better scheduler in place.
-impl Body {
-    /// Create a new empty body
+impl Output {
```
aajtodd, line 45: **future**: We may want to have a conversion that allows you to just access the bytes off the stream (i.e. user only cares about the contents and not any of the intermediate metadata), we can wait though as it can be added in a backwards compatible way

```rs
fn into_body(self) -> Body
```

aws-s3-transfer-manager/src/operation/download.rs
```diff
@@ -115,13 +115,37 @@ async fn send_discovery(
     }
 
     // acquire a permit for discovery
-    let permit = ctx.handle.scheduler.acquire_permit().await?;
+    let permit = ctx.handle.scheduler.acquire_permit().await;
+    let permit = match permit {
+        Ok(permit) => permit,
+        Err(err) => {
+            if comp_tx.send(Err(err)).await.is_err() {
+                tracing::debug!("Download handle for key({}) has been dropped, aborting during the discovery phase", input.key.unwrap_or("None".to_string()));
```
aajtodd, line 123: **fix**: You should be able to just pass `input.key` if you change the format to `key({:?})`

e.g.

```rs
tracing::debug!("Download handle for key({:?}) has been dropped, aborting during the discovery phase", input.key);
```

Same thing elsewhere

aws-s3-transfer-manager/tests/download_test.rs
```diff
@@ -148,26 +156,43 @@ async fn test_download_ranges() {
...
         .key("test-object")
         .initiate()
         .unwrap();
 
-    handle.join().await.unwrap();
+    let _ = handle.ouput_mut().next().await;
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_abort_download() {
```
aajtodd, line 180: You don't have to necessarily use it here but one thing to make you aware of is you can leverage `capture_test_logs` ([example](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/cancel-single-download/aws-s3-transfer-manager/tests/upload_test.rs#L117)) which you can use to assert the logs contain certain things (e.g. to verify particular branches were hit).

aws-s3-transfer-manager/src/operation/download.rs
waahm7: Thanks, updated.

aws-s3-transfer-manager/src/operation/download/output.rs
waahm7: Thanks, yeah, DownloadOutput makes more sense since there are different types of outputs. I have updated it.

aws-s3-transfer-manager/src/operation/download.rs
waahm7: Thanks, updated.

ysaito1001 (commented):
Looks good, leaving minor comments.

aws-s3-transfer-manager/examples/cp.rs
```diff
@@ -296,8 +296,8 @@ async fn main() -> Result<(), BoxError> {
     Ok(())
 }
 
-async fn write_body(body: &mut Body, mut dest: fs::File) -> Result<(), BoxError> {
-    while let Some(chunk) = body.next().await {
+async fn write_body(output: &mut Output, mut dest: fs::File) -> Result<(), BoxError> {
```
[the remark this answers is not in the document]
waahm7: I think `write_body` makes more sense here since we are only writing the body and ignoring the metadata.

ysaito1001 (approved):

aws-s3-transfer-manager/src/operation/download/handle.rs
```diff
@@ -53,29 +53,26 @@ impl DownloadHandle {
...
+    /// Object content and metadata
+    pub fn output(&self) -> &Output {
+        &self.output
     }
 
-    /// Mutable reference to the body
-    pub fn body_mut(&mut self) -> &mut Body {
-        &mut self.body
+    /// Mutable reference to the output
+    pub fn ouput_mut(&mut self) -> &mut Output {
```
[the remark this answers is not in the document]
waahm7: Thanks for catching this, updated.

aws-s3-transfer-manager/src/operation/download/handle.rs
```diff
@@ -53,29 +53,26 @@ impl DownloadHandle {
...
-    #[tracing::instrument(skip_all, level = "debug", name = "join-download")]
-    pub async fn join(mut self) -> Result<(), crate::error::Error> {
-        self.body.close();
-
-        self.discovery.await??;
+    /// Abort the download and cancel any in-progress work.
+    pub async fn abort(mut self) {
+        self.output.close();
+        self.discovery.abort();
+        let _ = self.discovery.await;
```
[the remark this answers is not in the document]
ysaito1001: Oh, it's because we're dropping an error using `let _ = self.discovery.await`. Instead, I was referring to
```
if let Err(e) = self.discovery.await {
  // debug log `e` for the record
}
```

> why would the discovery task panic?

I don't know either (according to [the doc](https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html#examples) it's due to panics), but because of that, I was thinking whether at least it records that fact as a debug log. Maybe my suggestion is instead of ignoring the result by `let _ =` we should use either `expect` or debug log that error.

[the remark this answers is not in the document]
waahm7: The result in most cases would be JoinedError as stated in the docs:
```
Abort the task associated with the handle.

Awaiting a cancelled task might complete as usual if the task was already completed at the time it was cancelled, but most likely it will fail with a [cancelled](https://docs.rs/tokio/latest/tokio/task/struct.JoinError.html#method.is_cancelled) JoinError.
```

So I think logging the error here will not be useful. Even if it's not this error, we don't care about errors if we are aborting the download.

aws-s3-transfer-manager/src/operation/download/service.rs
```diff
@@ -69,25 +70,33 @@ async fn download_specific_chunk(
...
-        ))
-        .await?;
-
-    Ok(ChunkOutput {
-        seq,
-        data: body,
-        metadata: resp.into(),
-    })
+    let mut cancel_rx = ctx.state.cancel_rx.clone();
+    tokio::select! {
```
graebm, line 74: N E A T

aws-s3-transfer-manager/src/operation/download/service.rs
```diff
@@ -139,12 +148,25 @@ pub(super) fn distribute_work(
 
         let svc = svc.clone();
         let comp_tx = comp_tx.clone();
+        let cancel_tx = ctx.state.cancel_tx.clone();
 
         let task = async move {
-            // TODO: If downloading a chunk fails, do we want to abort the download?
             let resp = svc.oneshot(req).await;
+            if let Err(err) = &resp {
```
graebm, line 155: trivial: comment
```suggestion
            // If any chunk fails, send cancel notification, to kill any other in-flight chunks 
            if let Err(err) = &resp {
```

aws-s3-transfer-manager/src/operation/download/service.rs
```diff
@@ -139,12 +148,25 @@ pub(super) fn distribute_work(
...
+                if *err.kind() != ErrorKind::OperationCancelled && cancel_tx.send(true).is_err() {
+                    tracing::debug!(
+                        "all receiver ends have dropped, unable to send a cancellation signal"
+                    );
+                }
+            }
+
             if let Err(err) = comp_tx.send(resp).await {
```
graebm, line 163: Is it OK that ErrorKind::OperationCancelled` gets sent on `comp_tx`? That could result in MANY chunks sending their cancellation error on the chunk transmitter

Do all of these end up reported to the user via the Body/Output iterator? Or does it filter errors so the user only sees the 1st? Can we guarantee that the ACTUAL error gets reported to the user before any cancellation errors?

waahm7: As soon as we encounter an error, we close the channel so user will only see the first error. 

> Can we guarantee that the ACTUAL error gets reported to the user before any cancellation errors?

Tokio does guarantee that "All data sent on Sender will become available on Receiver in the same order as it was sent." However, if the channel is full, all of them will be waiting at an await point. In that case (not 100% sure), the order won't be guaranteed. We will need a fair bit of complexity to work around that, i.e., maybe collect all the errors and then only send those that are not `OperationCancelled` errors. Not sure if there is a simpler way to do this. @aajtodd any ideas?

aajtodd: Do we need to send operation cancelled across the channel at all?

waahm7: I was thinking there can be a case where only the operation cancelled error is there, but yeah, we can just not send that error at all.

aajtodd: A user wouldn't act on that error right? I wouldn't think so anyway, it doesn't tell them anything they don't already know if they cancelled the operation, thus it probably doesn't make sense to yield it to them.

aws-s3-transfer-manager/src/operation/download/output.rs
```diff
@@ -42,35 +42,35 @@ pub struct ChunkOutput {
 // TODO: Do we want to expose something to yield multiple chunks in a single call, like
 // recv_many/collect, etc.? We can benchmark to see if we get a significant performance boost once
 // we have a better scheduler in place.
-impl Body {
-    /// Create a new empty body
+impl DownloadOutput {
```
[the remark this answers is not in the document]
waahm7: Thanks, discussed offline, I have reverted the naming to body.

aws-s3-transfer-manager/src/operation/download/service.rs
waahm7: Thanks, I have updated it to ignore `OperationCancelled` errors.

aws-s3-transfer-manager/src/operation/download/output.rs
waahm7: Discussed offline, reverted the name change.

aajtodd (approved):

aws-s3-transfer-manager/src/operation/download/handle.rs
```diff
@@ -23,11 +23,11 @@ pub struct DownloadHandle {
     /// Object metadata.
     pub(crate) object_meta: OnceCell<ObjectMetadata>,
 
-    /// The object content
+    /// The object content and metadata
```
graebm, line 26: trivial: "and metadata" makes it sound like there's 1 metadata, and this is where you get it ... which is confusing because there's also the `object_meta` field.

Probably more interesting to point out that the object data will come in chunks. You can point out that each chunk has its own metadata...
```suggestion
    /// The object content, in chunks, and the metadata for each chunk
```
or not mention that detail, since most people probably aren't interested
```suggestion
    /// The object content, in chunks
```

aws-s3-transfer-manager/src/operation/download/handle.rs
```diff
@@ -53,7 +53,7 @@ impl DownloadHandle {
         Ok(meta)
     }
 
-    /// Object content
+    /// Object content and metadata
```
graebm, line 56: same comment about same comment

graebm (approved):

aws-s3-transfer-manager/src/operation/download/handle.rs
waahm7: Thanks, updated.

PR #84, opened 2024-12-19
Make createMPU Call Async

waahm7:
*Description of changes:*
This PR is just a refactor of the Upload APIs and doesn't make any functional changes.

- This is a follow-up to https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/73 to return the handle ASAP for the UploadObject API as well. This approach will make it easier to implement features like reading the first part to determine PutObject vs MPU.
- `upload_handle.abort()` now takes ownership of the handle instead of a reference. It makes more sense to take ownership so that customers can't call `join` or do other things with the handle after aborting, as it won't be valid.
- I have added a TODO to make the completeMPU call asynchronous as well, instead of waiting for users to call `join`. From testing, completeMPU can be an expensive call for large objects.
- `upload_id` was an Option because it won't be available for PutObject. I have moved it to the `MultipartUploadData` struct, and it will be non-optional wherever MPU is involved.
- Move `test_cancel_single_upload_via_multipart_upload` test to upload_tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,9 +15,9 @@ env:
   # Change to specific Rust release to pin
   rust_stable: stable
   rust_nightly: nightly-2024-07-07
-  rust_clippy: '1.79'
+  rust_clippy: '1.81'
   # When updating this, also update relevant docs
-  rust_min: '1.79'
+  rust_min: '1.81'
 
 
 defaults:
@@ -55,7 +55,7 @@ jobs:
       - docs
       - minrust
     steps:
-      - run: exit 0 
+      - run: exit 0
 
   test-hll:
     name: Test S3 transfer manager HLL
@@ -133,7 +133,7 @@ jobs:
         run: |
           cargo doc --lib --no-deps --all-features --document-private-items
         env:
-          RUSTFLAGS: --cfg docsrs 
+          RUSTFLAGS: --cfg docsrs
           RUSTDOCFLAGS: --cfg docsrs
 
   minrust:
@@ -246,7 +246,7 @@ jobs:
       - name: Install cargo-hack
         uses: taiki-e/install-action@cargo-hack
       - uses: Swatinem/rust-cache@v2
-      - name: check --feature-powerset 
+      - name: check --feature-powerset
         run: cargo hack check --all --feature-powerset
 
   # TODO - get cross check working
--- a/aws-s3-transfer-manager/examples/cp.rs
+++ b/aws-s3-transfer-manager/examples/cp.rs
@@ -252,8 +252,7 @@ async fn do_upload(args: Args) -> Result<(), BoxError> {
         .bucket(bucket)
         .key(key)
         .body(stream)
-        .send()
-        .await?;
+        .initiate()?;
 
     let _resp = handle.join().await?;
     let elapsed = start.elapsed();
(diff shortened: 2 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -15,9 +15,9 @@ env:
   # Change to specific Rust release to pin
   rust_stable: stable
   rust_nightly: nightly-2024-07-07
-  rust_clippy: '1.79'
+  rust_clippy: '1.81'
   # When updating this, also update relevant docs
-  rust_min: '1.79'
+  rust_min: '1.81'
```
waahm7, line 20: What else I am supposed to update here?

aajtodd: thats it right now

aws-s3-transfer-manager/src/operation/upload/service.rs
```diff
@@ -96,67 +97,63 @@ pub(super) fn upload_part_service(
 /// * stream - the body input stream
 /// * part_size - the part_size for each part
 pub(super) fn distribute_work(
-    handle: &mut UploadHandle,
+    mpu_data: &mut MultipartUploadData,
+    ctx: UploadContext,
     stream: InputStream,
     part_size: u64,
+    upload_id: String,
```
aajtodd, line 104: isn't upload id part of `mpu_data`? 

Also the whole point of `UploadContext` was so to thread through all the state needed for an operation...

waahm7: > isn't upload id part of mpu_data?

Thanks, fixed by not passing upload_id explicitly. 

> Also the whole point of UploadContext was so to thread through all the state needed for an operation...

Yeah, but the problem was that we no longer have mutable access to UploadContext in the createMPU call, because we do not own the handle at that point.

aws-s3-transfer-manager/src/operation/upload/builders.rs
```diff
@@ -917,6 +917,6 @@ impl crate::operation::upload::input::UploadInputBuilder {
     ) -> Result<UploadHandle, crate::error::Error> {
         let mut fluent_builder = client.upload();
         fluent_builder.inner = self;
-        fluent_builder.send().await
+        fluent_builder.initiate()
```
graebm, line 920: if we're changing the name-and-asyncness of `UploadFluentBuilder::send()` then we should change the name-and-asyncness of `UploadInputBuilder::send_with()`

aws-s3-transfer-manager/examples/cp.rs
```diff
@@ -252,8 +252,7 @@ async fn do_upload(args: Args) -> Result<(), BoxError> {
         .bucket(bucket)
         .key(key)
         .body(stream)
-        .send()
-        .await?;
+        .initiate()?;
```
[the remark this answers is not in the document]
waahm7: Not really, Aaron suggested changing the name from `send` to avoid confusion with the Rust SDK at https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/73#discussion_r1838162739, and I couldn't think of something better than initiate/send.

aws-s3-transfer-manager/src/operation/upload/builders.rs
waahm7: Thanks, yeah, I can make it consistent once we have agreed on the `initiate` name in the previous feedback.

graebm (approved):
fix & ship

aajtodd (approved):

Issue #3, opened 2024-07-18
Checksum support


aajtodd:
This is a tracking issue for implementing checksum support.

Checksums in an S3 transfer manager are a bit tricky due to requests being potentially split into concurrent ranged `GET` or multipart uploads for get and put object respectively.

We probably want to put together a design doc for this. Need to cover all known use cases including but not limited to:

**Uploads**:
* Letting users control checksums for each part for uploads

**Downloads**
* Giving users the ability to see individual part checksums?

**Both**
* Try to have request/response types prevent users from constructing impossible to fulfill requests. Right now the input/output response types mirror the `aws_sdk_s3` API (or more specifically our API tries to look like a single Get/Put object request/response but maybe they shouldn't).

Issue #25, opened 2024-07-18
Integration tests


aajtodd:
Establish an integration test suite we can run. Some of these way may be able to use the replay recorder to capture traffic and turn them into more of a unit test.

e.g. single vs multipart upload, checksum scenarios, express buckets, pause/resume, etc.

PR #86, opened 2025-01-10
Checksums for Upload

graebm:
**Issue #:**
https://github.com/awslabs/aws-s3-transfer-manager-rs/issues/3

Users should have control over checksums during upload. They should be able to choose the checksum algorithm, and multipart checksum type. If the user already has a checksum, they should be able to provide it and have that value sent to S3, to ensure verify data integrity from absolute beginning to absolute end.

**Challenges:**

1) It's easy for users to misconfigure checksum settings. S3 models checksums with 7 distinct input fields (checksum_crc32, checksum_crc32_c, checksum_crc64_nvme, checksum_sha1, checksum_sha256, checksum_type, checksum_algorithm). Certain combinations of fields don't work together (e.g. not all algorithms can be used with all multipart-checksum-types, it's invalid to set multiple checksum values, etc).

    We should make it impossible to misconfigure, or at least make it hard to misconfigure and quickly detect a misconfiguration.

2) The "Composite" multipart checksum type makes things complicated. If we let users provide a composite checksum up front, the Transfer Manager would **HAVE TO** do a multipart upload (3+ HTTP requests). It wouldn't be able to optimize a small upload into 1 PutObject HTTP request, because PutObject can only provide full object checksums.

    And it's not clear anyone *wants* to use composite checksums. In the past, multipart uploads had to use composite checksums, or have no checksum whatsoever. But as of Re:Invent 2024 (see [announcement](https://aws.amazon.com/blogs/aws/introducing-default-data-integrity-protections-for-new-objects-in-amazon-s3/)) S3 supports full object checksums on multipart uploads (for CRC algorithms, but not SHA algorithms). All AWS SDKs will (soon) default to sending CRC full object checksums. S3 now defaults to assigning full object CRC64NVME checksums to all objects if the client didn't already do it. So full object checksums are now the default, and much simpler to use and understand. Should we even support uploading with composite checksums and all their complexity in this brand new codebase?

    But there are still some valid use cases for uploading with composite checksums. First, if the user needs to use a SHA algorithm (S3 currently only allows CRC algorithms to do full object checksums). Second, if the user really actually cares about the checksums on their individual parts (S3 does not currently store a checksum on individual parts if the upload uses a full object checksum in the end).

    So... we should probably still support composite checksums. But maybe don't bend over too far to support complicated use cases (e.g. letting user provide composite checksums up front). But also, keep the door open to supporting it in the future, if someone needs it bad enough to justify all the complexity.

**Description of changes:**

1) `ChecksumStrategy` is new struct, which serves as an optional field on `UploadInput`. This is easier to use than the 7 distinct fields mentioned above. It has many constructors to help users create a valid configuration. There's also a `ChecksumStrategyBuilder` that lets someone set the individual fields, if that's easier for their code to deal with. Its `build()` function checks for all possible misconfiguration.

2) Let users provide a full object checksum up front, but NOT provide composite checksum up front (see challenge 2 above on why that would be hard). We could still support it in the future, but probably nobody wants this. Also, we WILL support providing checksums *per part* in the future, regardless of the multipart checksum type, and that is very likely good enough for someone that needs to use composite checksums and wants to provide the values themselves.

**Planned Future Work:**
1) Let users provide a full object checksum AFTER streaming the body.
2) Let users provide a checksum for each part as it's streamed (this will work with both full object and composite checksums)

### DIFF
--- a/aws-s3-transfer-manager/src/operation/upload/builders.rs
+++ b/aws-s3-transfer-manager/src/operation/upload/builders.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 
 use crate::types::FailedMultipartUploadPolicy;
 
-use super::{UploadHandle, UploadInputBuilder};
+use super::{ChecksumStrategy, UploadHandle, UploadInputBuilder};
 
 /// Fluent builder for constructing a single object upload transfer
 #[derive(Debug)]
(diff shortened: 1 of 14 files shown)


aws-s3-transfer-manager/src/operation/upload/checksum_strategy.rs
```diff
@@ -0,0 +1,283 @@
...
+///
+/// <code>CRC64NVME</code> checksums are calculated by default (if no strategy is set and the underlying
+/// S3 client is configured with the default [`aws_sdk_s3::config::RequestChecksumCalculation::WhenSupported`]).
+///
+/// To disable checksum calculation, do not set a [`ChecksumStrategy`] and make sure the underlying S3 client is
+/// configured with the non-default [`aws_sdk_s3::config::RequestChecksumCalculation::WhenRequired`].
+/// If you do this, S3 will still calculate a `CRC64NVME` full object checksum for the object.
+#[non_exhaustive]
+#[derive(Debug, Clone)]
+pub struct ChecksumStrategy {
```
aajtodd, line 20: Does this file and it's types belong under the [types](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/main/aws-s3-transfer-manager/src/types.rs) module. You can keep it as a separate file as is by just creating a new `types/` directory and exporting just what you want to expose from `types.rs` if that's the direction we decide to go. 

We have a few other operation specific types, we should probably be consistent or define some guidelines on when something goes into `types` vs is part of the specific operation.

aws-s3-transfer-manager/src/operation/upload/checksum_strategy.rs
```diff
@@ -0,0 +1,283 @@
...
+        full_object_checksum: Option<String>,
+    ) -> Self {
+        Self {
+            algorithm,
+            type_if_multipart,
+            full_object_checksum,
+        }
+    }
+
+    pub(crate) fn validate(&self) -> Result<(), crate::error::Error> {
```
aajtodd, line 160: PR description mentions this is public not just pub crate.

aws-s3-transfer-manager/src/operation/upload/handle.rs
```diff
@@ -205,16 +205,29 @@ async fn complete_upload(handle: UploadHandle) -> Result<UploadOutput, crate::er
...
+
+                if let Some(value) = &checksum_strategy.full_object_checksum {
+                    // We have the full-object checksum value, so set it
+                    req = match &checksum_strategy.algorithm {
+                        aws_sdk_s3::types::ChecksumAlgorithm::Crc32 => req.checksum_crc32(value),
+                        aws_sdk_s3::types::ChecksumAlgorithm::Crc32C => req.checksum_crc32_c(value),
+                        aws_sdk_s3::types::ChecksumAlgorithm::Crc64Nvme => {
+                            req.checksum_crc64_nvme(value)
+                        }
+                        algo => panic!("unexpected algorithm `{algo}` for full object checksum"),
```
aajtodd, line 225: if this is meant to be a state a user can't get into at this point you might use `unreachable!` instead

aws-s3-transfer-manager/src/operation/upload/builders.rs
```diff
@@ -232,123 +232,55 @@ impl UploadFluentBuilder {
...
-        self.inner = self.inner.checksum_sha256(input);
+
+    /// Strategy for calculating checksum values during upload.
+    /// These values are sent to S3 and used to verify the integrity of the uploaded data.
+    /// For more information, see <https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html>.
+    ///
+    /// You can set a specific [`ChecksumStrategy`], if you wish to choose the
+    /// checksum algorithm or already know the checksum value.
+    ///
+    /// <code>CRC64NVME</code> checksums are calculated by default (if no strategy is set and the underlying
```
landonxjames, line 243: Nit: Pretty sure you can get away with backticks here instead of the full `<code>` tag

aws-s3-transfer-manager/tests/upload_checksum_test.rs
```diff
@@ -0,0 +1,829 @@
...
+    },
+    types::{ChecksumAlgorithm, ChecksumType},
+};
+use aws_smithy_mocks_experimental::{mock, Rule, RuleMode};
+use aws_smithy_runtime::test_util::capture_test_logs::capture_test_logs;
+use bytes::Bytes;
+use pin_project_lite::pin_project;
+use test_common::mock_client_with_stubbed_http_client;
+
+const MEBIBYTE: usize = 1024 * 1024;
```
aajtodd, line 27: We have some types for this already, we might want to be consistent and just use them ([example](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/dca33b430f4f89da6cc61c8b53f4d2232bc528f0/aws-s3-transfer-manager/src/config.rs#L15))

aws-s3-transfer-manager/tests/upload_checksum_test.rs
```diff
@@ -0,0 +1,829 @@
...
+    match checksum_type {
+        ChecksumType::Composite => calculate_composite_checksum(algorithm, parts),
+        ChecksumType::FullObject => calculate_full_object_checksum(algorithm, parts),
+        _ => panic!("unrecognized checksum type"),
+    }
+}
+
+/// Calculate ETag as it appears in S3 headers, with the extra quotes and everything
+fn calculate_etag(data: &[u8]) -> String {
+    let digest = md5::compute(data);
```
aajtodd, line 82: Am I missing something or is `md5` crate missing from our dependencies in `Cargo.toml`?

aws-s3-transfer-manager/src/operation/upload/builders.rs
```diff
@@ -232,123 +232,55 @@ impl UploadFluentBuilder {
...
+    /// configured with the non-default [`aws_sdk_s3::config::RequestChecksumCalculation::WhenRequired`].
+    pub fn checksum_strategy(mut self, input: ChecksumStrategy) -> Self {
+        self.inner = self.inner.checksum_strategy(input);
         self
     }
-    /// <p>This header can be used as a data integrity check to verify that the data received is the same data that was originally sent. This header specifies the base64-encoded, 256-bit SHA-256 digest of the object. For more information, see <a href="https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html">Checking object integrity</a> in the <i>Amazon S3 User Guide</i>.</p>
-    pub fn set_checksum_sha256(mut self, input: Option<String>) -> Self {
-        self.inner = self.inner.set_checksum_sha256(input);
+
+    /// Strategy for calculating checksum values during upload.
```
landonxjames, line 253: This comment seems to be exactly replicated in  a few places, might make more sense to move it out to a .md file and include it
```
#![doc = include_str!("path/to/docs.md")]
```

aws-s3-transfer-manager/tests/upload_checksum_test.rs
```diff
@@ -0,0 +1,829 @@
...
+        InputStream::from_part_stream(TestStream::new(parts.clone()))
+    } else {
+        parts[0].clone().into()
+    };
+
+    let tm_config = aws_s3_transfer_manager::Config::builder()
+        .client(s3_client)
+        .part_size(PartSize::Target(PART_SIZE as u64))
+        .multipart_threshold(PartSize::Target(PART_SIZE as u64))
+        .concurrency(ConcurrencySetting::Explicit(1)) // guarantee parts sent in order
```
aajtodd, line 494: This is good/safe for unit tests. Do we need/want to consider one or more tests against real S3?

aajtodd (commented):
Overall looks good. Few questions/minor comments.

aws-s3-transfer-manager/src/operation/upload/checksum_strategy.rs
```diff
@@ -0,0 +1,283 @@
...
+            algorithm: ChecksumAlgorithm::Sha256,
+            type_if_multipart: ChecksumType::Composite,
+            full_object_checksum: None,
+        }
+    }
+
+    /// Create checksum strategy from parts.
+    ///
+    /// It is possible to create an invalid strategy this way,
+    /// use [`ChecksumStrategy::validate()`] to check. e.g. S3 does not allow
```
landonxjames, line 145: `ChecksumStrategy::validate()` is only `pub(crate)` so we might want to remove this reference from documentation on a `pub` field (or make `validate` `pub` as well?)

aws-s3-transfer-manager/src/operation/upload/checksum_strategy.rs
```diff
@@ -0,0 +1,283 @@
...
+    /// any `SHA` algorithm to do full object multipart uploads, and does not
+    /// allow `CRC64NVME` to do composite multipart uploads.
+    pub fn new(
```
landonxjames, line 148: Not entirely sure we want to expose this, seems like it could make the backwards compatibility story more difficult if we want to add more checksum related configuration in the future

landonxjames (approved):
Looks good, couple of comments about doc stuff and one question about whether an function should actually be `pub`.

aws-s3-transfer-manager/src/operation/upload/checksum_strategy.rs
graebm: Good call, let's get consistent.

If we moved it to types, it should be renamed like ~ChecksumStrategy~ -> `UploadChecksumStrategy` so users don't confuse it with types we add later to deal with checksum-validation-during-download. I put it under `upload::` because I saw some other operation specific types, and figured this would keep the name short and reduce confusion. When there's no precedent, I personally tend to group things by system vs type.

Here's a selection of things from the SDK:
```
aws_sdk_s3::
    config::
        endpoint::
            DefaultResolver
        http::
        Config
        Region
    operation::
        create_multipart_object::
            CreateMultipartObjectInput
    primitives::
        AggregatedBytes
        ByteStream
        SdkBody
    types::
        ChecksumAlgorithm
```

Here's a selection of things from the transfer manager
```
aws_s3_transfer_manager::
    config::
        Config
        StalledStreamProtectionConfig
        download::
            DownloadInput
            DownloadHandle
            ChunkOutput
            ChunkMetadata
            ObjectMetadata
            Body <-- organizationally like aws_sdk_s3::primitives::SdkBody ?
        upload::
            Uploadinput
            UploadHandle
            UploadOutput
            ChecksumStrategy <-- organizationally like aws_sdk_s3::types::ChecksumAlgorithm ?
    types::
        ConcurrencySetting <-- organizationally like aws_sdk_s3::config::StalledStreamProtectionConfig ?
        FailedDownload
        FailedMultipartUploadPolicy
        UploadFilter
```

aws-s3-transfer-manager/tests/upload_checksum_test.rs
graebm: Yeah, my Cargo.toml has a bunch of edits to use a local version of the SDK so I haven't committed it

aws-s3-transfer-manager/src/operation/upload/checksum_strategy.rs
graebm: I made an issue to deal with module organization across the codebase: https://github.com/awslabs/aws-s3-transfer-manager-rs/issues/88

Leaving `ChecksumStrategy` where it is for now...

aws-s3-transfer-manager/src/operation/upload/checksum_strategy.rs
graebm: I do believe we need **some** way for users to create a `ChecksumStrategy` from it's individual parts. I expect most users that need to set a strategy will use the `ChecksumStrategy::with_xyz()` constructors. The `ChecksumStrategy::new(...)` is for users gluing the Transfer Manager to some other system with its own checksum settings.

Possible approaches:
1) Add a `ChecksumStrategyBuilder`, remove  `ChecksumStrategy::new(...)`
2) Keep `ChecksumStrategy::new(...)`, and in the future if we need more fields add a builder, or an alternate new() function

I'm inclined to keep it simple for now, with just the `::new(...)` function for super advanced users that need it

landonxjames: I would be more inclined towards adding the `ChecksumStrategyBuilder` since thats what we do for everything in the SDK. But I'm not totally clear on what the backwards compatibility requirements are for the TM vs the actual SDK. 

Not opposed to keeping it simple for now and keeping open the possibility of adding a builder before GA

aws-s3-transfer-manager/src/operation/upload/handle.rs
graebm: Oh cool, didn't know about `unreachable!`

aws-s3-transfer-manager/src/operation/upload/builders.rs
graebm: Thanks, will do. This started out copy/pasted from the codegenned documentation, which is HTML style, and evolved into something much more hand-written

aws-s3-transfer-manager/src/operation/upload/builders.rs
graebm: Oh cool, I didn't know about this! Trying it out...

graebm: This is great. One of the things that pains me most about having getters AND setters AND public fields AND builders that also have getters and 2Xsetters, is the copy/pasted docs. All that copy/pasting encourages you to write terse docs with little detail. Or you write good docs with lots of details and it's very likely someone makes an edit, but fails to chase down and edit the 16 other places it's copy/pasted

aws-s3-transfer-manager/src/operation/upload/checksum_strategy.rs
```diff
@@ -0,0 +1,283 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+impl Default for ChecksumStrategy {
+    /// The transfer manager calculates a `CRC64NVME` full object checksum while uploading.
+    fn default() -> Self {
+        Self::with_calculated_crc64_nvme()
```
landonxjames, line 208: This came up elsewhere, but I realized the SEP indicates that the checksum should align with the default checksum of the underlying SDK. So in this case the TM should be defaulting to Crc32. I'm not opposed to us deviating like this, but we should make sure we are purposefully acknowledging it. 

From the SEP:
> The S3 Transfer Manager MUST defer to the underlying SDK's default checksum algorithm if checksum algorithm is not already provided in the request or client-level setting.

graebm: I'm assuming the SDK will default to CRC64NVME before the Transfer Manager goes 1.0

I'll make an issue to track this...

graebm: https://github.com/awslabs/aws-s3-transfer-manager-rs/issues/93

aws-s3-transfer-manager/src/operation/upload/checksum_strategy.rs
graebm: OK, I made a builder. Do please look it over. There are lots of little details I was unsure of, so I looked at some of the builders in `aws_config` for stylistic guidance. Stuff like:
- I named it `Builder` instead of `ChecksumStrategyBuilder`
- I didn't make any getters
- I didn't make any `.set_xyz(Option<XYZ>)`, only `.xyz(XYZ)`

aws-s3-transfer-manager/src/operation/upload/checksum_strategy.rs
graebm: `validate()` function is removed. Builder::build() does the checking now, and it's impossible to make an invalid `ChecksumStrategy`

aws-s3-transfer-manager/tests/upload_checksum_test.rs
graebm: Fixed, and fixed some other tests that were doing `const MEBIBYTE: ...`

aws-s3-transfer-manager/src/operation/upload/checksum_strategy.rs
graebm: this function is gone now, `Builder::build()` does validation

aws-s3-transfer-manager/tests/upload_checksum_test.rs
graebm: That's an excellent idea, but I don't want to spend a week adding the repo's very first integration test. Let's tackle this along with:
- https://github.com/awslabs/aws-s3-transfer-manager-rs/issues/25

Issue #87, opened 2025-01-14
Hedging breaks upload to S3 Express


graebm:
The S3 [Performance guidelines for Amazon S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-guidelines.html#optimizing-performance-guidelines-retry) recommend:
> Retry requests for latency-sensitive applications
> 
> Aggressive timeouts and retries help drive consistent latency. Given the large scale of Amazon S3, if the first request is slow, a retried request is likely to take a different path and quickly succeed. The AWS SDKs have configurable timeout and retry values that you can tune to the tolerances of your specific application.

And so the Transfer Manager uses `tower::hedge` to retry the slowest requests.

But this is leading to failed multipart uploads, because S3 Express responds with random ETags (vs ETags based MD5 in standard S3 buckets). So if the same Uploadpart is sent twice, it gets back two different ETags from S3. If S3 believes the 1st UploadPart is the "correct" one, but the Transfer Manager believes the 2nd UploadPart is the "correct" one, then the CompleteMultipartUpload will fail because the ETag for that particular part is mismatched.

S3 should **probably** fix this on their side (or at least update docs), but in the meantime we should probably turn off hedging for UploadPart, at least in S3 Express, or perhaps ensure the 1st request is cancelled before sending the 2nd? So we're sure both sides agree that the 2nd is canonical?

TingDaoK:
in c-s3, we "accidentally" bypass this issue since the algo we have for the hedge is using a hard-coded 0.7s and the average of the real response-to-first-byte time based on regular s3 data.

So, in s3 express, no request is getting hedged from c-s3, and we can see from the canary result, that c-s3 still have "better" performance comparing to rust-ts with hedge for s3 express.

So, one temporary solution here while waiting for s3 to fix on their side, we can disable hedging for s3 express, and it will not impact our performance based on the data we have.

aajtodd:
fixed in #89

PR #89, opened 2025-01-15
Skip hedge for s3 express

TingDaoK:
*Issue #, if available:*

https://github.com/awslabs/aws-s3-transfer-manager-rs/issues/87

*Description of changes:*

allow hedge builder to take a policy and customize the policy for upload to skip retry on s3 express buckets

### DIFF
--- a/aws-s3-transfer-manager/src/middleware/hedge.rs
+++ b/aws-s3-transfer-manager/src/middleware/hedge.rs
@@ -37,7 +37,7 @@ pub(crate) struct Builder<P> {
 impl<P> Builder<P> {
     pub(crate) fn new(policy: P) -> Self {
         Self {
-            policy: policy,
+            policy,
             latency_percentile: LATENCY_PERCENTILE,
             min_data_points: MIN_DATA_POINTS,
             period: PERIOD,
--- a/aws-s3-transfer-manager/src/middleware/hedge.rs
+++ b/aws-s3-transfer-manager/src/middleware/hedge.rs
@@ -20,13 +20,13 @@ const MIN_DATA_POINTS: u64 = 20;
 const PERIOD: Duration = Duration::from_secs(2);
 
 /*
-* During uploads, S3 recommends retrying the slowest 5% of requests for latency-sensitive applications,
-* as some requests can experience high time to first byte. If a slow part is hit near the end of the request,
-* the application may spend the last few seconds waiting for those final parts to complete, which can reduce overall
-* throughput. This layer is used to retry the slowest 5% of requests to improve performance.
-* Based on our experiments, this makes a significant difference for multipart upload use-cases and
-* does not have a noticeable impact for the Download.
-*/
+ * During uploads, S3 recommends retrying the slowest 5% of requests for latency-sensitive applications,
+ * as some requests can experience high time to first byte. If a slow part is hit near the end of the request,
+ * the application may spend the last few seconds waiting for those final parts to complete, which can reduce overall
+ * throughput. This layer is used to retry the slowest 5% of requests to improve performance.
+ * Based on our experiments, this makes a significant difference for multipart upload use-cases and
+ * does not have a noticeable impact for the Download.
+ */
 pub(crate) struct Builder<P> {
     policy: P,
     latency_percentile: f32,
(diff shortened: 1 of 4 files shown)


aws-s3-transfer-manager/src/middleware/hedge.rs
```diff
@@ -37,16 +37,6 @@ pub(crate) struct Builder<P> {
 #[derive(Debug, Clone, Default)]
 pub(crate) struct DefaultPolicy;
 
-impl<T: Clone> Policy<T> for DefaultPolicy {
```
aajtodd, line 40: If we aren't using `DefaultPolicy` anymore remove it and if we are we probably want to leave this trait implementation in place or else it can't function as intended.

aws-s3-transfer-manager/src/operation/upload.rs
```diff
@@ -177,6 +177,10 @@ async fn try_start_mpu_upload(
 fn new_context(handle: Arc<crate::client::Handle>, req: UploadInput) -> UploadContext {
     UploadContext {
         handle,
+        is_s3_express: match req.bucket() {
```
aajtodd, line 180: We shouldn't need to store this in the context, the context already has access to the input, you can just move the logic into `clone_request`

```rs
impl Policy<UploadPartRequest> for hedge::DefaultPolicy {
    fn clone_request(&self, req: &UploadPartRequest) -> Option<UploadPartRequest> {
       let is_s3_express = req.ctx.request.bucket().ends_with("--x-s3");
        if is_s3_express {
            None
        } else {
            Some(req.clone())
        }
    }

    fn can_retry(&self, req: &UploadPartRequest) -> bool {
        true
    }
}
```

`can_retry()` can just return true since (IIRC correctly anyway) it is used for retrying a hedged request whereas `clone_request()` is what will actually determine if the initial request is even hedged. 

Putting the flag in the context duplicates information that we already have.

aws-s3-transfer-manager/src/operation/upload/service.rs
```diff
@@ -25,6 +24,21 @@ pub(super) struct UploadPartRequest {
     pub(super) upload_id: String,
 }
 
+impl Policy<UploadPartRequest> for hedge::DefaultPolicy {
```
aajtodd, line 27: Maybe add a simple unit test for the policy or otherwise test that we aren't breaking s3 express.

aws-s3-transfer-manager/src/middleware/hedge.rs
TingDaoK: we are still using `DefaultPolicy`, the only implementation we need is for `UploadPartRequest` since we only do hedging for upload.

aws-s3-transfer-manager/src/operation/upload.rs
TingDaoK: I started with that...
I just fee the string comparison for each request needs to happen for each request seems to be redundant. So, I moved the flag to ctx, which will have only one string comparison for the whole upload.

aws-s3-transfer-manager/src/operation/upload/service.rs
TingDaoK: I am not sure what kind of unit test we will need here. we don't have tests for hedge at all.
I ran some benchmark manually and prove it works. 
I guess it will fall into the integration tests we are missing.

I could add some test to check the s3 express was detected correctly?

aws-s3-transfer-manager/src/operation/upload.rs
aajtodd: My 2 cents is this feels like a premature optimization. We're sending an entire second network request here potentially with the full input struct cloned. The string comparison here is in all likelihood negligible.

aws-s3-transfer-manager/src/operation/upload/service.rs
aajtodd: Yeah a simple unit test of the policy that checks `clone_request` for express and non-express buckets. The logic is so simple it seems unnecessary but I think it serves more for future us if we refactor the logic or change the policy we have a simple test to keep us honest and not regress.

aajtodd (approved):

PR #94, opened 2025-02-04
Let user provide checksums via PartStream

graebm:
**Issue #:**
- https://github.com/awslabs/aws-s3-transfer-manager-rs/issues/3

Give users the ability to provide their own checksum values **as** they stream their data (the previous PR just let users provide the full checksum value up front)

**Description of changes:**
- Users may provide a checksum value with each part
- Users may provide a full object checksum value after streaming is complete

**Possible future issues:**
Currently, `PartStream` is always sent as a multipart upload. In this PR, the user can provide part checksums, and final checksum and that's exactly how multipart upload works.

BUT In the future, we may want to optimize a small stream into a single PutObject. This would add some ambiguities:
- if user is doing SHA, we can't use their part checksums, because we have to send it as a single PutObject
- if user is doing CRC, and provided part checksums but no full object checksum, then we'll never send those user-provided-part-checksums to S3

In both these cases, the Transfer Manager **could** do some local validation as it combines parts into a single PutObject, validating the user-provided checksums at the part boundaries. But it bends the promise we made to the user, that we'd send their checksums all the way to S3 for validation

### DIFF
--- a/aws-s3-transfer-manager/src/io/stream.rs
+++ b/aws-s3-transfer-manager/src/io/stream.rs
@@ -188,18 +188,16 @@ impl PartData {
         }
     }
 
-    /// Create a new part, with a precalculated checksum
+    /// Set this part's checksum, if you've calculated it yourself
     /// (base64 encoding of the big-endian checksum value for this part's data
     /// using the algorithm specified in the [ChecksumStrategy](crate::operation::upload::ChecksumStrategy)).
-    pub fn with_checksum(
-        part_number: u64,
-        data: impl Into<Bytes>,
-        checksum: impl Into<String>,
-    ) -> Self {
-        Self {
-            checksum: Some(checksum.into()),
-            ..Self::new(part_number, data)
-        }
+    ///
+    /// If you don't set this, the Transfer Manager will calculate one
+    /// automatically, unless you've explicitly disabled checksum calculation
+    /// (see [ChecksumStrategy](crate::operation::upload::ChecksumStrategy)).
+    pub fn with_checksum(mut self, checksum: impl Into<String>) -> Self {
+        self.checksum = Some(checksum.into());
+        self
     }
 }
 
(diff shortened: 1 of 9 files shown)


aws-s3-transfer-manager/src/io/stream.rs
```diff
@@ -183,6 +184,21 @@ impl PartData {
...
             part_number,
             data: data.into(),
+            checksum: None,
+        }
+    }
+
+    /// Create a new part, with a precalculated checksum
+    /// (base64 encoding of the big-endian checksum value for this part's data
+    /// using the algorithm specified in the [ChecksumStrategy](crate::operation::upload::ChecksumStrategy)).
+    pub fn with_checksum(
```
aajtodd, line 194: **style**: Wondering if this should be `new_with_checksum(...)` or have `with_checksum(self, checksum: impl Into<String>) -> Self`. 

So either:

```rs
let part = PartData::new_with_checksum(part_number, data, checksum);
```

OR

```rs
let part = PartData::new(part_number, data).with_checksum(checksum);
```

aws-s3-transfer-manager/src/operation/upload/handle.rs
```diff
@@ -211,21 +213,37 @@ async fn complete_upload(handle: UploadHandle) -> Result<UploadOutput, crate::er
                 .set_sse_customer_key(handle.ctx.request.sse_customer_key.clone())
                 .set_sse_customer_key_md5(handle.ctx.request.sse_customer_key_md5.clone());
 
+            // check for user-provided full-object checksum...
             if let Some(checksum_strategy) = &handle.ctx.request.checksum_strategy {
-                // TODO(aws-s3-transfer-manager-rs#3): allow user to pass full-object checksum value via callback on PartStream
+                if checksum_strategy.type_if_multipart() == &ChecksumType::FullObject {
+                    // check whether it was passed via ChecksumStrategy
+                    let mut full_object_checksum: Option<String> =
```
aajtodd, line 220: **style**: You can avoid the `mut` variable here.

```rs
let full_object_checksum = match checksum_strategy.full_object_checksum() {
    Some(checksum) => Some(checksum.into()),
    None => mpu_data.part_reader.full_object_checksum().await
}
```

aajtodd (approved):
Minor suggestions, nothing blocking.

PR #95, opened 2025-02-04
Replace concurrency knob

aajtodd:
**Description of changes:**

Replaces/expands the existing explicit concurrency knob with a target throughput mode that calculates concurrency based on estimated throughput. 

* `ConcurrencyMode::Explicit` - weights every request the same and operates generally as it did before. 
* `ConcurrencyMode::Auto` - defaults to 10Gbps target throughput for now (i.e. `ConcurrencyMode::TargetThroughput(10 gbps)`). In the future I would like for this to be an adaptive mode, perhaps adjusting the token bucket refill rate based on measuring actual throughput. 
* `ConcurrencyMode::TargetThroughput` - estimates the throughput for the request taking the smaller of the estimated max per/connection throughput (~ 90 MB/s based on [S3 docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-design-patterns.html) and internal discussions) and the throughput estimated based on the payload size and S3 P50 latency. 

This new token bucket approach _should_ allow for more concurrency for smaller payloads and result in the optimal concurrency for large requests. In testing the # of inflight requests matched both the number you'd get from following S3's own docs linked above and my own testing in the past that showed ~138 requests would max out a 100 Gbps NIC.

**Questions**
* The `num_workers` function is used to set a few things still that really ought to come from somewhere else
    * The buffer layer in upload service
    * The channel size for the download body
    * Number of workers used for `upload_objects` and `download_objects`
* The min token cost for small payloads (i.e. control plane)?
* Should we allow configuring upload/download target throughput independently? Should they come out of different token buckets even?

**TODO**
* Buffer layer for upload was introduced as a way to make the service cloneable after we introduced the hedging layer. The buffer size is still based on the concurrency which is now dynamic in one or more concurrency modes. This should _probably_ be some arbitrarily large number since the concurrency layer is what actually limits max in-fight requests anyway?
* Remove/replace some existing unit tests
    * Unfortunately some of the existing unit tests are not as simple because `poll_ready` can't be relied on

### DIFF
--- a/aws-s3-transfer-manager/src/middleware/limit/concurrency/future.rs
+++ b/aws-s3-transfer-manager/src/middleware/limit/concurrency/future.rs
@@ -8,6 +8,8 @@ use crate::runtime::scheduler::{AcquirePermitFuture, OwnedWorkPermit};
 
 use futures_util::ready;
 use pin_project_lite::pin_project;
+use std::sync::atomic::{self, AtomicUsize};
+use std::sync::Arc;
 use std::{future::Future, task::Poll};
 use tower::Service;
 
@@ -18,6 +20,7 @@ pin_project! {
     {
         request: Option<Request>,
         inner: S,
+        inflight: Arc<AtomicUsize>,
         #[pin]
         state: State<S::Future>
     }
@@ -50,10 +53,12 @@ where
         inner: S,
         req: Request,
         permit_fut: AcquirePermitFuture,
+        inflight: Arc<AtomicUsize>,
     ) -> ResponseFuture<S, Request> {
         ResponseFuture {
             request: Some(req),
             inner,
+            inflight,
             state: State::AcquiringPermit { permit_fut },
         }
     }
@@ -75,6 +80,8 @@ where
                     match res {
                         Ok(_permit) => {
                             let req = this.request.take().expect("request set");
+                            let inflight = this.inflight.fetch_add(1, atomic::Ordering::SeqCst) + 1;
+                            tracing::trace!("in-flight requests: {inflight}");
                             let fut = this.inner.call(req);
                             this.state.set(State::Called { fut, _permit });
                         }
@@ -83,6 +90,8 @@ where
                 }
                 StateProj::Called { fut, .. } => {
                     let result = ready!(fut.poll(cx));
+                    let inflight = this.inflight.fetch_sub(1, atomic::Ordering::SeqCst);
+                    tracing::trace!("in-flight requests: {inflight}");
                     return Poll::Ready(result);
                 }
             }
--- a/aws-s3-transfer-manager/src/middleware/limit/concurrency/future.rs
+++ b/aws-s3-transfer-manager/src/middleware/limit/concurrency/future.rs
@@ -11,6 +11,7 @@ use pin_project_lite::pin_project;
 use std::sync::atomic::{self, AtomicUsize};
 use std::sync::Arc;
 use std::{future::Future, task::Poll};
+use tower::util::Oneshot;
 use tower::Service;
 
 pin_project! {
@@ -19,10 +20,10 @@ pin_project! {
         where S: Service<Request>
     {
         request: Option<Request>,
-        inner: S,
+        svc: Option<S>,
         inflight: Arc<AtomicUsize>,
         #[pin]
-        state: State<S::Future>
+        state: State<Oneshot<S, Request>>
     }
 }
 
@@ -57,7 +58,7 @@ where
     ) -> ResponseFuture<S, Request> {
         ResponseFuture {
             request: Some(req),
-            inner,
+            svc: Some(inner),
             inflight,
             state: State::AcquiringPermit { permit_fut },
         }
@@ -82,7 +83,10 @@ where
                             let req = this.request.take().expect("request set");
                             let inflight = this.inflight.fetch_add(1, atomic::Ordering::SeqCst) + 1;
                             tracing::trace!("in-flight requests: {inflight}");
-                            let fut = this.inner.call(req);
+                            let svc = this.svc.take().expect("service set");
+                            // NOTE: because the service was (1) never polled for readiness
+                            // originally and (2) also cloned, we need to ensure it's ready now before calling it.
+                            let fut = Oneshot::new(svc, req);
                             this.state.set(State::Called { fut, _permit });
                         }
                         Err(err) => return Poll::Ready(Err(err.into())),
--- a/aws-s3-transfer-manager/src/middleware/limit/concurrency/future.rs
+++ b/aws-s3-transfer-manager/src/middleware/limit/concurrency/future.rs
@@ -4,12 +4,10 @@
  */
 
 use crate::error;
-use crate::runtime::scheduler::{AcquirePermitFuture, OwnedWorkPermit};
+use crate::runtime::scheduler::{AcquirePermitFuture, OwnedWorkPermit, Scheduler};
 
 use futures_util::ready;
 use pin_project_lite::pin_project;
-use std::sync::atomic::{self, AtomicUsize};
-use std::sync::Arc;
 use std::{future::Future, task::Poll};
 use tower::util::Oneshot;
 use tower::Service;
@@ -21,7 +19,7 @@ pin_project! {
     {
         request: Option<Request>,
         svc: Option<S>,
-        inflight: Arc<AtomicUsize>,
+        scheduler: Scheduler,
         #[pin]
         state: State<Oneshot<S, Request>>
     }
@@ -37,6 +35,8 @@ pin_project! {
             fut: F,
             // retain until dropped when future completes
             _permit: OwnedWorkPermit,
+            // auto decrement inflight metrics on drop
+            _inflight: Inflight,
         },
         // Polling the future from [`Scheduler::acquire_permit`]
         AcquiringPermit {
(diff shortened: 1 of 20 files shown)


aws-s3-transfer-manager/src/runtime/token_bucket.rs
```diff
@@ -0,0 +1,245 @@
...
+    .max(1.0) as usize;
+
+    cmp::max(estimated_mbps, MIN_PAYLOAD_COST_TOKENS)
+}
+
+/// Estimate the throughput of a given request payload based on S3 latencies
+/// and max per/connection estimates.
+fn estimated_throughput(
+    payload_size: u64,
+    estimated_p50_latency_ms: usize,
```
graebm, line 175: trivial: it's so weird that Semaphore::new() takes usize, but acquire_many() takes u32. It leads to a crazy mix of u32, usize, u64, etc based on how a variable or const will eventually be used (i.e. passing time here as usize)?

Can we try and reduce the number of types in use? Like stick with u32 or u64 whenever possible, and just clamp/cast at the last moment where necessary?

aws-s3-transfer-manager/src/runtime/token_bucket.rs
```diff
@@ -0,0 +1,245 @@
...
+
+/// Estimated per/request max throughput S3 is capable of
+const S3_MAX_PER_REQUEST_THROUGHPUT_MB_PER_SEC: u64 = 90;
+
+/// Minimum concurrent requests at full throughput we want to support
+const MIN_CONCURRENT_REQUESTS: usize = 8;
+
+/// Min bucket size tokens
+const MIN_TOKENS: usize =
+    (S3_MAX_PER_REQUEST_THROUGHPUT_MB_PER_SEC as usize) * 8 * MIN_CONCURRENT_REQUESTS;
```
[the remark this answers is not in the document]
graebm: Ah, figured it out later that 1 token corresponds to 1 megabit of throughput. Add comment about that here. I guess it feels dump to make a const `TOKEN_COST_IN_MEGABITS = 1` and use it once on this line, where it's an unsafe divide if its value were ever tweaked

aws-s3-transfer-manager/src/middleware/limit/concurrency/future.rs
```diff
@@ -3,35 +3,102 @@
...
+                            // originally and (2) also cloned, we need to ensure it's ready now before calling it.
+                            let fut = Oneshot::new(svc, req);
+                            this.state.set(State::Called { fut, _permit });
+                        }
+                        Err(err) => return Poll::Ready(Err(err.into())),
+                    }
+                }
+                StateProj::Called { fut, .. } => {
+                    let result = ready!(fut.poll(cx));
+                    let inflight = this.inflight.fetch_sub(1, atomic::Ordering::SeqCst);
```
graebm, line 97: Is there any chance that this future never reaches this point, and this shared atomic is never decremented? Should there be some drop logic to work around this?

Is there any chance that this future is polled again after reporting that it's ready, and this shared atomic is decremented again? Should we move to some new Completion state to avoid this?

I genuinely don't know. I've always been a bit confused over what is guaranteed to happen, or never happen, or happen exactly once in the world of Rust futures

aws-s3-transfer-manager/src/runtime/scheduler.rs
```diff
@@ -3,79 +3,79 @@
...
     }
 }
 
+pub(crate) type PayloadEstimate = u64;
+
+/// The type of work to be done
+#[derive(Debug, Clone)]
+pub(crate) enum PermitType {
+    /// A network request to a control plane API with insignificant payload
+    ControlPlane,
```
graebm, line 64: trivial: maybe don't use these terms? HeadObject is data plane, but has no payload, so ?

graebm (commented):
a million dumb nitpicks, but I love it

aws-s3-transfer-manager/src/runtime/token_bucket.rs
```diff
@@ -0,0 +1,245 @@
...
+const AUTO_TARGET_THROUGHPUT_GIGABYTES_PER_SEC: u64 = ByteUnit::Gigabit.as_bytes_u64() * 10;
+
+/// Estimated P50 latency for S3
+const S3_P50_REQUEST_LATENCY_MS: usize = 30;
+
+/// Estimated per/request max throughput S3 is capable of
+const S3_MAX_PER_REQUEST_THROUGHPUT_MB_PER_SEC: u64 = 90;
+
+/// Minimum concurrent requests at full throughput we want to support
+const MIN_CONCURRENT_REQUESTS: usize = 8;
```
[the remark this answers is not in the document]
aajtodd: Yeah I wasn't sure when I started and it ended up not being terrible difficult to shoehorn explicit in that a whole separate implementation didn't seem necessary but I don't have a strong opinion at the moment. Any gut feeling or preference? The only issue with a separate token bucket is we likely introduce more branching per acquire _somewhere_.

aws-s3-transfer-manager/src/runtime/token_bucket.rs
aajtodd: Ughh yeah I don't understand why acquire takes a u32, I'm sure there's a reason but it's annoying. I'll look at what makes sense.

aws-s3-transfer-manager/src/middleware/limit/concurrency/future.rs
aajtodd: It's a good question, the documentation for [`poll`](https://doc.rust-lang.org/std/future/trait.Future.html#tymethod.poll) says you shouldn't poll futures once complete as they could panic, block forever, or cause other issues. 

This was rammed in last minute for my own sanity testing things, but I envision we'll want some sort of runtime/scheduler metrics that this gets tacked onto _somewhere_.

aws-s3-transfer-manager/src/runtime/scheduler.rs
aajtodd: I think it was suggested to just collapse them to a single variant `Network(Option<PayloadEstimate>)` or similar which covers both use cases.

aws-s3-transfer-manager/src/operation/download/service.rs
```diff
@@ -31,6 +32,19 @@ pub(super) struct DownloadChunkRequest {
...
 
+impl ProvidePayloadSize for DownloadChunkRequest {
+    fn payload_size(&self) -> u64 {
+        // we can't know the actual size by calling next_seq() as that would modify the
+        // state. Instead we give an estimate based on the current sequence.
+        let seq = self.ctx.current_seq();
+        let remaining = self.remaining.clone();
+        let part_size = self.ctx.handle.download_part_size_bytes();
+        let start = remaining.start() + ((seq - self.start_seq) * part_size);
+        let end_inclusive = cmp::min(start + part_size - 1, *remaining.end());
```
graebm, line 43: trivial: can we get a common helper for this math? It's duplicated in the `next_chunk()` function right below here.

aws-s3-transfer-manager/src/types.rs
```diff
@@ -20,17 +22,59 @@ pub enum PartSize {
...
+    pub fn new_gigabits_per_sec(gbps: u64) -> Self {
+        let target = Throughput::new_bytes_per_sec(gbps * ByteUnit::Gigabit.as_bytes_u64());
+        Self::new(target, target)
+    }
+
+    // TODO: we don't actually support limiting throughput by upload/download independently (yet), it will require updates to scheduler.
+
+    /// Create a new target throughput using the given upload and download
+    /// throughputs.
+    fn new(upload_target: Throughput, download_target: Throughput) -> Self {
```
graebm, line 60: trivial: when separate speeds are given for download and upload, download is usually cited first

graebm (approved):
mostly trivial feedback

ysaito1001 (approved):

aws-s3-transfer-manager/src/types.rs
```diff
@@ -20,17 +22,59 @@ pub enum PartSize {
     Target(u64),
 }
 
-/// The concurrency settings to use for a single upload or download request.
+/// The concurrency mode the client should use for executing requests.
 #[derive(Debug, Clone, Default)]
-pub enum ConcurrencySetting {
-    /// Automatically configure an optimal concurrency setting based on the execution environment.
+pub enum ConcurrencyMode {
+    /// Automatically configured concurrency based on the execution environment.
```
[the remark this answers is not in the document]
ysaito1001: Ah, good catch

aws-s3-transfer-manager/src/runtime/token_bucket.rs
```diff
@@ -0,0 +1,277 @@
...
+/// > Make one concurrent request for each 85-90 MB/s of desired network throughput
+///
+/// Applies to: ConcurrencyMode::TargetThroughput
+const S3_MAX_PER_REQUEST_THROUGHPUT: Throughput = Throughput::new_bytes_per_sec(90 * 1000 * 1000);
+
+/// Minimum concurrent requests at full throughput we want to support
+///
+/// Source: None, reasonable default
+/// Applies to: ConcurrencyMode::TargetThroughput
+const MIN_CONCURRENT_REQUESTS: u64 = 8;
```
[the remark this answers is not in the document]
aajtodd: Yeah interesting, well it would be relatively easy to rip this out and just let whatever they configure be the target but anything below 100 Mbps is going to cause issues as a single 5MB part can reach 90Mbps on a single connection (in theory anyway). I think we'd need `MIN_CONCURRENT_REQUESTS = 1` still based on current implementation. I'll leave it for now though.

PR #97, opened 2025-02-07
Set x-amz-checksum-type on CompleteMultipartUpload

graebm:
**Issue:**
Multipart upload is failing when a full object checksum is provided by the user

**Description of Changes:**

Both CreateMultipartUpload AND CompleteMultipartUpload have `x-amz-checksum-type` fields. Set that field the same in both requests.

Previously we only set it in CreateMultipartUpload, which works when you're letting S3 calculate the full-object checksum value, but doesn't work when you provide the full-object checksum value yourself.

### DIFF
--- a/aws-s3-transfer-manager/src/operation/upload/handle.rs
+++ b/aws-s3-transfer-manager/src/operation/upload/handle.rs
@@ -213,8 +213,12 @@ async fn complete_upload(handle: UploadHandle) -> Result<UploadOutput, crate::er
                 .set_sse_customer_key(handle.ctx.request.sse_customer_key.clone())
                 .set_sse_customer_key_md5(handle.ctx.request.sse_customer_key_md5.clone());
 
-            // check for user-provided full-object checksum...
             if let Some(checksum_strategy) = &handle.ctx.request.checksum_strategy {
+                // The Transfer Manager SEP recommends setting checksum-type on both
+                // CompleteMultipartUpload AND CreateMultipartUpload.
+                req = req.checksum_type(checksum_strategy.type_if_multipart().clone());
+
+                // check for user-provided full-object checksum...
                 if checksum_strategy.type_if_multipart() == &ChecksumType::FullObject {
                     // it might have been passed via ChecksumStrategy or PartStream
                     let full_object_checksum = match checksum_strategy.full_object_checksum() {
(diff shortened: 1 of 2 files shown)


ysaito1001 (approved):

aws-s3-transfer-manager/src/operation/upload/handle.rs
```diff
@@ -213,8 +213,12 @@ async fn complete_upload(handle: UploadHandle) -> Result<UploadOutput, crate::er
                 .set_sse_customer_key(handle.ctx.request.sse_customer_key.clone())
                 .set_sse_customer_key_md5(handle.ctx.request.sse_customer_key_md5.clone());
 
-            // check for user-provided full-object checksum...
             if let Some(checksum_strategy) = &handle.ctx.request.checksum_strategy {
+                // The Transfer Manager SEP recommends setting checksum-type on both
+                // CompleteMultipartUpload AND CreateMultipartUpload.
```
TingDaoK, line 218: you actually has to set it in both, otherwise, s3 will just error out.

graebm: So weird...
If you don't pass the actual checksum value, then you only need to set type in CreateMPU.
But yeah, you're right, it's required in both if you're passing the actual value.

graebm: removing snarky comments...

TingDaoK: > If the user provides a checksum for the object and the object is multipart uploaded then the Transfer Manager MUST set the x-amz-checksum-type header to "full_object" when calling CreateMultipartUpload and CompleteMultipartUpload

I guess it's only required to be set when user provides the checksum for full object. 

But, other cases, it's kind undefined.

I think it seems to work with S3 to set the type on complete MPU for other cases, but do we want to consider just move the type for complete MPU when user sets the full object checksum?

graebm: I tested, and it's fine to set both. Tested the full matrix of Composite vs FullObject and with-checksum-value vs let-s3-calculate-it

TingDaoK (approved):

Issue #62, opened 2024-10-07
Support bumping up the part size based on ContentLength


waahm7:
We should bump the part_size if the request size if large enough so that we don't run into 10_000 part limits.

waahm7:
I tried uploading a 100GB file, and it failed with max parts exceeded. So, maybe we just need to verify that it is working as expected.

PR #103, opened 2025-02-13
Fix bumping up part_size based on Content-Length

waahm7:
*Issue #, if available:*
#62 

*Description of changes:*
- There was a bug that we were doing normal divide and not ceil divide.

### DIFF
--- a/aws-s3-transfer-manager/tests/upload_checksum_test.rs
+++ b/aws-s3-transfer-manager/tests/upload_checksum_test.rs
@@ -20,7 +20,6 @@ use aws_sdk_s3::{
     types::{ChecksumAlgorithm, ChecksumType},
 };
 use aws_smithy_mocks_experimental::{mock, Rule, RuleMode};
-use aws_smithy_runtime::test_util::capture_test_logs::capture_test_logs;
 use bytes::Bytes;
 use pin_project_lite::pin_project;
 use test_common::mock_client_with_stubbed_http_client;
--- a/aws-s3-transfer-manager/tests/upload_checksum_test.rs
+++ b/aws-s3-transfer-manager/tests/upload_checksum_test.rs
@@ -521,8 +521,6 @@ impl TestConfig {
 }
 
 async fn run_test(config: TestConfig) -> UploadOutput {
-    let (_guard, _rx) = capture_test_logs();
-
     // This is the ChecksumStrategy we expect the Transfer Manager to use while sending requests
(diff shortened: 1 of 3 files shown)


aws-s3-transfer-manager/tests/upload_test.rs
```diff
@@ -171,3 +178,35 @@ async fn test_many_uploads_no_deadlock() {
...
+        .key("test-large-upload-part-size".to_string())
+        .body(InputStream::from_part_stream(stream))
+        .initiate()
+        .unwrap();
+
+    // actual object is empty but we should have bumped the part_size from size_hint
+    drop(tx);
+    handle.join().await.unwrap();
+
+    let expected_part_size = 10737419;
```
TingDaoK, line 208: instead of the hard coded number, maybe use `size_hint.div_ceil(MAX_PARTS)`

aws-s3-transfer-manager/tests/upload_checksum_test.rs
```diff
@@ -521,8 +521,6 @@ impl TestConfig {
 }
 
 async fn run_test(config: TestConfig) -> UploadOutput {
-    let (_guard, _rx) = capture_test_logs();
```
TingDaoK, line 524: remove the import as well

TingDaoK (approved):

aws-s3-transfer-manager/tests/upload_checksum_test.rs
waahm7: Thanks, updated.

aws-s3-transfer-manager/tests/upload_test.rs
TingDaoK: but the hard coded one doesn't make any sense. just it will be easy to read.

Checking against the expected logic is what we want for test

waahm7: > but the hard coded one doesn't make any sense, it looks like a weird number comes from nowhere.
I have added a comment

>  And checking against the expected logic is what we want for test

I think that we should test the expected logic with the expected value, to validate that the logic is correct. Copying the logic won't validate the logic because it will always be equal.

waahm7: However, I think this case is simple enough that we can go either way.

aajtodd (approved):

PR #104, opened 2025-02-17
Discovery e2e tests

TingDaoK:
*Issue #, if available:*

*Description of changes:*


aws-s3-transfer-manager/tests/e2e_transfer_test.rs
```diff
@@ -305,3 +304,89 @@ async fn test_upload_with_long_running_stream() {
...
+        .initiate()?;
+
+    let body = drain(&mut download_handle).await?;
+
+    assert_eq!(body.len(), expected_length);
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_object_download_range() {
```
waahm7, line 348: trivial/ debatable: It might be cleaner to split success/fail cases into two separate test_cases.

aws-s3-transfer-manager/tests/e2e_transfer_test.rs
```diff
@@ -305,3 +304,89 @@ async fn test_upload_with_long_running_stream() {
...
+        .key(object_key)
+        .initiate()
+        .unwrap();
+
+    let body = drain(&mut download_handle).await.unwrap();
+
+    assert_eq!(body.len(), 0);
+}
+
+async fn range_download_helper(
```
waahm7, line 327: This helper is only 3 lines and used at one place. Should we just get rid of it for now?

aws-s3-transfer-manager/tests/e2e_transfer_test.rs
```diff
@@ -17,9 +17,8 @@ use std::task::Poll;
 use std::time::Duration;
 use test_common::{drain, global_uuid_str};
 
-use tokio::time::Sleep;
-
 use aws_s3_transfer_manager::types::PartSize;
+use tokio::time::Sleep;
```
waahm7, line 21: Bad import? I don't see it used anywhere.

waahm7 (approved):

aws-s3-transfer-manager/tests/e2e_transfer_test.rs
```diff
@@ -305,3 +304,89 @@ async fn test_upload_with_long_running_stream() {
...
+            success_ranges[i],
+            success_expected_length[i],
+        )
+        .await
+        .unwrap();
+    }
+
+    for i in expect_fail_ranges {
+        // The fail case should error out.
+        range_download_helper(&tm, &bucket_name, object_key, i, 0)
```
aajtodd, line 388: No strong opinion but as an FYI you can setup a test with the [expectation of panic](https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html#testing-panics).

aws-s3-transfer-manager/tests/e2e_transfer_test.rs
```diff
@@ -305,3 +304,89 @@ async fn test_upload_with_long_running_stream() {
...
+
+#[tokio::test]
+async fn test_object_download_range() {
+    let (tm, _) = test_tm().await;
+    let (bucket_name, _) = get_bucket_names();
+
+    let object_key = "pre-existing-10MB";
+
+    let success_ranges = [
+        "bytes=0-10",              // Range is inclusive
```
aajtodd, line 355: You might consider modifying this to either include the test description or create a macro to make them separate tests making any failures easier to diagnose or dive into. 

Example taking your existing descriptions:

```rs
let success_ranges = [
    "bytes=0-10", "range is inclusive",
   ...
];
```

Then modify the `assert_eq` in `range_download_helper` to include the description:

```rs
async fn range_download_helper(
    tm: &aws_s3_transfer_manager::Client,
    bucket: &str,
    key: &str,
    range: &str,
    expected_length: usize,
    description: &str,
) -> Result<(), aws_s3_transfer_manager::error::Error> {
     ...

    assert_eq!(body.len(), expected_length, "desc: {}; range: {}", description, range);
    Ok(())
}
```

aajtodd (approved):

PR #106, opened 2025-02-26
Add release process

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,52 @@
+name: Publish to crates.io
+
+# on:
+#   push:
+#     tags:
+#       - 'v*'
+
+on:
+  pull_request: # for test
+
+env:
+  RUSTFLAGS: -Dwarnings
+  RUST_BACKTRACE: 1
+  # Change to specific Rust release to pin
+  rust_stable: stable
+  rust_nightly: nightly-2024-07-07
+  rust_clippy: '1.81'
+  # When updating this, also update relevant docs
+  rust_min: '1.81'
+  RELEASE_ROLE_ARN: ${{ secrets.RELEASE_ROLE_ARN }}
+  TOKEN_KEY_NAME: ${{ secrets.TOKEN_KEY_NAME }}
+  AWS_REGION: us-west-2
+
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ env.RELEASE_ROLE_ARN }}
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Get Cargo registry token
+        id: cargo-registry-token
+        run: CARGO_REGISTRY_TOKEN=$(aws secretsmanager get-secret-value --secret-id ${{ secrets.TOKEN_KEY_NAME }} --region ${{ env.AWS_REGION }}  | jq -r ".SecretString")
+
+      # - name: Publish
+      #   uses: katyo/publish-crates@v2
+      #   with:
+      #     registry-token: \${{ secrets.CARGO_REGISTRY_TOKEN }}
+      #     dry-run: true # for test
+      #     # Optional arguments:
+      #     # ignore-unpublished-changes: true


aws-s3-transfer-manager/Cargo.toml
```diff
@@ -44,7 +43,7 @@ fastrand = "2.1.1"
 futures-test = "0.3.30"
 md5 = "0.7.0"
 tempfile = "3.12.0"
-test-common = { path = "./test-common", version = "0.1.0" }
+test-common = { path = "./test-common" }
```
TingDaoK, line 46: hmmm, with version, cargo publish will fail https://github.com/rust-lang/cargo/issues/4242#issuecomment-2578428685
Without version, cargo deny will fail. https://github.com/EmbarkStudios/cargo-deny/issues/241
🤦‍♂️

landonxjames:
Not sure if you’ve come across this tool, but it might help tighten up your debug loop since you can run your actions locally: https://github.com/nektos/act

aws-s3-transfer-manager/Cargo.toml
aajtodd: Maybe refactor the test common stuff so it's not a separate project with it's own `Cargo.toml` and instead is just in the `tests/` directory. See https://doc.rust-lang.org/book/ch11-03-test-organization.html#submodules-in-integration-tests

.github/workflows/release.yml
```diff
@@ -0,0 +1,172 @@
...
+        run: |
+          echo "Publishing version ${{ env.TEST_VERSION }}"
+          echo "Dry run: ${{ env.TEST_DRY_RUN }}"
+
+      - name: Configure Git
+        run: |
+          git config --global user.name "GitHub Actions"
+          git config --global user.email "actions@github.com"
+
+      - name: Update version in Cargo.toml
```
aajtodd, line 75: Do we want to be auto updating versions and tagging in the release? Given we aren't likely to be releasing on a frequent cadence (compared to SDKs for example) I feel like we could probably get away with:

1. Manually updating the version and creating the appropriate tag (with/without a script depending on how many steps there are) and changelog entries. 
2. Add a pre-flight release check in here that the version we are trying to release doesn't already exist

aws-s3-transfer-manager/Cargo.toml
ysaito1001: Uh, that's a pain. Maybe it's time to fix `mock_client` in `aws_smithy_mocks_experimental` to properly stub out an HTTP client upstream. We have this `test-common` as a separate helper crate because `mock_client_with_stubbed_http_client` (a workaround for mock_client) are used both in unit tests and in integration tests.

If `mock_client` is fixed, then we can remove the `mock_client_with_stubbed_http_client` from `test-common` and move `drain` and `global_uuid_str` to a common module under `tests/common.rs`.

ysaito1001: Uh, that's a pain. Maybe it's time to fix `mock_client` in `aws_smithy_mocks_experimental` to properly stub out an HTTP client upstream. We have this `test-common` as a separate helper crate because `mock_client_with_stubbed_http_client` (a workaround for mock_client) are used both in unit tests and in integration tests.

If `mock_client` is fixed, then we can remove the `mock_client_with_stubbed_http_client` from `test-common`, move `drain` and `global_uuid_str` to a common module `tests/common.rs`, and remove the `test-common` crate.

ysaito1001: Maybe [allow-wildcard-paths](https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html?highlight=allow-wildcard-paths#the-allow-wildcard-paths-field-optional) can be used to unblock error from `cargo-deny`?

ysaito1001: Alternatively, maybe [allow-wildcard-paths](https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html?highlight=allow-wildcard-paths#the-allow-wildcard-paths-field-optional) can be used to unblock error from `cargo-deny`?

.github/workflows/release.yml
TingDaoK: I started with an action to run when the tag was cut, then I realize we need to update the version in code first, and make sure the version we cut in github matches the version we have in code, which is easy to get wrong, IMO.

aajtodd: Then a release shell script. I think having the code already staged for release will simplify this workflow (IMHO anyway).

TingDaoK: Just want to double check I got you, you mean we manually run a release script to make the commit ready for release (changing the version and add changlog), then after we merge the commit, we cut a release tag to trigger the publish to crate.io?

It's some extra steps and manually things to make sure we have the right version and steps to follow, but we can start from there.

In aws-crt-cpp, we have some like https://github.com/awslabs/aws-crt-cpp/blob/main/update-version.sh
which runs after the release is cut, then update the code -> make a commit -> delete the old tag -> cut new tag point to the commit.

aajtodd: Yeah was just trying to simplify a bit. I'll leave it up to you but in my head it seemed easier to:

1. Manually update the version, changelog via PR and then tag it
2. Trigger CI to release and specify the tag to release

.github/workflows/release.yml
```diff
@@ -0,0 +1,117 @@
+name: Publish to crates.io
+
+# on:
+#   push:
+#     tags:
+#       - 'v*'
+
+on:
+  pull_request: # Test trigger
```
aajtodd, line 9: Shouldn't this be on workflow dispatch? (i.e. manual)

.github/workflows/release.yml
```diff
@@ -0,0 +1,117 @@
+name: Publish to crates.io
+
+# on:
```
aajtodd, line 3: Remove unused/commented out code

.github/workflows/release.yml
```diff
@@ -0,0 +1,117 @@
...
+  # Change to specific Rust release to pin
+  rust_stable: stable
+  rust_nightly: nightly-2024-07-07
+  # When updating this, also update relevant docs
+  rust_min: '1.81'
+  RELEASE_ROLE_ARN: ${{ secrets.RELEASE_ROLE_ARN }}
+  TOKEN_KEY_NAME: ${{ secrets.TOKEN_KEY_NAME }}
+  AWS_REGION: us-west-2
+  # Test values
+  TEST_DRY_RUN: 'true'
```
aajtodd, line 23: simply `DRY_RUN` would work but also probably should be an input for workflow dispatch

.github/workflows/release.yml
```diff
@@ -0,0 +1,117 @@
...
+      - name: Verify tag matches Cargo.toml version
+        shell: bash
+        working-directory: ./aws-s3-transfer-manager/
+        run: |
+          # Determine the tag - use actual tag or a test tag
+          if [[ "${{ github.event_name }}" == "push" && "${{ github.ref_type }}" == "tag" ]]; then
+            # Extract tag from GitHub ref (e.g. refs/tags/v1.2.3 -> v1.2.3)
+            TAG="${{ github.ref_name }}"
+          else
+            # If not triggered by a tag push, use a test tag for testing
```
aajtodd, line 58: **correctness**: This workflow isn't configured to be triggered by a tag push

.github/workflows/release.yml
```diff
@@ -0,0 +1,117 @@
...
+          if [[ "$TAG_VERSION" != "$CARGO_VERSION" ]]; then
+            echo "::error::Version mismatch! Git tag ($TAG_VERSION) does not match Cargo.toml version ($CARGO_VERSION)"
+            exit 1
+          else
+            echo "✅ Version check passed: Git tag matches Cargo.toml version ($TAG_VERSION)"
+          fi
+
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ env.RELEASE_ROLE_ARN }}
```
aajtodd, line 82: What account is this pulling from?

new_release.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+
```
aajtodd, line 2: move to a `scripts` or `tools` directory instead of littering the root of the repository.

new_release.sh
```diff
@@ -0,0 +1,102 @@
...
+# Check arguments
+if [ $# -lt 2 ]; then
+    echo "Usage: $0 <new-tag-version> <release-notes>"
+    echo "Example: $0 v1.2.0 \"Fixed critical bugs and improved performance\""
+    exit 1
+fi
+
+NEW_TAG=$1
+RELEASE_NOTES=$2
```
aajtodd, line 28: We don't need this. Just put a `CHANGELOG.md` at the root of the repository and we'll manually craft the changelog as we go or when we're ready for release. I don't think we need separate `.changelog/release_**` entries.

aajtodd (commented):
IMO think we're still trying to do to much. Also I would switch to a workflow dispatch trigger that takes in the dry run parameter as well as the ref to release (which will normally be a tag but could be a commit sha for testing without a tag).

.github/workflows/release.yml
ysaito1001: yeah, and probably give a default input value to `true` so that it lets the user to opt-in for production release.

.github/workflows/release.yml
TingDaoK: I plan to use the commented out code instead of trigger on PR, which is only me testing it out.

I think when we cut the tag from github means we ready for release. I would like to automate this process, instead three different manual steps to publish

.github/workflows/release.yml
TingDaoK: the rust sdk production account, I agree we can move it to a separate account, but i still want one single place to host the token and rotate it, instead of copy the string around.

new_release.sh
TingDaoK: I prefer to have some formatted changelong and automate the process as much as we could. The change log will be something like:
```
# Release v0.11.2

**Release Date:** 2025-03-03

## Release Notes
test

## Changes since v0.11.1
- 9fd58f97: aws_base64_compute_encoded_len() is now exact, doesn't add 1 extra for null-terminator (#1188) (Michael Graeb)
```

I followed the boto/s3transfer for the structure https://github.com/boto/s3transfer/tree/develop/.changes

waahm7: +1 to manual changelog during Dev Preview because I would say that we are more likely to batch up breaking changes and then craft better release notes than a simple string during Dev Preview.

PR #108, opened 2025-03-04
Manual release

TingDaoK:
*Issue #, if available:*
- follow up on https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/106

*Description of changes:*
- Manually setup the change log, and trigger the release pipeline
- changed crate name from `aws-s3-transfer-manager` to `aws-sdk-s3-transfer-manager`


.github/workflows/release.yml
```diff
@@ -0,0 +1,103 @@
+name: Publish to crates.io
+
+on:
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Dry run mode (create branch but skip tag creation and use cargo publish --dry-run)'
+        required: false
+        type: boolean
+        default: true
```
waahm7, line 10: Trivial: I prefer a boolean that is false by default and that we can enable if required. Maybe reverse it to something like do_release or make_release or release_mode which will be false by default.

.github/workflows/release.yml
```diff
@@ -0,0 +1,103 @@
...
+        uses: dtolnay/rust-toolchain@stable
+        with:
+            toolchain: ${{ env.rust_stable }}
+
+      - name: Install cargo-nextest
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-nextest
+
+      - uses: Swatinem/rust-cache@v2
```
waahm7, line 46: Add a comment regarding why these are required.

.github/workflows/release.yml
```diff
@@ -0,0 +1,103 @@
...
+        run: |
+          # Get latest tag from all tags
+          LATEST_TAG=$(git describe --tags --abbrev=0)
+          echo "Latest tag: $LATEST_TAG"
+
+          # Extract version from tag (remove 'v' prefix)
+          TAG_VERSION="${LATEST_TAG#v}"
+
+          # Extract version from Cargo.toml
+          CARGO_VERSION=$(grep -m 1 '^version = ' ./aws-s3-transfer-manager/Cargo.toml | sed 's/version = "\(.*\)"/\1/')
```
waahm7, line 59: Do we need the update the folder name here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,103 @@
...
+        with:
+          role-to-assume: ${{ env.RELEASE_ROLE_ARN }}
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Get Cargo registry token
+        shell: bash
+        id: cargo-registry-token
+        run: |
+          CARGO_REGISTRY_TOKEN=$(aws secretsmanager get-secret-value --secret-id ${{ secrets.TOKEN_KEY_NAME }} --region ${{ env.AWS_REGION }} | jq -r ".SecretString")
+          echo "::add-mask::$CARGO_REGISTRY_TOKEN"
```
waahm7, line 82: Add a comment that if we don't do this, github will log this token.

.github/workflows/release.yml
```diff
@@ -0,0 +1,103 @@
...
+          echo "token=$CARGO_REGISTRY_TOKEN" >> $GITHUB_OUTPUT
+
+      - name: Yank previous version (for testing token)
+        shell: bash
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ steps.cargo-registry-token.outputs.token }}
+        run: |
+          # This version has already been yanked. This command succeeds if we have a token with permission to yank the crate.
+          cargo yank aws-sigv4 --version 0.55.0
```
waahm7, line 91: What? Why are we trying to yank aws-sigv4? Can we just try to publish and it will work if token is good.

TingDaoK: https://github.com/smithy-lang/smithy-rs/blob/main/.github/workflows/release.yml#L239-L241
Make sure we are not getting some random token from other account

.github/workflows/release.yml
TingDaoK: good catch, thanks

.github/workflows/release.yml
TingDaoK: well, I honest don't know, I just copy/paste from ci.yml, thought that we need those initial steps to set up the rust envrionment

.github/workflows/release.yml
waahm7: >  Make sure we are not getting some random token from other account

Wouldn't cargo publish check the token and if the token is good, then publish will succeed otherwise it will fail?

TingDaoK: I mean if the token is from some other account (Like i can create an account and try to publish this), I assume the publish will also succeed.

Just some extra testing steps to make sure we are on the right track.

waahm7: If we publish the first version of transfer manager crate using token X, I don't think token Y will result in a successful publish unless it has the authorization. If it has the authorization to publish our crate, it will probably have the authorization to yank aws-sigv4.

ysaito1001: Yeah, an attempt to `aws-sigv4` version 0.55.0 is a sanity check that we have a valid crates.io token (it's an idempotent operation BTW since the crate of that version has already been yanked and doing so repeatedly is safe).
That setup makes most sense where releases occur every day and we cannot afford to fail with an invalid crates.io token; the risk of having an invalid token becomes high after we've just rotated it to a new token, and having this sanity check is useful. If we manually publish one crate, however, we can probably afford to fail with an invalid token, and simply replace it with a new one.

Also, if we source crates.io token from smithy-rs, we probably don't need to have this check, since the sanity check has already been done in on the smithy-rs side.

ysaito1001: Yeah, an attempt to `aws-sigv4` version 0.55.0 is a sanity check that we have a valid crates.io token (it's an idempotent operation BTW since the crate of that version has already been yanked and doing so repeatedly is safe).
That setup makes most sense where releases occur every day and we cannot afford to fail with an invalid crates.io token; the risk of having an invalid token becomes usually high after we've just rotated it to a new token, and having this sanity check is useful. If we manually publish one crate, however, we can probably afford to fail with an invalid token, and simply replace it with a new one.

Also, if we source crates.io token from smithy-rs, this check is probably not strictly necessary, since the sanity check has already been done on the smithy-rs side.

TingDaoK: IMHO, since it's idempotent operation, why not keep it? There is basically no cost to perform this extra check.

ysaito1001: Because in this particular setting, it's easier to ask for forgiveness than permission, as in Python. Especially if the token is coming from a known good source.

aajtodd (approved):
I like this better (for now anyway). Looks good overall.

Fix and ship.

PR #111, opened 2025-03-12
use default HTTPS client from SDK crate and cleanup http-02 usage

aajtodd:
**Description of changes:**

Stop overriding the HTTP client and use the default from the SDK crate which as of [latest release](https://github.com/awslabs/aws-sdk-rust/discussions/1257) is the same as what we were setting. Also cleaned up internal usage of legacy http-02 crate.

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/http.rs
+++ b/aws-sdk-s3-transfer-manager/src/http.rs
@@ -2,14 +2,5 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
-use aws_smithy_experimental::hyper_1_0::{CryptoMode, HyperClientBuilder};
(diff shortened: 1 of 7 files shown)


aws-sdk-s3-transfer-manager/src/http.rs
```diff
@@ -2,14 +2,5 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
waahm7, line 2: Should we just remove this file for now?

waahm7 (approved):

ysaito1001 (approved):

aws-sdk-s3-transfer-manager/src/http.rs
aajtodd: It still contains the needed module declaration (line 6) `pub(crate) mod header` for the HTTP header utils

PR #112, opened 2025-03-24
Fix DownloadObjects for S3Express

waahm7:
*Issue #, if available:*

*Description of changes:*
Our current benchmark for S3Express download_objects is failing with the following:
```

  thread 'main' panicked at src/main.rs:49:17:
  child operation failed

  Caused by:
      0: service error
      1: unhandled error (InvalidRequest)
      2: Error { code: "InvalidRequest", message: "This bucket does not support a prefix that does not end in a delimiter. Specify a prefix path ending with a delimiter and try again.", aws_request_id: "01b629e01a000195c918446e0509a490957f08b1", s3_extend
  ed_request_id: "HT0YP" }
```
Make sure that prefix ends with delimiter for ListObjects.

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/operation/download_objects/list_objects.rs
+++ b/aws-sdk-s3-transfer-manager/src/operation/download_objects/list_objects.rs
@@ -13,6 +13,8 @@ use aws_smithy_runtime_api::http::Response;
 use std::mem;
 use tracing::Instrument;
 
+use crate::operation::DEFAULT_DELIMITER;
+
 use super::DownloadObjectsContext;
 
 /// Custom paginator for `ListObjectsV2` operation that handles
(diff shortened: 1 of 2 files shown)


aws-sdk-s3-transfer-manager/src/operation/download_objects/list_objects.rs
```diff
@@ -101,7 +103,14 @@ impl State {
 
 impl ListObjectsPaginator {
     fn new(context: DownloadObjectsContext) -> Self {
-        let prefix = context.state.input.key_prefix.to_owned();
+        let mut prefix = context.state.input.key_prefix.to_owned();
+        let delimiter = context.state.input.delimiter().unwrap_or(DEFAULT_DELIMITER);
+        // S3Express requires that prefix must end with delimiter
+        if let Some(prefix) = prefix.as_mut() {
```
aajtodd, line 109: Shouldn't we only do this for S3 express buckets (and we have a function to test if a bucket is s3 express don't we)?

waahm7: I thought it won't hurt for s3 buckets as well but this is not the correct solution as prefix may or may not end with delimiter. It will have different output depending upon the bucket structure like if we have something like
```
prefixFile1
prefixFile2
prefix/file3
prefix/file4
```. 
I am looking more into it, maybe we just need to fix it in the benchmark runner.

aajtodd: Yeah I'm a little worried about overriding the prefix by default especially since it doesn't have to end in a delimiter (for regular buckets anyway). You _could_ argue this is just bad input and should be fixed by caller (i.e. the runner as you suggested).

waahm7: @aajtodd Thanks, yeah. I also discussed it with Dmitriy and will fix it in the runner.

PR #114, opened 2025-08-19
Validate UploadPart requests

ysaito1001:
*Description of changes:*
This PR adds the following validation to `UploadPart` requests:
- The total number of `UploadPart` requests must match the expected number of parts ([code change](https://github.com/awslabs/aws-s3-transfer-manager-rs/compare/ysaito/validate-upload-part-request?expand=1#diff-19c08f62e1609d29de38ce208ca94ab679812552d302b5895f15ac797b497c2aR195-R204))
- The content length of each `UploadPart` request must match the expected part size (except for the last part, which may be smaller than the target part size) ([code change](https://github.com/awslabs/aws-s3-transfer-manager-rs/compare/ysaito/validate-upload-part-request?expand=1#diff-03378f35eac44d2e7ef4657ae211c1805c0e071d52dcd8d5c8d1e93cc980aadaR225-R238))
- The part number assigned to each `UploadPart` request should align with the starting offset of the source (applicable to bytes and file-based) ([BytesPartReader](https://github.com/awslabs/aws-s3-transfer-manager-rs/compare/ysaito/validate-upload-part-request?expand=1#diff-3edc995013d58106e512d7ece68c317c9f11d1144a0c421db74e8a36f3ed55d4R156-R163),  [PathBodyPartReader](https://github.com/awslabs/aws-s3-transfer-manager-rs/compare/ysaito/validate-upload-part-request?expand=1#diff-3edc995013d58106e512d7ece68c317c9f11d1144a0c421db74e8a36f3ed55d4R235-R241))

Any validation error above will result in returning an `Err`, terminating the upload operation.

---

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/error.rs
+++ b/aws-sdk-s3-transfer-manager/src/error.rs
@@ -48,6 +48,10 @@ pub enum ErrorKind {
     /// The operation is being canceled because the user explicitly called `.abort` on the handle,
     /// or a child operation failed with the abort policy.
     OperationCancelled,
+
+    /// Request/response validation failed (e.g., the total number of UploadPart requests sent
+    /// did not match the expected number of parts)
+    ValidationFailed,
 }
 
 impl Error {
@@ -80,6 +84,7 @@ impl fmt::Display for Error {
             ErrorKind::NotFound => write!(f, "resource not found"),
             ErrorKind::ChildOperationFailed => write!(f, "child operation failed"),
             ErrorKind::OperationCancelled => write!(f, "operation cancelled"),
+            ErrorKind::ValidationFailed => write!(f, "validation failed"),
         }
     }
 }
--- a/aws-sdk-s3-transfer-manager/src/operation/upload/service.rs
+++ b/aws-sdk-s3-transfer-manager/src/operation/upload/service.rs
@@ -222,6 +222,20 @@ pub(super) async fn read_body(
         .instrument(tracing::debug_span!("read-upload-body"))
         .await?
     {
+        // The content length of each UploadPart request matches the expected part size
+        if part_reader.part_size() != part_data.data.remaining()
+            && part_data.is_last() == Some(false)
+        {
+            return Err(error::Error::new(
+                error::ErrorKind::ValidationFailed,
+                format!(
+                    "upload part size mismatch for non-last part {}: configured part size {}, got data size {}",
+                    part_data.part_number,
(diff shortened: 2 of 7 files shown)


aws-sdk-s3-transfer-manager/src/error.rs
```diff
@@ -48,6 +48,10 @@ pub enum ErrorKind {
     /// The operation is being canceled because the user explicitly called `.abort` on the handle,
     /// or a child operation failed with the abort policy.
     OperationCancelled,
+
+    /// Request/response validation failed (e.g., the total number of UploadPart requests sent
+    /// did not match the expected number of parts)
+    ValidationFailed,
```
aajtodd, line 54: This is probably ok though you could maybe roll this up into `ChunkFailed` if I squint.

aws-sdk-s3-transfer-manager/src/operation/upload/service.rs
```diff
@@ -222,6 +222,20 @@ pub(super) async fn read_body(
         .instrument(tracing::debug_span!("read-upload-body"))
         .await?
     {
+        // The content length of each UploadPart request matches the expected part size
+        if part_reader.part_size() != part_data.data.remaining()
+            && part_data.is_last() == Some(false)
```
aajtodd, line 227: What about a stream that can't tell you if it's the last part? The new method indicates this is a possibility.

aajtodd (commented):
One question, otherwise looks good.

aws-sdk-s3-transfer-manager/src/operation/upload/service.rs
ysaito1001: I suppose we're more lenient in streaming cases in that regard (i.e. the validation won't be triggered). The second guard will become true for cases where `part_data.is_last` can be definitively set to a `Some(...)` variant.; it will always be `None` for streaming cases.

Makes me think whether there is a way to be more strict about streaming cases. Updating the second guard to be `part_data.is_last() == Some(false) || part_data.is_last().is_none()` may trigger validation failure for a streaming case inadvertently where a part is legitimately smaller than configured part size 🤔

aws-sdk-s3-transfer-manager/src/error.rs
ysaito1001: Updated in 4b7b9c1, just to avoid increasing the number of variants when we can.

aajtodd (approved):

PR #121, opened 2025-09-15
Initial metrics implementation

landonxjames:
*Issue #, if available:*

*Description of changes:*
* Add `Gauge`, `Counter`, and `Histogram` implementations
* Add various new types for aggregating metrics
* Refactor existing `metrics.rs` into smaller sub-modules
* Add instrumentation to Upload and Download operations

Work still to be done:
* Instrument `ByteStream` for throughput calculation
* More scheduler and concurrency metrics
* Latency tracking
* Middleware for metrics

And when all of that is done I need to benchmark the `feature/metrics` branch against the existing `main` branch and see if the new metrics introduce enough of a performance impact (even when disabled) to justify hiding them all behind a cargo feature flag. 


aws-sdk-s3-transfer-manager/src/metrics/instruments.rs
```diff
@@ -0,0 +1,243 @@
...
+    pub fn value(&self) -> u64 {
+        self.value.load(Ordering::Relaxed)
+    }
+}
+
+/// A value that can increase or decrease over time.
+/// Minimum value is 0.
+#[derive(Debug, Clone, Default)]
+pub struct Gauge {
+    value: Arc<AtomicU64>,
```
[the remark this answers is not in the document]
landonxjames: Lol I actually stole the inner type of `Gauge` from your doc on this. We could do something like OpenTelemetry where the Gauge wraps a generic so we could use any number type: https://github.com/open-telemetry/opentelemetry-rust/blob/b7ff11b50838f1d45c08b3eeabf4bf573415b325/opentelemetry/src/metrics/instruments/gauge.rs#L14

Or it could be an `f64`, it would save the awkward clamping logic.

aws-sdk-s3-transfer-manager/src/metrics/instruments.rs
```diff
@@ -0,0 +1,243 @@
...
+        } else {
+            Self::new()
+        }
+    }
+}
+
+/// Default bucket boundaries for latency measurements in seconds.
+fn default_latency_buckets() -> Vec<f64> {
+    vec![
+        0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0,
```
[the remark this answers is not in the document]
landonxjames: No, I just picked a pattern that seemed reasonable (10%, 25%, 50%, next order of magnitude) and stuck with it. I should have removed the `0.005` though, that was left over from a more granular initial approach. Open to suggestions if you have experience on a better approach.

aws-sdk-s3-transfer-manager/src/client.rs
```diff
@@ -60,11 +62,29 @@ impl Handle {
     }
 }
 
+impl Drop for Handle {
+    fn drop(&mut self) {
+        // Log final metrics summary when the client is dropped
+        tracing::debug!(
```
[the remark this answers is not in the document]
landonxjames: This was mostly for my testing to make sure things were getting piped through correctly, will delete.

aws-sdk-s3-transfer-manager/src/operation/download/handle.rs
```diff
@@ -74,6 +77,38 @@ impl DownloadHandle {
...
+            }
+        }
+
+        if has_error {
+            self.handle.metrics.increment_transfers_failed();
+            Err(error::from_kind(ErrorKind::RuntimeError)(
+                "One or more download tasks failed",
+            ))
+        } else {
+            self.handle.metrics.increment_transfers_completed();
```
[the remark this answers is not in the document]
landonxjames: Yes, I initially wanted to do the metrics aggregation on `Drop` so it didn't require any explicit user action, but I had trouble getting that to work since it isn't async. Another option we could consider is moving this logic into each of the `orchestrate` functions, which shouldn't require anything from the user

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
```diff
@@ -0,0 +1,638 @@
...
+
+    /// Get the maximum throughput observed.
+    pub(crate) fn max(&self) -> Throughput {
+        let bps_scaled = self.max_throughput_bps.load(Ordering::Relaxed);
+        Throughput::new_bytes_per_sec(bps_scaled / 1000)
+    }
+
+    /// Get the average throughput.
+    pub(crate) fn avg(&self) -> Throughput {
+        let bps_scaled = self.avg_throughput_bps.load(Ordering::Relaxed);
```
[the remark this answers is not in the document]
landonxjames: This was mostly a placeholder, more granular throughput metrics (probably by instrumenting the Body) are coming in phase two. Should have thrown a `todo!` in instead of an oversimplified impl though. Changed it to calculate a cumulative avg in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/17156457f169455ecc5213a4e82f62a01c9a6065

aws-sdk-s3-transfer-manager/src/metrics/instruments.rs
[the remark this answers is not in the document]
landonxjames: Updated to `f64` in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/17156457f169455ecc5213a4e82f62a01c9a6065. Note that the implementation is a little weird. There is no `AtomicF64`, so I continued using the `AtomicU64` to store the bytes and transforming them to `f64` for operations/returning to the user. This should all be stack based (and hopefully mostly compiled away) so I expect it to be fast, but at some point probably worth benching against a `Mutex<f64>` based impl.

aws-sdk-s3-transfer-manager/src/operation/download.rs
```diff
@@ -219,15 +222,22 @@ fn handle_discovery_chunk(
...
             let chunk = AggregatedBytes::from_byte_stream(stream)
                 .await
-                .map(|aggregated| ChunkOutput {
-                    seq,
-                    data: aggregated,
-                    metadata: metadata.expect("chunk metadata is available"),
+                .map(|aggregated| {
+                    // Track bytes transferred
+                    let bytes_len = aggregated.remaining() as u64;
+                    handle.metrics.add_bytes_transferred(bytes_len);
```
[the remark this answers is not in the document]
landonxjames: Moved into `AggregatedBytes` in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/17156457f169455ecc5213a4e82f62a01c9a6065

aws-sdk-s3-transfer-manager/src/operation/download/service.rs
```diff
@@ -136,6 +137,10 @@ async fn download_specific_chunk(
                            error::chunk_failed(ChunkId::Download(seq), err)
                         })?;
 
+                    // Track bytes transferred
+                    let bytes_len = body.remaining() as u64;
+                    ctx.handle.metrics.add_bytes_transferred(bytes_len);
```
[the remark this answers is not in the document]
landonxjames: Moved into `AggregatedBytes` in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/17156457f169455ecc5213a4e82f62a01c9a6065

aws-sdk-s3-transfer-manager/src/client.rs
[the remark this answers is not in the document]
landonxjames: Removed in Moved into `AggregatedBytes` in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/17156457f169455ecc5213a4e82f62a01c9a6065

aws-sdk-s3-transfer-manager/src/operation/download.rs
```diff
@@ -287,4 +301,92 @@ impl DownloadContext {
     fn bucket_type(&self) -> BucketType {
         self.state.bucket_type
     }
+
+    /// Returns the transfer metrics for this download
+    #[allow(unused)]
+    fn metrics(&self) -> &Arc<TransferMetrics> {
```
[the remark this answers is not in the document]
landonxjames: Updated to just return the `Arc` in Moved into `AggregatedBytes` in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/17156457f169455ecc5213a4e82f62a01c9a6065

aws-sdk-s3-transfer-manager/src/metrics/instruments.rs
```diff
@@ -0,0 +1,243 @@
...
+struct HistogramInner {
+    count: u64,
+    sum: f64,
+    min: f64,
+    max: f64,
+    bucket_bounds: Vec<f64>,
+    bucket_counts: Vec<u64>,
+}
+
+impl Histogram {
```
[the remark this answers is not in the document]
landonxjames: Added ability to get arbitrary quantile and the specific quantiles you suggested in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/e5bd769c105d61dd083b30d1df16665dced23bca

aws-sdk-s3-transfer-manager/src/operation/download/handle.rs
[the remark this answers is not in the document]
landonxjames: Moved this tracking out of `join` in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/53ff51062ba41cec49dd1532fb48d94ab728d429. Now failure is tracked at each failure point and the actual metric is updated in an `impl Drop for DownloadHandle` block.

aws-sdk-s3-transfer-manager/src/operation/download/handle.rs
```diff
@@ -74,6 +77,38 @@ impl DownloadHandle {
         tasks.abort_all();
         while (tasks.join_next().await).is_some() {}
     }
+
+    /// Wait for the download to complete and track completion metrics.
+    pub async fn join(self) -> Result<(), error::Error> {
```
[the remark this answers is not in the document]
landonxjames: Added method to get these from `DownloadHandle` in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/53ff51062ba41cec49dd1532fb48d94ab728d429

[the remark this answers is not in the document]
landonxjames: I ended up moving the metrics collection out of `join` in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/53ff51062ba41cec49dd1532fb48d94ab728d429. But I still think it makes sense to keep it around, especially since I updated it to aggregate failures and report all of them. Seems like that could be helpful for user debugging.

aws-sdk-s3-transfer-manager/src/metrics/instruments.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+}
+
+impl Default for Histogram {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+//TODO: maybe this should be a try_clone and it should fail if the mutex is poisoned?
```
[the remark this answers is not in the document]
landonxjames: Removed the custom `Clone` in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/d67cd514f4eb8213eead312cfa57814e157b002c

aws-sdk-s3-transfer-manager/src/metrics/instruments.rs
```diff
@@ -0,0 +1,243 @@
...
+}
+
+impl Histogram {
+    /// Create a new histogram with default buckets for latency measurements.
+    pub fn new() -> Self {
+        Self::with_buckets(default_latency_buckets())
+    }
+
+    /// Create a new histogram with custom bucket boundaries.
+    pub fn with_buckets(bucket_bounds: Vec<f64>) -> Self {
```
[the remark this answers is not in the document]
landonxjames: Updated to take `impl IntoIterator<Item = f64>` in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/1749c876780f228070ab787ed141fcd8bd408a01

aws-sdk-s3-transfer-manager/src/metrics/instruments.rs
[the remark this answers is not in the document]
landonxjames: Looking into this more, I was wrong about a lot of it hopefully being compiled away. 

Playground for the `f64` version: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=5b44786a073aabd3d24b393211c79295

Produces this ASM for the `increment`:
```asm
playground::Gauge::increment:
	subq	$120, %rsp
	movq	%rdi, 8(%rsp)
	movsd	%xmm0, (%rsp)
	movq	%rdi, 56(%rsp)
	movsd	%xmm0, 64(%rsp)
	callq	*<alloc::sync::Arc<T,A> as core::ops::deref::Deref>::deref@GOTPCREL(%rip)
	movq	%rax, %rdi
	xorl	%esi, %esi
	callq	core::sync::atomic::AtomicU64::load
	movq	%rax, %rdi
	callq	core::num::<impl u64>::to_be_bytes
	movq	%rax, 72(%rsp)
	movq	72(%rsp), %rax
	movq	%rax, 40(%rsp)
	movq	40(%rsp), %rax
	movq	%rax, 80(%rsp)
	movq	80(%rsp), %rdi
	callq	core::f64::<impl f64>::from_be_bytes
	movsd	(%rsp), %xmm1
	movq	8(%rsp), %rdi
	movsd	%xmm0, 32(%rsp)
	movsd	%xmm0, 88(%rsp)
	addsd	%xmm1, %xmm0
	movsd	%xmm0, 16(%rsp)
	movsd	%xmm0, 96(%rsp)
	callq	*<alloc::sync::Arc<T,A> as core::ops::deref::Deref>::deref@GOTPCREL(%rip)
	movsd	16(%rsp), %xmm0
	movq	%rax, 24(%rsp)
	callq	core::f64::<impl f64>::to_be_bytes
	movq	%rax, 104(%rsp)
	movq	104(%rsp), %rax
	movq	%rax, 48(%rsp)
	movq	48(%rsp), %rax
	movq	%rax, 112(%rsp)
	movq	112(%rsp), %rdi
	callq	core::num::<impl u64>::from_be_bytes
	movq	24(%rsp), %rdi
	movq	%rax, %rsi
	xorl	%edx, %edx
	callq	core::sync::atomic::AtomicU64::store
	movsd	32(%rsp), %xmm0
	addq	$120, %rsp
	retq
```

Playground for the for the original `u64` version: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e898664babe0cf76dfccba27dc4f18a1

Produces this ASM for `increment` (which is much shorter):

```asm
playground::Gauge::increment:
	subq	$40, %rsp
	movq	%rsi, 8(%rsp)
	movq	%rdi, 24(%rsp)
	movq	%rsi, 32(%rsp)
	callq	*<alloc::sync::Arc<T,A> as core::ops::deref::Deref>::deref@GOTPCREL(%rip)
	movq	8(%rsp), %rsi
	movq	%rax, %rdi
	xorl	%edx, %edx
	callq	core::sync::atomic::AtomicU64::fetch_add
	movq	8(%rsp), %rsi
	addq	%rsi, %rax
	movq	%rax, 16(%rsp)
	setb	%al
	jb	.LBB15_2
	movq	16(%rsp), %rax
	addq	$40, %rsp
	retq
```

Playground for a `Mutex` based version that naively unwraps the result from `.lock()`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=cbc1c51051358a1ec4fbf87d08e950e8

Produces this ASM for `increment` (that splits the difference in length):

```asm
playground::Gauge::increment:
	subq	$184, %rsp
	movsd	%xmm0, 56(%rsp)
	movq	%rdi, 104(%rsp)
	movsd	%xmm0, 112(%rsp)
	callq	*<alloc::sync::Arc<T,A> as core::ops::deref::Deref>::deref@GOTPCREL(%rip)
	movq	%rax, %rsi
	leaq	80(%rsp), %rdi
	callq	*std::sync::poison::mutex::Mutex<T>::lock@GOTPCREL(%rip)
	testq	$1, 80(%rsp)
	je	.LBB24_6
	movq	88(%rsp), %rcx
	movb	96(%rsp), %al
	movq	%rcx, 136(%rsp)
	movb	%al, 144(%rsp)
	leaq	.Lanon.b599ee7fd3c786799892909dac0b871d.23(%rip), %rdi
	leaq	.Lanon.b599ee7fd3c786799892909dac0b871d.22(%rip), %rcx
	leaq	.Lanon.b599ee7fd3c786799892909dac0b871d.32(%rip), %r8
	movq	core::result::unwrap_failed@GOTPCREL(%rip), %rax
	movl	$43, %esi
	leaq	136(%rsp), %rdx
	callq	*%rax
	jmp	.LBB24_3
	movq	%rax, %rcx
	movl	%edx, %eax
	movq	%rcx, 168(%rsp)
	movl	%eax, 176(%rsp)
	movq	core::ptr::drop_in_place<std::sync::poison::PoisonError<std::sync::poison::mutex::MutexGuard<f64>>>@GOTPCREL(%rip), %rax
	leaq	136(%rsp), %rdi
	callq	*%rax
	jmp	.LBB24_5
```

Haven't benchmarked anything yet, so no idea what difference this makes in practice, but the `f64` casting impl has a lot more instructions (~2.5x as many). There is probably a better way to do this, I'll poke at it, but probably not in this PR.

aws-sdk-s3-transfer-manager/src/io/aggregated_bytes.rs
```diff
@@ -45,15 +46,24 @@ impl AggregatedBytes {
...
+        if let Some(metrics) = metrics {
+            metrics.add_bytes_transferred(output.remaining() as u64);
+        }
         Ok(AggregatedBytes(output))
     }
+
+    // /// Consumes self to create a [ChunkOutput]
+    // pub(crate) fn chunk_output(self, seq: u64, metadata: Option<ChunkMetadata>) -> ChunkOutput {
+
+    // }
```
ysaito1001: Meant to be removed?

aws-sdk-s3-transfer-manager/src/operation/download/handle.rs
```diff
@@ -63,17 +70,69 @@ impl DownloadHandle {
...
+        }
+    }
+}
+
+impl Drop for DownloadHandle {
+    fn drop(&mut self) {
+        if self.ctx.metrics().transfer_failed() {
+            self.ctx.handle.metrics.increment_transfers_failed();
+        } else {
+            self.ctx.handle.metrics.increment_transfers_completed();
```
ysaito1001: These should suffice. Technically a user can explicitly call `drop` while data transfer is still ongoing, but it's questionable why someone would want to do that and still expect metrics to come out accurately.

aws-sdk-s3-transfer-manager/src/io/aggregated_bytes.rs
landonxjames: Good catch, abandoned experiment, removed in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/440401ea5a181427c1f81e536820eb4d7dca3cc6

aws-sdk-s3-transfer-manager/src/operation/download/handle.rs
landonxjames: I think we have to consider manually dropping mid-transfer as UB when it comes to metrics. I don't really see any way we could maintain accurate counts in that case. Probably worth documenting as a restriction somewhere.

aws-sdk-s3-transfer-manager/src/io/aggregated_bytes.rs
```diff
@@ -45,13 +46,17 @@ impl AggregatedBytes {
...
+        metrics: Option<&ClientMetrics>,
     ) -> Result<Self, aws_smithy_types::byte_stream::error::Error> {
         let mut value = value;
         let mut output = SegmentedBuf::new();
         while let Some(buf) = value.next().await {
             let buf = buf?;
             output.push(buf);
         }
+        if let Some(metrics) = metrics {
+            metrics.add_bytes_transferred(output.remaining() as u64);
```
aajtodd: This isn't really any different from doing it outside of this function. I think I originally meant as we process each chunk from the bytestream. 

e.g.

```rs
        while let Some(buf) = value.next().await {
            let buf = buf?;
            output.push(buf);
            metrics.add_bytes_transferred(...);
        }
```

The rationale being able to potentially identify more granular instantaneous throughput and issues with processing a chunk. 

The only thing I don't know is if this will introduce too much overhead (I'd imagine not but only benchmarking will tell for sure). Perhaps doing it at the chunk level is good enough though that limits us to `part_size` view of throughput.

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
```diff
@@ -0,0 +1,698 @@
...
+    unit,
+};
+
+/// Client-level metrics aggregating all transfers
+#[derive(Debug, Clone, Default)]
+pub struct ClientMetrics {
+    /// Total number of transfers initiated
+    transfers_initiated: IncreasingCounter,
+    /// Total number of transfers completed successfully
+    transfers_completed: IncreasingCounter,
```
aajtodd: perhaps `transfers_successful` to make this clear, otherwise completed could be taken to mean transfers completed (successful and failed).

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
```diff
@@ -0,0 +1,698 @@
...
+        &self.throughput
+    }
+
+    /// Get request latency histogram
+    pub fn request_latency(&self) -> &Histogram {
+        &self.request_latency
+    }
+
+    /// Determine if the transfer has failed
+    pub fn transfer_failed(&self) -> bool {
```
aajtodd: `is_failed`?

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
```diff
@@ -0,0 +1,698 @@
...
+    pub(crate) fn mark_transfer_failed(&self) {
+        self.transfer_failed.store(true, Ordering::Relaxed);
+    }
+}
+
+/// Scheduler-level metrics for concurrency tracking
+#[derive(Debug, Clone, Default)]
+pub(crate) struct SchedulerMetrics {
+    /// Number of active permits in use
+    active_permits: Gauge,
```
aajtodd, line 169: Would an `(UpDown)Counter` be more appropriate here? Things like queue sizes probably make more sense as a regular counter (non monotonic).

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
```diff
@@ -0,0 +1,698 @@
...
+    /// Number of permits waiting to be acquired
+    permits_waiting: Gauge,
+    /// Total permits acquired
+    permits_acquired: IncreasingCounter,
```
aajtodd, line 173: How will this be used do you imagine?

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
```diff
@@ -0,0 +1,698 @@
...
+    }
+
+    /// Get the maximum throughput observed.
+    pub(crate) fn max(&self) -> Throughput {
+        let bps_scaled = self.max_throughput_bps.load(Ordering::Relaxed);
+        Throughput::new_bytes_per_sec(bps_scaled / 1000)
+    }
+
+    /// Get the average throughput.
+    pub(crate) fn avg(&self) -> Throughput {
```
aajtodd: May want to clarify this is a cumulative average

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
```diff
@@ -0,0 +1,698 @@
...
+                    hist.samples.pop_front();
+                }
+
+                hist.last_sample_time = Some(now);
+            }
+        }
+    }
+
+    /// Get the minimum throughput observed.
+    pub(crate) fn min(&self) -> Throughput {
```
aajtodd: Many of these accessors should be public I would imagine (min, max, avg, etc)

aws-sdk-s3-transfer-manager/src/operation/download/builders.rs
```diff
@@ -27,6 +27,7 @@ impl DownloadFluentBuilder {
         key = self.inner.key.as_deref().unwrap_or_default(),
     ))]
     pub fn initiate(self) -> Result<DownloadHandle, crate::error::Error> {
+        self.handle.metrics.increment_transfers_initiated();
```
aajtodd, line 30: One thing to consider is download and upload multiple objects make use of the singular `download` / `upload` operations.

e.g.

```rs
    let handle = tm
        .upload_objects()
        .source(source_dir)
        .bucket(bucket)
        .key_prefix(key_prefix)
        .recursive(true)
        .send()
        .await?;

    let output = handle.join().await?;

```

 Do we consider this a single "transfer" or not?

aws-sdk-s3-transfer-manager/src/operation/download/handle.rs
```diff
@@ -63,17 +70,69 @@ impl DownloadHandle {
...
+        self.ctx.metrics().clone()
+    }
+
     /// Abort the download and cancel any in-progress work.
     pub async fn abort(mut self) {
         self.body.close();
         self.discovery.abort();
-        let _ = self.discovery.await;
+
+        let discovery = std::mem::replace(&mut self.discovery, tokio::spawn(async {}));
```
aajtodd, line 85: why?

aws-sdk-s3-transfer-manager/src/operation/download/handle.rs
aajtodd: I'm not sure this is quite right. If they drop a handle (manually or not) it does not mean a transfer completed successfully. Only once the body is processed completely should we mark it as complete.

aws-sdk-s3-transfer-manager/src/io/aggregated_bytes.rs
landonxjames: Updated in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/6a336f3c3932c3699f1c5964ba0ab0d746a0c895,

Will take a shot at benching both versions now that the benchmark scaffolding is in place

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
landonxjames: Updated in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/6a336f3c3932c3699f1c5964ba0ab0d746a0c895

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
landonxjames: Updated in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/6a336f3c3932c3699f1c5964ba0ab0d746a0c895

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
landonxjames: Updated comment in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/6a336f3c3932c3699f1c5964ba0ab0d746a0c895, but could update the method name as well if you think that would be helpful

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
landonxjames: Made pub in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/6a336f3c3932c3699f1c5964ba0ab0d746a0c895

aws-sdk-s3-transfer-manager/src/operation/download/builders.rs
landonxjames: In my head a "transfer" is for a single S3 object. So the multi-object operations would comprise N transfers. I was considering adding an `OperationMetrics` aggregator for those that would store and aggregate across the individual transfers, but haven't gotten around to it yet.

aws-sdk-s3-transfer-manager/src/operation/download/handle.rs
landonxjames: This was a new one to me, but it turns out the implementing the `Drop` trait changes the move semantics a bit. The previous code generated this error once `impl Drop` was added:
```
error[E0509]: cannot move out of type `DownloadHandle`, which implements the `Drop` trait
  --> aws-sdk-s3-transfer-manager/src/operation/download/handle.rs:86:17
   |
86 |         let _ = self.discovery.await;
   |                 ^^^^^^^^^^^^^^
   |                 |
   |                 cannot move out of here
   |                 move occurs because `self.discovery` has type `tokio::task::JoinHandle<()>`, which does not implement the `Copy` trait

For more information about this error, try `rustc --explain E0509`.
error: could not compile `aws-sdk-s3-transfer-manager` (lib) due to 1 previous error
```

This [blog post](https://strongly-typed-thoughts.net/blog/rust-no-drop) does a pretty good job of explaining it. But agree it does look weird.

I should probably add a comment explaining the weirdness.

aws-sdk-s3-transfer-manager/src/operation/download/handle.rs
landonxjames: Updated `Body` to track completion and the metrics tracking in `Drop` to take the in to consideration when updating the succ/fail metric in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/5bed873bb659a4b5aae190d8883675dab10a5e68

Note: the CI is failing, but it seems unrelated to this, probably something to do with the recent serde refactor to split some things into a serde_core crate

aws-sdk-s3-transfer-manager/src/operation/download/builders.rs
aajtodd: From a user and API perspecitve the multi-object operations get a single "transfer handle" back though. If we ever add pause/resume/progress support it would be for the overall transfer not a single object. I think we'd want to align metrics this way as well but I admit I'm not yet entirely sure how we want to structure that. I think this needs some thought.

landonxjames: I think its kind of semantics at this point, but I would call that an "operation handle" since the comments refer to it as a "transfer operation". My intuitive sense of a transfer manager is that it manages the single file/object operations that the low level SDK does, which makes each of those a "transfer".

Regardless of naming I do agree that we need metrics for each ${insert term for single file download/upload} and for the operations that aggregate those. I'll work on adding that in future PRs.

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
landonxjames: This is kind of just a placeholder until I actually start instrumenting the scheduler. Will iterate on it then (that is the next PR I plan), so going to leave this as is for the moment to avoid cluttering this PR

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
landonxjames: See [comment](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121#discussion_r2383218311) above

aws-sdk-s3-transfer-manager/src/operation/download/handle.rs
landonxjames: Added explanatory comment in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/121/commits/2db102fe15bdc2d46d33390012ebef8bce82224b

PR #122, opened 2025-09-19
bootstrap benchmarks

aajtodd:
*Description of changes:*
Adds some basic benchmarks using the new/experimental mock server. In local testing I compared results to `iperf3` and experimented with `warp` as well. Throughput was slightly above or on par with results from `iperf3` over localhost. 

### DIFF
--- a/benches/throughput.rs
+++ b/benches/throughput.rs
@@ -1,0 +1,142 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use aws_sdk_s3_transfer_manager::error::Error;
+use aws_sdk_s3_transfer_manager::operation::download::DownloadHandle;
+use bytes::{BufMut, Bytes, BytesMut};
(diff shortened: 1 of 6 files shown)


benches/throughput.rs
```diff
@@ -0,0 +1,162 @@
...
+
+                        let start = std::time::Instant::now();
+                        for _ in 0..iters {
+                            let mut dl_handle = tm
+                                .download()
+                                .bucket("test-bucket")
+                                .key("test-key")
+                                .initiate()
+                                .expect("successful transfer initiate");
+                            drain(&mut dl_handle).await.unwrap();
```
ysaito1001: Do we need to wrap `drain(...).await.unwrap()` with [black_box](https://doc.rust-lang.org/beta/std/hint/fn.black_box.html) to avoid getting optimized away?

ysaito1001 (approved):
LGTM. Are we planning to detect potential performance regressions in future code changes at some point (like internal perf suite)?

landonxjames (approved):

aajtodd:
> Are we planning to detect potential performance regressions in future code changes at some point (like internal perf suite)?

Ideally yes but right now I still am building confidence these benchmarks are right/useful.

PR #123, opened 2025-09-25
Normailze `UploadContext` with the other OperationContext types

landonxjames:
*Description of changes:*

Updating the `UploadContext` type to use `TransferContext` instead of having the state flattened into the top level struct. This brings it in line with the other `${Operation}Context` types that already use this pattern


aajtodd (approved):
Nice.

PR #125, opened 2025-10-08
Add Scheduler and TokenBucket metrics

landonxjames:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
+++ b/aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
@@ -167,11 +167,9 @@ impl TransferMetrics {
 #[derive(Debug, Clone, Default)]
 pub(crate) struct SchedulerMetrics {
     // Total tasks in flight
-    inflight: Arc<AtomicU64>,
+    inflight: UpDownCounter,
     // High water mark for inflight tasks
-    max_inflight: Gauge,
-    // Count of permits acquired
-    permits_acquired: UpDownCounter,
+    max_inflight: IncreasingCounter,
     // Count of failed permit acquisitions
     permit_acquisition_failures: IncreasingCounter,
     // Tracking the wait time of permits
@@ -185,9 +183,8 @@ impl SchedulerMetrics {
     /// Create new scheduler metrics
     pub(crate) fn new() -> Self {
         Self {
-            inflight: Arc::new(AtomicU64::new(0)),
-            max_inflight: Gauge::new(),
-            permits_acquired: UpDownCounter::new(),
+            inflight: UpDownCounter::new(),
+            max_inflight: IncreasingCounter::new(),
             permit_acquisition_failures: IncreasingCounter::new(),
             permit_wait_time: Histogram::new(),
             scheduler_saturation: Gauge::new(),
@@ -197,38 +194,24 @@ impl SchedulerMetrics {
     /// Increment the number of in-flight requests and returns the number currently in-flight after
     /// incrementing.
     pub(crate) fn increment_inflight(&self) -> u64 {
-        (self.inflight.fetch_add(1, Ordering::Relaxed) + 1)
+        let val = self.inflight.increment(1);
+        self.update_max_inflight();
+        val
     }
 
     /// Decrement the number of in-flight requests and returns the number currently in-flight after
     /// decrementing.
     pub(crate) fn decrement_inflight(&self) -> u64 {
-        (self.inflight.fetch_sub(1, Ordering::Relaxed) - 1)
-    }
-
-    /// Get the current number of in-flight requests
-    #[cfg(test)]
-    pub(crate) fn inflight(&self) -> u64 {
-        self.inflight.load(Ordering::Relaxed)
+        self.inflight.decrement(1)
     }
 
     /// Update max inflight metric
-    pub(crate) fn update_max_inflight(&self, current_inflight: u64) {
+    fn update_max_inflight(&self) {
         let current_max = self.max_inflight.value();
-        let current_inflight = current_inflight as f64;
+        let current_inflight = self.inflight();
         if current_inflight > current_max {
-            self.max_inflight.set(current_inflight);
+            self.max_inflight.increment(current_inflight - current_max);
         }
-    }
-
-    /// Record permit acquisition
-    pub(crate) fn record_permit_acquisition(&self) {
-        self.permits_acquired.increment(1);
-    }
-
-    /// Record permit release
-    pub(crate) fn record_permit_release(&self) {
-        self.permits_acquired.decrement(1);
     }
 
     /// Record permit acquisition failure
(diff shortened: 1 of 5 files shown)


aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
```diff
@@ -165,46 +166,152 @@ impl TransferMetrics {
...
-    permits_waiting: Gauge,
-    /// Total permits acquired
-    permits_acquired: IncreasingCounter,
-    /// Permit acquisition wait time
-    permit_wait_time: Histogram,
-    /// Current number of in-flight requests (for backward compatibility)
+    // Total tasks in flight
     inflight: Arc<AtomicU64>,
+    // High water mark for inflight tasks
+    max_inflight: Gauge,
```
aajtodd: Wouldn't a high water mark be a simple monotonic counter? A gauge is for instantaneous measurements usually

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
```diff
@@ -165,46 +166,152 @@ impl TransferMetrics {
...
+    // Count of permits acquired
+    permits_acquired: UpDownCounter,
```
aajtodd: Shouldn't this be the same as in-flight?

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
```diff
@@ -165,46 +166,152 @@ impl TransferMetrics {
...
+    pub(crate) fn scheduler_saturation(&self) -> &Gauge {
+        &self.scheduler_saturation
+    }
+}
+
+#[derive(Debug, Default, Clone)]
+pub(crate) struct TokenBucketMetrics {
+    max_tokens: u64,
+    available_tokens: UpDownCounter,
+    token_wait_time: Histogram,
```
aajtodd, line 267: I'm not sure if this will be meaningful or not given permit requests come with different token amounts requested.

aws-sdk-s3-transfer-manager/src/middleware/limit/concurrency/future.rs
```diff
@@ -82,7 +82,13 @@ where
                         Ok(_permit) => {
                             let req = this.request.take().expect("request set");
                             let inflight = this.scheduler.metrics.increment_inflight();
+                            this.scheduler.metrics.update_max_inflight(inflight);
```
aajtodd: Shouldn't `increment_inflight()` just handle this?

aws-sdk-s3-transfer-manager/src/runtime/token_bucket.rs
```diff
@@ -149,32 +150,43 @@ pub(crate) struct TokenBucket {
     // NOTE: tokio semaphore is fair, permits are given out in the order requested
     semaphore: Arc<Semaphore>,
     mode: ConcurrencyMode,
+    tb_metrics: Arc<TokenBucketMetrics>,
```
aajtodd, line 153: I'm kind of wondering whether token bucket metrics provide value at this point in time. 🤔

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
landonxjames: Updated in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/125/commits/1d9e705f796e540b6281dfafed3279c0d606a291

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
landonxjames: I was thinking of them slightly differently. `inflight` was the futures in flight, and `permits_acquired` was the actual `OwnedPermit` structs handed out. But you're right those do functionally count the same thing and the time when they aren't aligned is tiny. Updated to only track inflight in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/125/commits/1d9e705f796e540b6281dfafed3279c0d606a291

aws-sdk-s3-transfer-manager/src/middleware/limit/concurrency/future.rs
landonxjames: Moved inside `increment_inflight` in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/125/commits/1d9e705f796e540b6281dfafed3279c0d606a291

aws-sdk-s3-transfer-manager/src/runtime/token_bucket.rs
landonxjames: I am also not sure, but will probably leave them in place until I can actually run some benchmarks and test all of this on some real use cases. Can remove if we find it not necessary after that testing.

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
landonxjames: I wonder if tracking percentage of available tokens over time or something would be more useful?

aws-sdk-s3-transfer-manager/src/runtime/scheduler.rs
```diff
@@ -21,30 +23,49 @@ use crate::types::ConcurrencyMode;
 #[derive(Debug, Clone)]
 pub(crate) struct Scheduler {
     token_bucket: TokenBucket,
-    pub(crate) metrics: SchedulerMetrics,
+    pub(crate) metrics: Arc<SchedulerMetrics>,
```
ysaito1001, line 26: nit: will this end up double-`Arc` ing? Individual fields of `SchedulerMetrics` seem to have their types backed by `Arc` already, e.g. `Histogram`, `Gauge`?

aws-sdk-s3-transfer-manager/src/metrics/aggregators.rs
```diff
@@ -165,46 +166,135 @@ impl TransferMetrics {
...
-    permits_waiting: Gauge,
-    /// Total permits acquired
-    permits_acquired: IncreasingCounter,
-    /// Permit acquisition wait time
+    // Total tasks in flight
+    inflight: UpDownCounter,
+    // High water mark for inflight tasks
+    max_inflight: IncreasingCounter,
+    // Count of failed permit acquisitions
+    permit_acquisition_failures: IncreasingCounter,
```
aajtodd, line 174: not sure we need this at this stage

aws-sdk-s3-transfer-manager/src/runtime/token_bucket.rs
aajtodd: Ok maybe add a comment or something to that effect

aws-sdk-s3-transfer-manager/src/runtime/token_bucket.rs
```diff
@@ -187,35 +199,69 @@ impl TokenBucket {
...
     }
 }
 
 impl AcquireTokenFuture {
-    fn new(sem: PollSemaphore, tokens: u32) -> Self {
-        Self { sem, tokens }
+    fn new(
+        sem: PollSemaphore,
+        tokens: u32,
+        tb_metrics: Arc<TokenBucketMetrics>,
```
aajtodd, line 254: this _feels_ off/wrong to need all these parameters for this future but I get why it's done. Wondering if there isn't a cleaner approach though.

aajtodd (approved):

Issue #90, opened 2025-01-21
Allow streaming upload of unknown content length


aajtodd:
This has been asked for before. 

See:
* https://github.com/awslabs/aws-sdk-rust/issues/541#issuecomment-2602915350
* https://github.com/awslabs/aws-sdk-rust/issues/494#issuecomment-2602878569

waahm7:
Don't we already support it? Customers will have to provide a size_hint upper bound to figure out the part_size but other than that we should support it because of https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/main/aws-s3-transfer-manager/src/operation/upload/service.rs#L213. We should probably add tests for this use-case, though.

aajtodd:
No precisely because we force them to provide an upper bound size hint they can't use a stream of unknown content length (e.g. it may be useful to provide a stream that does compression as it's read and therefore they wouldn't know the size without compressing first).

waahm7:
Yeah, but they can just guess the upper bound size. It's kind of the same in aws-c-s3 that the upper bound size of unknown content length stream is part_size * 10000. Unless we are planning to do something like bumping the part_size if we are running out of parts dynamically, then customers will have to provide the upper bound either directly or indirectly due to 10000 parts limit.

If someone sets the size_hint to 30 GB but the stream ends at 1 GB, our code will work fine.

PR #132, opened 2026-03-05
move uploads to new scheduler

aajtodd:
## Summary

Replace the Tower service-based upload orchestration (`ConcurrencyLimitLayer`, `service_fn`, `Buffer`) with a scheduler-driven state machine. The upload transfer implements the `Transfer` trait from #131. The scheduler polls the transfer for work; the transfer produces work items lazily, one at a time.

### Multipart upload

```
orchestrate() → UploadTransfer enqueued to scheduler

poll_work → CreateMPU           [Network]
poll_work → Pending              (waiting for upload_id)
  ... CreateMPU completes, wake ...
poll_work → UploadPart(1)       [DataIO]   ← disk read
poll_work → UploadPart(2)       [DataIO]   ← disk read
  ... DataIO completes → follow-on: UploadPart(1) [Network] ...
  ... all parts complete ...
poll_work → CompleteMPU         [Network]
poll_work → Done
```

### Small file (below MPU threshold)

Skips multipart entirely — produces a single `PutObject` work item.

## Design

Each `UploadPart` has two phases: a **DataIO** phase that reads the part from the `PartReader`, then a **Network** phase that sends it to S3. The Network phase is a follow-on work item that bypasses the ready set, so the buffer isn't held waiting behind other transfers.

Error handling follows the download transfer's contract:
- `fail()` stores the original error via `ctx.set_failed(error)` and signals terminal
- `join()` awaits `completion_rx` from `TransferContext`, then reads the error from `ctx.take_error()` on failure or the result from `transfer.take_result()` on success
- No separate result channel — errors and results flow through `TransferContext` and the transfer struct respectively

## What changed

| File | Change |
|------|--------|
| `operation/upload/transfer.rs` | **New.** Upload state machine implementing `Transfer` trait. |
| `operation/upload/upload.rs` | `orchestrate()` creates `UploadTransfer` and enqueues to scheduler instead of spawning Tower service pipeline. |
| `operation/upload/handle.rs` | Holds `UploadTransfer` + `completion_rx` directly. `join()` follows download's pattern. Abort waits for scheduler idle, calls `AbortMultipartUpload`. |
| `operation/upload/service.rs` | **Deleted.** Tower middleware chain no longer needed. |
| `scheduler/mod.rs` | Re-exports `StateMachineTerminalReceiver`. |

## What did not change

- Download still uses the old `runtime::scheduler` and service pattern
- `upload_objects` / `download_objects` unchanged
- `middleware/`, `runtime/` kept for download and multi-object operations
- No scheduler changes beyond the re-export
- No public API changes


aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -0,0 +1,805 @@
...
+
+/// Internal state for upload transfer.
+#[derive(Debug)]
+struct UploadTransferInner {
+    /// Common transfer lifecycle management
+    ctx: TransferContext,
+    /// State machine for work progression
+    state: Mutex<UploadState>,
+    /// The original request (body taken for processing)
+    request: Arc<UploadInput>,
```
ysaito1001, line 69: Will this end up being double `Arc`, seen from `UploadTransfer`?

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -0,0 +1,805 @@
...
+        &self.inner.request
+    }
+
+    /// Take the upload result. Available after successful completion.
+    pub(crate) fn take_result(&self) -> Option<UploadOutput> {
+        self.inner.result.lock().unwrap().take()
+    }
+
+    /// Get the upload_id if MPU was started.
+    pub(crate) fn upload_id(&self) -> Option<String> {
```
ysaito1001, line 130: nit: do we always want to clone a returned `String`, as opposed to `Option<&str>`?

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -0,0 +1,805 @@
...
+    /// Returns:
+    /// - `PollWork::Ready(work)` - work available to execute
+    /// - `PollWork::Pending` - blocked waiting for in-flight work
+    /// - `PollWork::Done` - transfer complete
+    pub(crate) fn poll_work(&self) -> PollWork {
+        if !self.inner.ctx.is_active() {
+            return PollWork::Done;
+        }
+
+        let mut state = self.inner.state.lock().unwrap();
```
landonxjames:
```suggestion
        let mut state = self.inner.state.lock().expect("Failed to unlock transfer state");
```

landonxjames: This suggestion applies to a few places, just putting it here so I don't spam it everywhere.

aws-sdk-s3-transfer-manager/src/operation/upload.rs
```diff
@@ -67,222 +55,70 @@ impl Upload {
...
-        read_body_tasks: Default::default(),
-        response: Some(mpu),
-        upload_id,
-        part_reader,
-    };
-
-    distribute_work(&mut mpu_data, ctx)?;
-    Ok(UploadType::MultipartUpload(mpu_data))
-}
+        // TODO: Relax this constraint - unknown content length implies MPU
```
landonxjames:
```suggestion
        // TODO(vnext): Relax this constraint - unknown content length implies MPU
```

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
landonxjames: More generally, this Mutex seems like it is going to be locked for every `poll_work` and `execute_*` and like that could cause a lot of contention. I wonder if using something that is supposed to be more performant, like [parking_lot::Mutex](https://docs.rs/parking_lot/0.12.5/parking_lot/type.Mutex.html) would make sense here?

Probably a premature optimization, but worth evaluating when you get to benchmarking this

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -0,0 +1,805 @@
...
+                WorkOutcome::Success(Some(WorkItem {
+                    kind: WorkKind::Network,
+                    data: Some(Box::new(UploadWork::UploadPart {
+                        part_number,
+                        part_data: part_data.take(),
+                    })),
+                }))
+            }
+            Ok(None) => {
+                tracing::warn!("part_reader returned None for part {}", part_number);
```
landonxjames: Think this should just be a `trace` log? If `content_length` is just an upper bound instead of exact (which is how I think `size_hint` works?) then this will happen sometimes and shouldn't be exposed as something failing.

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -0,0 +1,805 @@
...
+            Ok(resp) => resp,
+            Err(e) => return self.fail(e.into()),
+        };
+
+        let upload_id = resp.upload_id().expect("upload_id present").to_string();
+        let response_builder = UploadOutputBuilder::from(resp);
+
+        let (stream, content_length) = {
+            let mut state = self.inner.state.lock().unwrap();
+            match std::mem::replace(&mut *state, UploadState::Done) {
```
landonxjames: Is replacing this with `Done` correct? If `poll_work` is called in the window what `state` is set to `Done` won't it return `PollWork::Done`, causing the scheduler to remove
the transfer entirely? Seems like it would be possible since you have to re-acquire the lock on the state later in this function?

aws-sdk-s3-transfer-manager/src/operation/upload/handle.rs
```diff
@@ -3,247 +3,184 @@
...
-    while (tasks.join_next().await).is_some() {}
-
-    let abort_policy = ctx
-        .state
-        .request
-        .failed_multipart_upload_policy
-        .clone()
-        .unwrap_or_default();
-    match abort_policy {
-        FailedMultipartUploadPolicy::Retain => Ok(AbortedUpload::default()),
```
landonxjames, line 135: We are no longer checking the `abort_policy` is that intended?

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -0,0 +1,805 @@
...
+                .upload_id(&upload_id)
+                .part_number(part_num_i32)
+                .content_length(content_length)
+                .body(ByteStream::from(data.data)),
+            data.checksum.as_ref(),
+        );
+
+        let resp = match req
+            .customize()
+            .disable_payload_signing()
```
landonxjames, line 441: We disable this for MPU but not `PutObject`? Might make sense but would probably want a comment explaining why

aws-sdk-s3-transfer-manager/src/operation/upload.rs
```diff
@@ -67,222 +55,70 @@ impl Upload {
...
-    use std::sync::Arc;
-    use std::sync::Barrier;
+
+    use crate::io::InputStream;
+    use crate::metrics::unit::ByteUnit;
+    use crate::operation::upload::UploadInput;
+    use crate::types::{ConcurrencyMode, PartSize};
 
     #[tokio::test]
-    async fn test_basic_mpu() {
```
landonxjames, line 234: Doesn't seem to be a new version of the happy path `test_basic_mpu` and `test_basic_upload_object` tests?

landonxjames (approved):
Few questions about some of the logic. Nothing major or blocking.

ysaito1001:
Again, to aid code review:

<details>
<summary> a runtime sequence diagram for MPU </summary>

```
                          MPU Upload Runtime Lifecycle
                          ===========================
 
 builders.rs              upload.rs                transfer.rs              scheduler/          handle.rs
 (initiate)            (Upload::orchestrate)       (UploadTransfer)         (Scheduler)         (UploadHandle)
     │                        │                         │                       │                    │
     │──initiate_with()──────>│                         │                       │                    │
     │                        │ resolve checksum_strategy│                       │                    │
     │                        │ take_body() from input   │                       │                    │
     │                        │ check size_hint upper    │                       │                    │
     │                        │ BucketType::from_bucket_name                     │                    │
     │                        │                         │                       │                    │
     │                        │ TransferContext::new(handle)                     │                    │
     │                        │──────────────────────────> new(ctx, bucket_type, │                    │
     │                        │                           input, stream)        │                    │
     │                        │                         │                       │                    │
     │                        │                         │  state = PendingInit  │                    │
     │                        │                         │  {stream,             │                    │
     │                        │                         │   content_length,     │                    │
     │                        │                         │   init_in_flight:false}│                    │
     │                        │                         │                       │                    │
     │                        │─────────────────────────────enqueue_transfer(───>│                    │
     │                        │                         │  Box<transfer.clone>)  │                    │
     │                        │                         │                       │                    │
     │                        │──────────────────────────────────────────────────────> new(completion_rx,
     │<───UploadHandle────────│                         │                       │        transfer)   │
     │                        │                         │                       │                    │
 
 
                    ┌─── Scheduler poll loop begins ───┐
                    │                                   │
                    ▼                                   │
 
 scheduler/                          transfer.rs
 (Scheduler)                         (UploadTransfer)
     │                                    │
     │──poll_work()──────────────────────>│
     │                                    │ state == PendingInit
     │                                    │ content_length >= mpu_threshold
     │                                    │ set init_in_flight = true
     │<──PollWork::Ready(CreateMPU, ──────│
     │                   Network)         │
     │                                    │
     │──execute(CreateMPU)───────────────>│
     │                                    │ s3_client.create_multipart_upload().send()
     │                                    │ take stream + content_length from PendingInit
     │                                    │ compute part_size, total_parts
     │                                    │ build PartReader
     │                                    │ state = Transferring {
     │                                    │   upload_id, part_reader,
     │                                    │   next_part: 1, total_parts,
     │                                    │   parts_in_flight: 0,
     │                                    │   completed_parts: [],
     │                                    │   response_builder }
     │                                    │ create_mpu_complete.notify_waiters()
     │                                    │ ctx.try_wake()
     │<──WorkOutcome::Success(None)───────│
     │                                    │
     │──poll_work()──────────────────────>│  (repeated for each part 1..total_parts)
     │                                    │ state == Transferring
     │                                    │ next_part++, parts_in_flight++
     │<──PollWork::Ready(UploadPart{N},───│
     │                   DataIO)          │
     │                                    │
     │──execute(UploadPart{N}, DataIO)───>│
     │                                    │ part_reader.next_part()  (disk read)
     │                                    │ store part_data in work item
     │<──WorkOutcome::Success(Some(───────│  ◄── follow-on work item
     │     WorkItem{UploadPart{N},        │
     │              Network}))            │
     │                                    │
     │──execute(UploadPart{N}, Network)──>│
     │                                    │ s3_client.upload_part()
     │                                    │   .upload_id(&upload_id)
     │                                    │   .part_number(N)
     │                                    │   .body(data)
     │                                    │   .disable_payload_signing()
     │                                    │   .send()
     │                                    │ build CompletedPart from response
     │                                    │ push to completed_parts
     │                                    │ parts_in_flight--
     │                                    │ if next_part > total_parts && parts_in_flight == 0:
     │                                    │   state = Completing {
     │                                    │     upload_id, part_reader,
     │                                    │     completed_parts, response_builder,
     │                                    │     complete_in_flight: false }
     │                                    │   ctx.try_wake()
     │<──WorkOutcome::Success(None)───────│
     │                                    │
     │──poll_work()──────────────────────>│
     │                                    │ state == Completing
     │                                    │ set complete_in_flight = true
     │<──PollWork::Ready(CompleteMPU, ────│
     │                   Network)         │
     │                                    │
     │──execute(CompleteMPU)─────────────>│
     │                                    │ sort completed_parts by part_number
     │                                    │ s3_client.complete_multipart_upload()
     │                                    │   .upload_id(&upload_id)
     │                                    │   .multipart_upload(parts)
     │                                    │   .send()
     │                                    │ build UploadOutput via response_builder
     │                                    │ store in self.result
     │                                    │ ctx.set_completed()
     │                                    │ ctx.signal_terminal()
     │<──WorkOutcome::Success(None)───────│
     │                                    │
     │──poll_work()──────────────────────>│
     │                                    │ state == Done
     │<──PollWork::Done───────────────────│
```
</details>

aws-sdk-s3-transfer-manager/src/operation/upload.rs
ysaito1001: Or https://github.com/awslabs/aws-s3-transfer-manager-rs/issues/90

ysaito1001 (approved):
Left nit comments. Looks good!

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -0,0 +1,805 @@
...
+                parts_in_flight,
+                ..
+            } => {
+                if *next_part > *total_parts {
+                    if *parts_in_flight > 0 {
+                        self.inner.ctx.set_pending();
+                        return PollWork::Pending;
+                    }
+                    self.inner.ctx.set_pending();
+                    return PollWork::Pending;
```
ysaito1001, line 229: Just to double check, once `next_part` exceeds `total_parts`, is returning `PollWork::Pending` here is expected?

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
aajtodd: Good catch, no this is wrong.

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
aajtodd: No, in practice this state is never observed I think because we transition to completing before poll_work sees it. I'll update it though to transition to completing as well.

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
aajtodd: Just a miss when converting.

aws-sdk-s3-transfer-manager/src/operation/upload/handle.rs
aajtodd: Nope will fix

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
aajtodd: It's not on a hot path, only for abort, I'm going to leave for now.

PR #133, opened 2026-03-18
migrate download to state machine and new scheduler

aajtodd:
## Summary

Replace the Tower service-based download orchestration with a scheduler-driven state machine and introduce a slot buffer for body delivery. Also pulls forward shared infrastructure (transfer abstraction layer, execution runtime trait, metrics, concurrency types) to reduce divergence with upcoming PRs.

The download transfer implements the `Transfer` trait from [#131](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/131). The scheduler polls the transfer for work; the transfer produces work items lazily: discovery first, then ranged GetObject requests with backpressure from the slot buffer.

## Download Lifecycle

```
orchestrate() → DownloadTransfer enqueued to scheduler

poll_work → Discovery            [Network]
poll_work → Pending               (waiting for discovery response)
  ... Discovery completes, wake, initial chunk fills slot 0 ...
poll_work → GetObjectRange(1)    [Network]   ← slot claimed
poll_work → GetObjectRange(2)    [Network]   ← slot claimed
  ... slot buffer full → Pending (backpressure) ...
  ... consumer reads slot → wake ...
poll_work → GetObjectRange(3)    [Network]
  ... all ranges complete ...
poll_work → Done
```

## Slot Buffer Body

Replaces the `mpsc` channel + `BinaryHeap` sequencer with a fixed-size ring buffer of slots. Each slot transitions `Empty → Filled` with atomic state, indexed by `seq % capacity`.

The old body received chunks out of order over an mpsc channel and re-sequenced them in a `BinaryHeap`. The slot buffer eliminates the channel and the sequencer entirely: producers write directly to the slot for their sequence number, and the consumer reads slots in order. Sequencing is implicit in the ring index.

Backpressure: `try_claim` fails when `claimed - consumed >= capacity`, causing the transfer to return `Pending`. When the consumer reads a slot, it wakes the transfer to generate more work.

`UnsafeCell` safety relies on three invariants documented on `SlotBuffer`: exclusive producer access (unique seq via CAS), state-gated consumer access (acquire/release on atomic state), and the seq window preventing producer/consumer overlap on the same slot.

The `Body` type wraps `SlotBodyConsumer` and detects terminal transfer state (failure/cancellation) to surface errors through `next()`.

### Transfer Abstraction (`transfer.rs`)

Top-level module consolidating types that were previously split across `scheduler/context.rs`, `scheduler/work/item.rs`, and `scheduler/transfer.rs`:

- `Transfer` trait, `TransferContext`, `StateMachineStatus` for transfer lifecycle
- `IoRequest`, `IoKind`, `WorkOutcome` (renamed from `WorkItem`/`WorkKind`)
- `WorkOutcome::Success` is now a struct variant `{ schedule_next, data, metrics }` instead of a tuple

### Execution Runtime (`runtime/tokio_mt.rs`)

The existing worker pool and worker loop extracted behind an `ExecutionRuntime` trait. Same behavior as before: `tokio::spawn` workers pulling from a shared queue. The trait (`dispatch`, `shutdown`, `remove_pending_for_transfer`) decouples the scheduler from the execution model so future runtimes can be swapped in without changing the scheduler.

### Concurrency & Metrics

- `ErrorKind`, `CompletionSample`, `classify_error`: error classification for the concurrency controller. Used by upload's `fail()` and wired through scheduler `on_completion`.
- `IoSample`, `IOWindow`, `IOCounters`: rolling throughput measurement. `IOWindow` tracks bytes over a short fixed-duration window (e.g. 500ms) to estimate current throughput. Used by the scheduler to report to the concurrency controller.
- `SchedulerBuilder`: breaks the scheduler/runtime circular dependency via a factory pattern. The scheduler needs the runtime to dispatch work, and the runtime needs the scheduler to report completions. The builder creates the scheduler first, passes it to the factory which creates the runtime, then stores the runtime on the scheduler.

### Tracing Targets (`telemetry.rs`)

Structured tracing targets for filtering logs by concern rather than module path:

```
RUST_LOG=aws_s3_transfer_manager::execution=trace    # per-work-item dispatch/complete
RUST_LOG=aws_s3_transfer_manager::scheduling=debug   # capacity decisions, worker growth
RUST_LOG=aws_s3_transfer_manager::transfer=debug     # transfer lifecycle events
RUST_LOG=aws_s3_transfer_manager::concurrency=debug  # concurrency controller decisions
```

Used by the worker loop, scheduler, and transfer state machines.

## How to Review

1. **Start with `transfer.rs`** (the top-level module). This is the type foundation everything else builds on. The `Transfer` trait, `TransferContext`, `IoRequest`/`IoKind`/`WorkOutcome` types, and `StateMachineStatus` are the key abstractions.

2. **Review the slot buffer in `download/body.rs`**. The `UnsafeCell` safety argument is the most important thing to verify. The three invariants on `SlotBuffer` should be checked against the actual access patterns in `fill()`, `try_take_next()`, and `try_claim()`.

3. **Review `download/transfer.rs`** for the state machine. The `poll_work` / `execute` split, the `DownloadState` transitions, and the `fail()` / `bail_if_terminal!` error handling pattern.

4. **Review `download/handle.rs`** for the public API: `join()`, `abort()`, `object_meta()`, `body_mut()`, and the `Drop` impl.

5. **Skim the scheduler changes** (`scheduler/scheduler.rs`). The main additions are `SchedulerBuilder`, `dispatched` counter for capacity tracking, and `on_completion` with metrics reporting. The generate/dispatch/completion loop is the same structure as before.

6. **The runtime extraction** (`runtime/tokio_mt.rs`, `runtime/tokio_mt/worker_pool.rs`) is a move, not a rewrite. The worker loop and pool are the same code that was in `scheduler/scheduler.rs` and `scheduler/work/pool.rs`.

7. **Upload changes are mechanical**: type renames (`WorkItem`→`IoRequest`, etc.) and `classify_error` added to `fail()`. All PR #132 review feedback is preserved.

## What Changed

| Area | Files | Change |
|------|-------|--------|
| **Download** | `download/transfer.rs` | **New.** Download state machine implementing `Transfer`. |
| | `download/body.rs` | **Rewritten.** Slot buffer replaces mpsc + sequencer. |
| | `download/handle.rs` | **Rewritten.** `join()` awaits completion, `abort()` cancels via scheduler. |
| | `download/discovery.rs` | Adapted for `TransferContext`. |
| | `download/output.rs` | **New.** `DownloadOutput` wrapping `ObjectMetadata`. |
| | `download/service.rs` | **Deleted.** |
| | `download/retry.rs` | **Deleted.** |
| | `download.rs` | `orchestrate()` creates transfer + slot body, enqueues to scheduler. |
| **Transfer types** | `transfer.rs` | **New.** Consolidated from `scheduler/context.rs` + `scheduler/work/item.rs`. |
| | `telemetry.rs` | **New.** Tracing target constants. |
| | `metrics.rs` | Added `IoSample`, `IOWindow`, `IOCounters`. |
| **Scheduler** | `scheduler/scheduler.rs` | `SchedulerBuilder`, `dispatched` counter, `on_completion` with metrics, worker loop moved to runtime. |
| | `scheduler/concurrency.rs` | Added `ErrorKind`, `CompletionSample`, `classify_error`, `on_dispatch`/`on_completion` on trait. |
| | `scheduler/test_util.rs` | **New.** Combined from `scheduler/transfer/mock.rs` + `test_util.rs`. |
| | `scheduler/context.rs` | **Deleted.** Absorbed into `transfer.rs`. |
| | `scheduler/work/` | **Deleted.** Types moved to `transfer.rs`, pool moved to runtime. |
| **Runtime** | `runtime/tokio_mt.rs` | **New.** `TokioMultiThreadRuntime` implementing `ExecutionRuntime`. |
| | `runtime/tokio_mt/worker_pool.rs` | **Moved** from `scheduler/work/pool.rs` + `queue.rs`. |
| | `runtime.rs` | Added `ExecutionRuntime` trait, `ScheduledWork`. |
| **Upload** | `upload/transfer.rs` | Type renames, `classify_error` in `fail()`. `UploadState` inlined from `context.rs`. |
| | `upload/context.rs` | **Deleted.** Inlined into `transfer.rs`. |
| | `upload/handle.rs`, `upload.rs` | Import path updates. |
| **Multi-object** | `operation.rs` | `TransferContext<State>` → `LegacyTransferContext<State>`, `CancelNotification` → `CancelBroadcast`. |
| | `download_objects/`, `upload_objects/` | Same renames, `body_mut().next()` instead of `mem::replace`. |
| **Deps** | `Cargo.toml` | `aws-sdk-s3` 1.97.0 → 1.122.0, `aws-smithy-mocks` 0.1.1 → 0.2. |

## What Did Not Change

* Upload logic: all PR #132 review feedback preserved (original error in `fail()`, `take_result()` pattern, `Option` fields with `.take()`, `abort_policy`, `disable_payload_signing`)
* `upload_objects` / `download_objects` beyond import renames
* `middleware/`, `runtime/scheduler.rs`, `runtime/token_bucket.rs` (kept for legacy paths)
* `io/`, `config/`, `types.rs`, `error.rs`
* No public API changes beyond `Body` (which was already `#[non_exhaustive]`)

---


aws-sdk-s3-transfer-manager/src/operation/download/body.rs
```diff
@@ -3,209 +3,324 @@
...
+    pub(crate) async fn next(&self, is_terminal: impl Fn() -> bool) -> Option<ChunkOutput> {
+        loop {
+            if let Some(chunk) = self.buffer.try_take_next() {
+                return Some(chunk);
+            }
+            if is_terminal() {
+                // Drain any slot filled between our check and here.
+                return self.buffer.try_take_next();
+            }
+            self.buffer.notify.notified().await;
```
landonxjames: Is the ordering here correct? The [Notify docs](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html) say:
>If notify_one() is called before notified().await, then the next call to notified().await will complete immediately, consuming the permit. Any subsequent calls to notified().await will wait for a new permit.

The `notified()` future is created after the `is_terminal()` 
check. If a producer calls `notify_one()` between the `is_terminal()` check
and the `.notified().await`, the notification is lost. The consumer then
waits indefinitely because `notify_one()` only wakes a waiting 
future. 

I think the fix would be simple, call `self.buffer.notify.notified()` before the 
checks and then only `await` it if both checks fail

aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
```diff
@@ -0,0 +1,1038 @@
...
+            } = &mut *work
+            {
+                *ranges_in_flight = ranges_in_flight.saturating_sub(1);
+                *ranges_in_flight == 0
+            } else {
+                false
+            }
+        };
+        if should_wake {
+            self.inner.ctx.try_wake();
```
landonxjames, line 511: Seems like it only calls `try_wake()` if nothing is in flight? That seems like it is missing something.

The transfer is marked `Pending` when the `SlotBuffer` is exhausted.
When a range completes and the consumer isn't drained yet, the 
transfer won't be woken to check if the slot window has opened. The 
consumer's `Body::next()` calls `try_wake()` after consuming a slot, 
which covers part of this, but misses that if the consumer hasn't 
called `Body::next()` yet but a range has been completed (freeing up an in-flight count) the transfer stays `Pending` even though it could generate 
more work.

aws-sdk-s3-transfer-manager/src/operation/download/service.rs
```diff
@@ -1,305 +0,0 @@
...
-    }
-
-    tracing::trace!("work fully distributed");
-}
-
-/// Validates that the response content range matches the requested range.
-///
-/// Handles both "1024-2047" and "bytes=1024-2047" formats for the requested content range.
-/// The response_content_range is of the form Some("bytes 1024-2047/4096").
-fn validate_content_range(
```
landonxjames, line 246: I might be missing it or it might have just been renamed, but this check seems to have been left out of the newer code?

aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
```diff
@@ -0,0 +1,1038 @@
...
+        if let Some(ref etag) = etag {
+            req = req.if_match(etag.as_ref());
+        }
+
+        let resp = match req.send().await {
+            Ok(r) => r,
+            Err(e) => return self.fail_range(seq, e),
+        };
+
+        bail_if_terminal!(self);
```
landonxjames, line 438: This is only called before and after the body read. If there is a cancellation in the middle of a long read that won't be reflected until the read completes. Feels like this should be a `select!` or other channel based implementation.

aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
```diff
@@ -0,0 +1,1038 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Download transfer implementation for scheduler integration.
+
+use std::cmp;
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::{Arc, Mutex};
```
landonxjames, line 11: Using `std::sync::Mutex` in `async` contexts always makes me nervous, easy to add an `await` while the `MutexGuard` is held and lock everything up. Should this be `tokio::sync::Mutex`? I think you've mentioned wanting to be async runtime agnostic, so maybe that wouldn't work

landonxjames (commented):
Overall looks good, couple of questions, no blockers (but I accidentally posted before I clicked approve)

landonxjames (approved):

aws-sdk-s3-transfer-manager/src/operation/download/body.rs
```diff
@@ -3,209 +3,324 @@
...
-            chunks: BinaryHeap::with_capacity(8),
-            next_seq: 0,
-        }
-    }
+/// A claimed slot in the body buffer. Produced by [`BodyWriter::try_claim`],
+/// consumed by [`fill`](Self::fill). Cannot be constructed outside this module,
+/// preventing writes to unclaimed slots. Consuming `self` on fill prevents
+/// double-writes.
+#[derive(Debug)]
+pub(crate) struct BodySlot {
```
vcjana, line 144: I might be missing something, but if a worker is cancelled after claiming a slot but before `fill()`, the slot stays `EMPTY` until the terminal flag fires. Would a `Drop` impl on `BodySlot` that notifies the consumer help here?

aws-sdk-s3-transfer-manager/src/operation/download/body.rs
```diff
@@ -3,209 +3,324 @@
...
+                if transfer.ctx().is_cancelled() {
+                    Some(Err(crate::error::from_kind(
+                        crate::error::ErrorKind::OperationCancelled,
+                    )("download cancelled")))
+                } else if transfer.ctx().is_failed() {
+                    let kind = transfer
+                        .ctx()
+                        .error_kind()
+                        .unwrap_or(crate::error::ErrorKind::ChildOperationFailed);
+                    Some(Err(crate::error::from_kind(kind)("transfer failed")))
```
vcjana: should this return the stored error instead so users don't have to call `join()` separately?

aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
vcjana: `is_active()` check inside the `while let` loop might be simpler than a full `select!` and would still catch cancellation between chunks.

vcjana (approved):

aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
aajtodd: The transfer goes `Pending` because `try_claim()` returned `None` (buffer full). The only thing that frees a slot is the consumer calling `try_take_next()`.  Decrementing in-flight doesn't free a slot. So waking the transfer on every completion would just have it call `try_claim()` again and get `None` again.

aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
aajtodd: It's intentional. I understand the concern but that's the kind of thing we have to deal with to get the right performance. We have some very small critical sections (see also tokio's own [discussion](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use) on this)

ysaito1001 (commented):
excellent, LGTM!

PR #135, opened 2026-03-27
implement support for adaptive concurrency

aajtodd:
# Adaptive Concurrency Controller

## What this PR adds

An adaptive concurrency controller that discovers and maintains the optimal number of concurrent S3 requests for the current network conditions. When `ConcurrencyMode::Auto` is configured, the controller replaces the fixed concurrency target with a three-state algorithm that ramps up during warmup, fine-tunes at steady state, and sheds excess when conditions change.

This is PR 4 in the redesign series:

4. **This PR**: Adaptive concurrency controller
5. Collapse two-phase work model + upload optimization
6. Managed thread runtime + download-to-file

## Why

Fixed concurrency doesn't adapt. A value tuned for a c6in.16xlarge (100 Gbps NIC, 64 vCPU) overshoots on a c5.xlarge (10 Gbps, 4 vCPU), and a conservative default leaves throughput on the table on larger instances. The customer either tunes manually or accepts suboptimal performance.

CRT derives both connection count and request concurrency from a throughput target at client creation. Connection count is [`ceil(target_gbps / 0.4)`](https://github.com/awslabs/aws-c-s3/blob/v0.12.0/source/s3_client.c#L163), and max requests in flight is [`connections * 4`](https://github.com/awslabs/aws-c-s3/blob/v0.12.0/source/s3_client.c#L211). For a 100 Gbps target: 250 connections, 1000 max in-flight requests. These values are computed once at client creation and don't change. Better than a blind default, but still static. It doesn't adapt to competing processes, NIC sharing, or S3 backend variability.

The adaptive controller observes actual network goodput and adjusts concurrency to maximize it, without customer configuration.

### Goodput measurement

The scheduler records bytes transferred on every work completion into `IOCounters`, a set of sliding-window byte counters (one per I/O direction: network tx/rx, disk read/write). The controller reads aggregate network throughput from these counters at evaluation boundaries. Because the counters accumulate over a 500ms sliding window, individual request variance is smoothed out.

Raw throughput measurements are further smoothed through a **windowed max filter** inspired by [TCP BBR](https://research.google/pubs/bbr-congestion-based-congestion-control/)'s bandwidth estimation. The algorithm sees the maximum of the last 4 raw measurements (~2 seconds). A single low measurement (common at transfer boundaries when one transfer finishes and the next ramps up) doesn't affect the max. Sustained throughput changes are tracked as old samples age out of the window.

### Three-state algorithm

The algorithm uses a three-state machine (SlowStart, StableProbing, StableShedding) with an exercised check and asymmetric acceptance (aggressive upward, conservative downward). We added windowed max smoothing and error-aware transitions on top.

```
                 ┌─────────────┐
                 │  SlowStart  │
                 │ (+32/window)│
                 └──────┬──────┘
                        │ 4 consecutive
                        │ rejections
                        ▼
               ┌─────────────────┐
           ┌──>│ StableProbing   │
           │   │ (+2% upward)    │
           │   └────────┬────────┘
           │            │ 3 consecutive
           │            │ rejections
           │            ▼
           │   ┌─────────────────┐
           └───┤ StableShedding  │
          1    │ (-2%, -4%, -6%  │
        reject │  accelerating)  │
               └─────────────────┘
```

**SlowStart** (300ms evaluation interval): Additive growth, +32 concurrency per accepted window. Throughput must improve by ≥10% over the best recent measurement to accept. Requires 4 consecutive rejections before exiting. This tolerates noisy measurements during connection warmup, where TLS handshakes and TCP slow start cause throughput to lag behind concurrency increases.

**StableProbing** (500ms evaluation interval): Small upward probes, +2% of current concurrency. Must beat the best recent measurement by ≥2%. Comparing against the best (not just previous) prevents noise-driven acceptance. After 3 consecutive rejections, transitions to StableShedding.

**StableShedding**: Accelerating downward probes (-2%, -4%, -6%...). Accepts while throughput stays within 2% of peak (max of recent 10 measurements, a higher bar than probing). On rejection (throughput dropped too far), transitions back to StableProbing. History is preserved across StableProbing/StableShedding transitions since they're fine-tuning around the same operating point.

### Exercised check

Before evaluating a probe, the controller checks whether the probed concurrency was actually used. If peak in-flight during the window didn't reach the accepted concurrency level, the throughput measurement doesn't reflect the probed level. The probe is skipped and re-tried. This prevents false positives from accepting a concurrency level that was never tested, e.g. when memory pressure or transfer completion limited actual in-flight work.

### Error-aware transitions

S3 throttle responses (503, 429, SlowDown) bypass the evaluation timer and transition immediately to StableShedding. The error classification infrastructure (`classify_error`, `ErrorKind`) walks the SDK error source chain to identify throttle vs transient vs permanent errors.

### Lock-free hot path

`target()` is a single `AtomicUsize` load, called by the scheduler on every `generate_work` cycle. `on_completion()` does two atomic operations (decrement in-flight, check for throttle) then `try_lock` on the algorithm state. If another thread holds the lock, the completion returns immediately without blocking. The algorithm runs at window boundaries (~every 500ms), not on every completion.

## What's in the change

```
scheduler/concurrency/
  mod.rs        - ConcurrencyController trait, FixedConcurrency, CompletionSample,
                  ErrorKind, classify_error (existing, moved to directory module)
  adaptive.rs   - AdaptiveConcurrencyController, AdaptiveConfig, three-state algorithm
                  (new, ~500 lines including 19 tests)

client.rs       - Auto mode creates AdaptiveConcurrencyController (was FixedConcurrency)
```

## Defaults

| Parameter | Value | Rationale |
|-----------|-------|-----------|
| initial_concurrency | 32 | Low enough for small instances; ramp finds the rest |
| SlowStart evaluation | 300ms | Faster ramp during warmup |
| Stable evaluation | 500ms | Responsive without being noisy |
| max_growth (SlowStart) | +32 | Limits overshoot per step |
| acceptance_margin (SlowStart) | 10% | Must beat best-seen by this much |
| exit_threshold (SlowStart) | 4 | Consecutive rejections; tolerates warmup noise |
| probe_pct (Stable) | 2% | Small enough to not disrupt, large enough to detect |
| acceptance_margin (Stable) | 2% | Tight, fine-tuning not exploring |
| shedding_transition_threshold | 3 | Consecutive rejections before shedding |
| history_size | 10 | ~5s of history at 500ms evaluation |
| smoothing_window | 4 | ~2s windowed max; filters transfer boundary gaps |

## Known limitations and future work

- **No re-entry to SlowStart.** Once exited, the controller never returns to SlowStart. If conditions change fundamentally, StableProbing's +2% crawl is slow to find the new optimum. A possible trigger for re-entry: sustained throughput well below historical peak.
- **Many-small-transfers workloads.** The windowed max filter assumes transfer boundary gaps are short relative to the 2s smoothing window. Workloads with many small transfers (frequent boundaries) may see sustained depressed throughput that the algorithm interprets as a real signal rather than inter-transfer gaps.
- **Connection warmth.** The controller can't distinguish "throughput is low because connections are cold" from "throughput is low because we're at the optimum." The 4-rejection exit threshold in SlowStart provides buffer but doesn't fully solve this. Future: integrate with the connection pool to observe establishment timings and connection rates. A cold pool (no warm connections to the target endpoint) should re-enter SlowStart rather than probing from wherever the previous transfer left off.
- **Instance-aware initial concurrency.** Starting at 32 on a c6in.16xlarge means ~4 seconds of suboptimal throughput during SlowStart. Future: detect instance type or NIC bandwidth and set a higher initial concurrency to skip most of the ramp.

## Tests

19 unit tests covering:
- Bootstrap behavior (first measurement)
- SlowStart acceptance, rejection, additive growth, exit on plateau
- StableProbing acceptance, rejection, transition to shedding
- StableShedding acceptance (near peak), rejection (below peak), transition back to probing
- Throttle error forces immediate shedding
- Max concurrency clamping, min probe size enforcement
- Exercised check (rejects unexercised, passes exercised)
- Idle window skips evaluation
- History cleared on SlowStart exit, preserved between stable phases
- No creep on flat throughput
- Peak knowledge retained across phase transitions

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/client.rs
+++ b/aws-sdk-s3-transfer-manager/src/client.rs
@@ -75,6 +75,7 @@ impl Client {
         let controller: Arc<dyn crate::scheduler::ConcurrencyController> =
             match config.concurrency() {
                 ConcurrencyMode::Explicit(c) => Arc::new(FixedConcurrency::new(*c)),
+                // TODO(vnext): implement support for target throughput
                 _ => Arc::new(AdaptiveConcurrencyController::new(
                     AdaptiveConfig::default(),
                     io_counters.clone(),
--- a/aws-sdk-s3-transfer-manager/src/client.rs
+++ b/aws-sdk-s3-transfer-manager/src/client.rs
@@ -70,16 +70,22 @@ impl Client {
         };
         use std::time::Duration;
 
-        let io_counters = Arc::new(IOCounters::new(Duration::from_millis(500)));
-
-        let controller: Arc<dyn crate::scheduler::ConcurrencyController> =
+        let (controller, io_counters): (Arc<dyn crate::scheduler::ConcurrencyController>, _) =
             match config.concurrency() {
-                ConcurrencyMode::Explicit(c) => Arc::new(FixedConcurrency::new(*c)),
+                ConcurrencyMode::Explicit(c) => (
+                    Arc::new(FixedConcurrency::new(*c)),
+                    Arc::new(IOCounters::new(Duration::from_millis(500))),
+                ),
                 // TODO(vnext): implement support for target throughput
-                _ => Arc::new(AdaptiveConcurrencyController::new(
-                    AdaptiveConfig::default(),
-                    io_counters.clone(),
-                )),
+                _ => {
+                    let adaptive_config = AdaptiveConfig::default();
+                    let io_counters = Arc::new(IOCounters::new(adaptive_config.window.duration));
+                    let controller = Arc::new(AdaptiveConcurrencyController::new(
+                        adaptive_config,
+                        io_counters.clone(),
+                    ));
+                    (controller, io_counters)
+                }
             };
 
         let new_scheduler = SchedulerBuilder::new(controller, io_counters)
--- a/aws-sdk-s3-transfer-manager/src/client.rs
+++ b/aws-sdk-s3-transfer-manager/src/client.rs
@@ -64,13 +64,25 @@ impl Handle {
 impl Client {
     /// Creates a new client from a transfer manager config.
     pub fn new(config: Config) -> Client {
-        use crate::scheduler::FixedConcurrency;
-        let new_scheduler = crate::scheduler::Scheduler::with_controller(Arc::new(
-            FixedConcurrency::new(match config.concurrency() {
-                ConcurrencyMode::Explicit(c) => *c,
-                _ => crate::DEFAULT_CONCURRENCY,
-            }),
-        ));
+        use crate::metrics::IOCounters;
+        use crate::scheduler::{
+            AdaptiveConcurrencyController, AdaptiveConfig, FixedConcurrency, SchedulerBuilder,
+        };
+        use std::time::Duration;
+
+        let io_counters = Arc::new(IOCounters::new(Duration::from_millis(500)));
+
+        let controller: Arc<dyn crate::scheduler::ConcurrencyController> =
+            match config.concurrency() {
+                ConcurrencyMode::Explicit(c) => Arc::new(FixedConcurrency::new(*c)),
+                _ => Arc::new(AdaptiveConcurrencyController::new(
+                    AdaptiveConfig::default(),
+                    io_counters.clone(),
+                )),
+            };
+
+        let new_scheduler = SchedulerBuilder::new(controller, io_counters)
+            .build(|scheduler| Arc::new(crate::runtime::TokioMultiThreadRuntime::new(scheduler)));
         let legacy_scheduler = LegacyScheduler::new(config.concurrency().clone());
         let handle = Arc::new(Handle {
             config,
--- a/aws-sdk-s3-transfer-manager/src/scheduler/concurrency/adaptive.rs
+++ b/aws-sdk-s3-transfer-manager/src/scheduler/concurrency/adaptive.rs
@@ -1,0 +1,889 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Adaptive concurrency controller.
+//!
+//! Three-state machine that discovers and maintains the optimal concurrency
+//! for the current network conditions.
+//!
+//! ## Throughput Measurement
+//!
+//! Throughput comes from shared [`IOCounters`] sliding windows. Raw measurements
+//! are smoothed through a windowed max filter (inspired by BBR's bandwidth
+//! estimation) before reaching the algorithm. A single low measurement — common
+//! at transfer boundaries — doesn't affect the max. Sustained throughput changes
+//! are tracked as old samples age out of the window.
+//!
+//! ## Phases
+//!
+//! **SlowStart**: Additive growth (+`max_growth` per evaluation). Compares
+//! against the best recent measurement; requires consecutive rejections before
+//! exiting to ride through noisy warmup windows.
+//!
+//! **StableProbing**: Small upward probes (+2%). Must beat the best recent
+//! measurement by 2%. Three consecutive rejections trigger StableShedding.
+//!
+//! **StableShedding**: Accelerating downward probes (-2%, -4%, -6%...).
+//! Accepts while throughput stays within 2% of peak. Rejects back to
+//! StableProbing when throughput drops too far.
+//!
+//! ## Error-Aware Transitions
+//!
+//! Throttle errors (503, 429, SlowDown) bypass the evaluation timer and
+//! transition immediately to StableShedding.
+
+use std::collections::VecDeque;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+
+use super::{CompletionSample, ConcurrencyController, ErrorKind};
+use crate::metrics::IOCounters;
+
+/// Configuration for the adaptive concurrency controller.
+#[derive(Debug, Clone)]
+pub(crate) struct AdaptiveConfig {
+    pub initial_concurrency: usize,
+    pub max_concurrency: Option<usize>,
+    pub window: WindowConfig,
+    pub slow_start: SlowStartConfig,
+    pub stable: StableConfig,
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct WindowConfig {
+    /// Evaluation interval. The controller checks throughput this often.
+    pub duration: Duration,
+    /// Number of raw measurements to retain for windowed-max smoothing.
+    /// The algorithm sees the max of the last N measurements, filtering
+    /// transient drops (transfer boundary gaps, warmup noise).
+    pub smoothing_window: usize,
+}
+
+/// SlowStart phase settings.
+#[derive(Debug, Clone)]
+pub(crate) struct SlowStartConfig {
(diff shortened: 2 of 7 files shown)


aws-sdk-s3-transfer-manager/src/scheduler/concurrency/adaptive.rs
```diff
@@ -0,0 +1,889 @@
...
+    fn target(&self) -> usize {
+        self.target.load(Ordering::Relaxed)
+    }
+
+    fn on_dispatch(&self) {
+        let n = self.in_flight.fetch_add(1, Ordering::Relaxed) + 1;
+        self.peak_in_flight.fetch_max(n, Ordering::Relaxed);
+    }
+
+    fn on_completion(&self, sample: &CompletionSample) {
```
landonxjames, line 491: PR says:
> the completion returns immediately without blocking

But I'm not sure that is correct. The `self.state.lock()` could block on the throttling path if we are waiting on other threads to finish running the algorithm. Ex: a burst of 503s could cause worker threads to back up waiting on the mutex. `self.maybe_update_window()` uses `try_lock` to return early, maybe that would work here as well?

aws-sdk-s3-transfer-manager/src/scheduler/concurrency/adaptive.rs
```diff
@@ -0,0 +1,889 @@
...
+        self.in_flight.fetch_sub(1, Ordering::Relaxed);
```
landonxjames: Could this underflow? Should it saturate at 0?

aws-sdk-s3-transfer-manager/src/client.rs
```diff
@@ -64,13 +64,25 @@ impl Handle {
...
+            AdaptiveConcurrencyController, AdaptiveConfig, FixedConcurrency, SchedulerBuilder,
+        };
+        use std::time::Duration;
+
+        let io_counters = Arc::new(IOCounters::new(Duration::from_millis(500)));
+
+        let controller: Arc<dyn crate::scheduler::ConcurrencyController> =
+            match config.concurrency() {
+                ConcurrencyMode::Explicit(c) => Arc::new(FixedConcurrency::new(*c)),
+                _ => Arc::new(AdaptiveConcurrencyController::new(
```
landonxjames: `ConcurrencyMode::TargetThroughput` still exists, should it be falling through to the `AdaptiveConcurrencyController`?

aws-sdk-s3-transfer-manager/src/scheduler/concurrency/adaptive.rs
```diff
@@ -0,0 +1,889 @@
...
+    StableShedding = 2,
+}
+
+impl Phase {
+    fn from_usize(v: usize) -> Self {
+        match v {
+            0 => Phase::SlowStart,
+            1 => Phase::StableProbing,
+            2 => Phase::StableShedding,
+            _ => Phase::StableProbing,
```
landonxjames, line 154: Is the fallback to `Phase::StableProbing` correct? Feel like this should be an error of some kind

aws-sdk-s3-transfer-manager/src/scheduler/concurrency/adaptive.rs
```diff
@@ -0,0 +1,889 @@
...
+    /// windowed max of raw measurements at that evaluation point.
+    recent_goodput: VecDeque<f64>,
+    /// Raw goodput measurements for windowed-max smoothing. Transient drops
+    /// don't affect the max; sustained drops age out naturally.
+    raw_goodput: VecDeque<f64>,
+    /// Consecutive probe results in the same direction within the current phase.
+    /// SlowStart: consecutive rejections (triggers exit at threshold).
+    /// StableProbing: consecutive rejections (triggers shedding at threshold).
+    /// StableShedding: consecutive accepts (accelerates reduction).
+    consecutive_probes: usize,
```
landonxjames, line 178: The overloading of this field feels weird. Counting both rejections and accepts in different phases seems like it could cause confusion. Could there be a different field for each?

aws-sdk-s3-transfer-manager/src/scheduler/concurrency/adaptive.rs
```diff
@@ -0,0 +1,889 @@
...
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+
+use super::{CompletionSample, ConcurrencyController, ErrorKind};
+use crate::metrics::IOCounters;
+
+/// Configuration for the adaptive concurrency controller.
+#[derive(Debug, Clone)]
+pub(crate) struct AdaptiveConfig {
+    pub initial_concurrency: usize,
```
landonxjames: `pub(crate)`?

landonxjames (approved):
Mercifully short 🙏

aws-sdk-s3-transfer-manager/src/scheduler/concurrency/adaptive.rs
aajtodd: This is true but this is on the throttle path not the normal hot path. Even on a throttle we have a trivial critical section and worse case they have contention on a throttle which causes further slow down. I think I'm going to leave it for now.

aws-sdk-s3-transfer-manager/src/client.rs
aajtodd: Added a todo, yes this will end up some kind of constraint on the adaptive controller but it's not implemented yet.

aws-sdk-s3-transfer-manager/src/scheduler/concurrency/adaptive.rs
aajtodd: It's just defense in depth. On the chance of a corrupted atomic it's still correct/safe to default to stable probing which if "wrong" will bounce back to stable shedding.

aws-sdk-s3-transfer-manager/src/scheduler/concurrency/adaptive.rs
aajtodd: There could be but it's reset to 0 on phase transition. Within each phase it has a single meaning. Having separate counters is just more state to know "which one to check? did we reset them all? etc". Not a clear win to me, leaving.

ysaito1001 (approved):
Nice! TIL: oscillating between Probing and Shedding to hunt the optimal point.

aws-sdk-s3-transfer-manager/src/client.rs
```diff
@@ -64,13 +64,26 @@ impl Handle {
...
-                _ => crate::DEFAULT_CONCURRENCY,
-            }),
-        ));
+        use crate::metrics::IOCounters;
+        use crate::scheduler::{
+            AdaptiveConcurrencyController, AdaptiveConfig, FixedConcurrency, SchedulerBuilder,
+        };
+        use std::time::Duration;
+
+        let io_counters = Arc::new(IOCounters::new(Duration::from_millis(500)));
```
vcjana: `AdaptiveConfig` defines `window_config.duration: 500ms` and `slow_start.evaluation_interval: 300ms` separately does this 500ms need to stay in sync with those?  or could this be derived from the config?

vcjana (approved):
Overall LGTM. just had one small question

aws-sdk-s3-transfer-manager/src/client.rs
aajtodd: Yeah they should match the evaluation interval. It's right in the redux branch but wrong here, fixed.

PR #136, opened 2026-04-02
collapse work model for vnext

aajtodd:
## Summary

Collapse the two-phase work model into single logical operations and add per-operation telemetry with stall detection. An upload part does read+send in one `execute()` call. A download discovery reads its response body inline instead of scheduling follow-on work. The scheduler no longer tracks work phases or dispatches follow-on items. Transfers record throughput and latency directly to a shared `Telemetry` struct, and SDK calls are wrapped in adaptive timeout+retry to mitigate straggler requests.

This is PR 5 in the redesign series:

4. ~~Adaptive concurrency controller~~ (#135)
5. **This PR**: Collapse two-phase work model, telemetry, stall detection
6. Managed thread runtime + download-to-file

## Why

The two-phase model split each upload part into a Disk work item (read from file) and a Network work item (send to S3), connected by follow-on dispatch in the scheduler. This added complexity without benefit:

- The scheduler tracked `IoKind` (Disk/Network) on every work item and `schedule_next` on every completion, but used a single capacity gate regardless of kind.
- Follow-on dispatch added a re-enqueue hop between phases. Read data sat in memory waiting for the scheduler to dispatch the send.
- `CompletionSample` carried per-phase metrics (`IoSample` with bytes, duration, kind) that the adaptive controller never uses. It only reads aggregate goodput from `IOCounters`.
- Download discovery was split into a `Discovery` request and a `ReadDiscoveryBody` follow-on, even though the body is available immediately after the GET response.

Collapsing also simplifies the path to managed threads in the next PR, where each thread runs the full pipeline on one core. Splitting into phases would require cross-thread handoff or sticky routing.

Separately, S3 requests occasionally straggle (5+ seconds for an 8 MB part that normally completes in ~120ms). Without detection, everything waits for the last part. The `LatencyTracker` observes per-operation P99 latency and `guarded()` wraps SDK calls with an adaptive timeout: cancel and retry on a fresh connection when a request exceeds 2x the observed P99.

### Work model types (`transfer.rs`)

- Removed `IoKind` enum
- Removed `kind` field from `IoRequest`
- `WorkOutcome::Success`: removed `schedule_next` and `metrics` fields, now carries only `data`
- `CompletionSample`: removed `io` and `kind` fields, now carries only `error`

### Telemetry (`telemetry.rs`, `metrics/latency.rs`)

- `Telemetry` struct groups `send_latencies`, `recv_latencies` (both `LatencyTracker`), and `io_counters` (`Arc<IOCounters>`)
- `Telemetry` lives on `Handle`, accessible to all transfers
- `LatencyTracker` uses an HdrHistogram to compute P99-based adaptive deadlines
- `guarded()` wraps a future with timeout+retry (up to 3 attempts). On timeout, the in-flight future is dropped (tearing down the HTTP connection), and the closure is called again to build a fresh request
- Cold start: no timeout until `WARM_THRESHOLD` (10) samples collected
- Escape hatch: if average latency exceeds 5s, disable timeout (retry costs more than waiting)
- Rate-based backoff: +100ms at >0.1% timeout rate, +1s at >1% (matches CRT)

### Scheduler (`scheduler.rs`)

- Removed follow-on dispatch from `on_completion`. Previously, a successful completion with `schedule_next: Some(kind)` triggered immediate dispatch of a follow-on work item. Now completions free capacity and trigger `generate_work`.
- Removed `io_counters` from scheduler. Transfers record bytes directly to `handle.telemetry.io_counters`.

### Upload (`operation/upload/`)

- Extracted `UploadState` to `context.rs`
- `UploadWork::UploadPart` no longer carries `part_data`. The part number is sufficient; data is read during execute.
- `execute_upload_part` does read+send in one call. Previously `execute_read_part` returned `Success { schedule_next: Some(Network) }` with the read data, then `execute_send_part` sent it.
- Upload part send wrapped in `send_latencies.guarded()` for stall detection
- Records `network_tx` bytes to `io_counters` after successful send

### Download (`operation/download/`)

- Extracted `DownloadState` to `context.rs`
- Removed `ReadDiscoveryBody` variant from `DownloadWork`. The discovery response body is now read inline within `execute_discovery`.
- Renamed `ObjectMetadata::content_length()` to `total_object_size()`. The previous name was misleading: for ranged responses, Content-Length is the range size, not the object size. The value comes from the Content-Range header's total field.
- Range request wrapped in `recv_latencies.guarded()` for stall detection
- Records `network_rx` bytes to `io_counters` after successful receive

### Part reader (`io/part_reader.rs`)

`PathBodyPartReader` now opens the file once at construction and holds `Arc<File>`. Previously each `next_part` call cloned the path and opened the file inside `spawn_blocking`. Concurrent part reads share the fd via positioned I/O (`read_exact_at`).

### Utilities

- Added `parse_content_range()` to `http/header.rs`
- `into_byte_stream()` for file-backed streams now preserves offset and length for partial file reads

### CI

Clippy set to `continue-on-error`. Pre-existing dead code warnings from the legacy middleware and token bucket scheduler cause failures with `-Dwarnings`. Temporary until those modules are removed in the next PR.

## What changed

```
transfer.rs                          - IoKind removed, WorkOutcome/IoRequest simplified
telemetry.rs                         - New: Telemetry struct (latency trackers + io_counters)
metrics/mod.rs                       - Restructured from metrics.rs, added latency submodule
metrics/latency.rs                   - New: LatencyTracker, guarded(), adaptive deadlines
scheduler/concurrency/mod.rs         - CompletionSample simplified
scheduler/scheduler.rs               - Follow-on dispatch removed, io_counters removed
operation/upload/context.rs          - New: UploadState enum
operation/upload/transfer.rs         - Single-phase execute, guarded() send, io_counters recording
operation/download/context.rs        - New: DownloadState enum
operation/download/transfer.rs       - Inline discovery body, guarded() receive, io_counters recording
operation/download/object_meta.rs    - Renamed to total_object_size(), uses parse_content_range
operation/download/discovery.rs      - Caller update
io/part_reader.rs                    - Shared Arc<File>, open once at construction
io/stream.rs                         - into_byte_stream offset/length fix
http/header.rs                       - parse_content_range()
client.rs                            - Telemetry on Handle, io_counters wiring
Cargo.toml                           - hdrhistogram dependency
.github/workflows/ci.yml            - Clippy soft failure (temporary)
```

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/io/part_reader.rs
+++ b/aws-sdk-s3-transfer-manager/src/io/part_reader.rs
@@ -47,7 +47,7 @@ impl Builder {
         self
     }
 
-    pub(crate) fn build(self) -> PartReader {
+    pub(crate) fn build(self) -> Result<PartReader, crate::error::Error> {
         let stream = self.stream.expect("input stream set");
         PartReader::new(stream, self.part_size)
     }
@@ -60,15 +60,15 @@ pub(crate) struct PartReader {
 }
 
 impl PartReader {
-    fn new(raw: RawInputStream, part_size: usize) -> Self {
+    fn new(raw: RawInputStream, part_size: usize) -> Result<Self, crate::error::Error> {
         let inner = match raw {
             RawInputStream::Buf(buf) => Inner::Bytes(BytesPartReader::new(buf)),
-            RawInputStream::Fs(path_body) => Inner::Fs(PathBodyPartReader::new(path_body)),
+            RawInputStream::Fs(path_body) => Inner::Fs(PathBodyPartReader::new(path_body)?),
             RawInputStream::Dyn(box_body) => Inner::Dyn(DynPartReader::new(box_body)),
         };
 
         let stream_cx = StreamContext::new(part_size);
-        Self { inner, stream_cx }
+        Ok(Self { inner, stream_cx })
     }
 
     pub(crate) fn part_size(&self) -> usize {
@@ -187,21 +187,20 @@ struct PathBodyPartReader {
 }
 
 impl PathBodyPartReader {
-    fn new(body: PathBody) -> Self {
+    fn new(body: PathBody) -> Result<Self, crate::error::Error> {
         let offset = body.offset;
         let content_length = body.length;
-        // Open the file eagerly so all part reads share a single fd.
-        // Panics if the file cannot be opened — callers should validate paths
-        // before constructing an InputStream.
-        let file = Arc::new(
-            File::open(&body.path)
-                .unwrap_or_else(|e| panic!("failed to open {}: {e}", body.path.display())),
-        );
-        Self {
+        let file = Arc::new(File::open(&body.path).map_err(|e| {
+            crate::error::Error::new(
+                crate::error::ErrorKind::IOError,
+                format!("failed to open {}: {e}", body.path.display()),
+            )
+        })?);
+        Ok(Self {
             body,
             state: Mutex::new(PartReaderState::new(content_length).with_offset(offset)),
             file,
-        }
+        })
     }
 }
 
@@ -379,7 +378,7 @@ mod test {
         let data = Bytes::from("a lep is a ball, a tay is a hammer, a flix is a comb");
         let stream = InputStream::from(data.clone());
         let expected = data.chunks(5).collect::<Vec<_>>();
-        let reader = Builder::new().part_size(5).stream(stream).build();
+        let reader = Builder::new().part_size(5).stream(stream).build().unwrap();
         let parts = collect_parts(reader).await;
         let actual = parts.iter().map(|p| p.data.chunk()).collect::<Vec<_>>();
 
@@ -406,7 +405,7 @@ mod test {
         let expected = data.chunks(part_size).collect::<Vec<_>>();
 
         let stream = builder.build().unwrap();
-        let reader = Builder::new().part_size(part_size).stream(stream).build();
+        let reader = Builder::new().part_size(part_size).stream(stream).build().unwrap();
 
         let parts = collect_parts(reader).await;
         let actual = parts.iter().map(|p| p.data.chunk()).collect::<Vec<_>>();
@@ -477,7 +476,7 @@ mod test {
                 .collect::<Vec<_>>(),
         );
         let stream = InputStream::from_part_stream(stream);
-        let reader = Builder::new().part_size(5).stream(stream).build();
+        let reader = Builder::new().part_size(5).stream(stream).build().unwrap();
         let parts = collect_parts(reader).await;
         let actual = parts.iter().map(|p| p.data.chunk()).collect::<Vec<_>>();
         assert_eq!(expected, actual);
@@ -524,7 +523,7 @@ mod test {
             offset: 10,
             length: 20,
         };
-        let reader = PathBodyPartReader::new(path_body);
+        let reader = PathBodyPartReader::new(path_body).unwrap();
         let stream_cx = StreamContext::new(5);
 
         // First advance should succeed
@@ -562,7 +561,7 @@ mod test {
             offset: 0,
             length: 3,
         };
-        let reader = PathBodyPartReader::new(path_body);
+        let reader = PathBodyPartReader::new(path_body).unwrap();
         let stream_cx = StreamContext::new(10);
 
         let result = reader.advance(&stream_cx).unwrap().unwrap();
--- a/aws-sdk-s3-transfer-manager/src/io/part_reader.rs
+++ b/aws-sdk-s3-transfer-manager/src/io/part_reader.rs
@@ -3,8 +3,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 use std::cmp;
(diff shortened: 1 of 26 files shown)


aws-sdk-s3-transfer-manager/src/telemetry.rs
```diff
@@ -25,3 +30,36 @@ pub(crate) const TARGET_EXECUTION: &str = "aws_s3_transfer_manager::execution";
...
+        }
+    }
+}
+
+impl std::fmt::Debug for Telemetry {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Telemetry")
+            .field("send_latencies", &self.send_latencies)
+            .field("recv_latencies", &self.recv_latencies)
+            .finish()
```
ysaito1001: `io_counters` missing?

aws-sdk-s3-transfer-manager/src/metrics/latency.rs
```diff
@@ -0,0 +1,430 @@
...
+            sample_count: AtomicUsize::new(0),
+            timeout_count: AtomicUsize::new(0),
+        }
+    }
+
+    /// Record a completed request duration.
+    pub(crate) fn record(&self, duration: Duration) {
+        let mut hist = self.hist.lock().unwrap();
+        // Clamp to histogram range; values outside are recorded at the boundary.
+        let _ = hist.record(duration.as_micros() as u64);
```
vcjana: if duration is over 60s this silently drops the sample but still bumps `sample_count`. should the count only increment on a successful record? Or use `record_correct()` which auto-clamps?

aws-sdk-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -173,20 +174,33 @@ impl BytesPartReader {
...
 impl PathBodyPartReader {
     fn new(body: PathBody) -> Self {
         let offset = body.offset;
         let content_length = body.length;
+        // Open the file eagerly so all part reads share a single fd.
+        // Panics if the file cannot be opened — callers should validate paths
+        // before constructing an InputStream.
+        let file = Arc::new(
+            File::open(&body.path)
+                .unwrap_or_else(|e| panic!("failed to open {}: {e}", body.path.display())),
```
vcjana: file could become unreadable between the metadata check in`build()` and this open. would it make sense to make this fallible instead of panicking?

vcjana (approved):
LGTM. Left a few minor comments

aws-sdk-s3-transfer-manager/src/io/part_reader.rs
aajtodd: Yes and this is right on the redux branch, just missed in translation trying to PR it back to vnext. Good catch.

aws-sdk-s3-transfer-manager/src/metrics/latency.rs
aajtodd: Updated to clamp

aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
```diff
@@ -417,56 +353,66 @@ impl DownloadTransfer {
...
+                }
+
+                async move {
+                    let resp = req.send().await.map_err(crate::error::Error::from)?;
+                    validate_content_range(seq, &rh, resp.content_range())?;
+                    let mut resp = resp;
+                    let body_stream = std::mem::replace(
+                        &mut resp.body,
+                        aws_sdk_s3::primitives::ByteStream::new(
+                            aws_smithy_types::body::SdkBody::empty(),
```
ysaito1001: nit: The performance difference is negligible, but I think `SdkBody::taken()` fits better?

aws-sdk-s3-transfer-manager/src/metrics/latency.rs
```diff
@@ -0,0 +1,430 @@
...
+            //   false-positive but catches truly stuck connections
+            // - First-byte timeout separate from total request timeout
+            // - Timeout derived from part_size / expected_throughput
+            return None;
+        }
+
+        let hist = self.hist.lock().unwrap();
+        let avg = Duration::from_micros(hist.mean() as u64);
+        if avg > RETRY_COST_THRESHOLD {
+            return None;
```
ysaito1001, line 129: Do we want to log something when this "escape hatch" kicks in?

aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
```diff
@@ -417,56 +353,66 @@ impl DownloadTransfer {
...
-            .key(input.key().unwrap_or_default())
-            .range(&range_header);
-
-        if let Some(ref etag) = etag {
-            req = req.if_match(etag.as_ref());
-        }
+            .handle
+            .telemetry
+            .recv_latencies
+            .guarded(|| {
```
ysaito1001, line 362: This is cool

ysaito1001 (approved):
Fantastic PR 🚀

PR #137, opened 2026-04-07
managed runtime support

aajtodd:
## Summary

Add the managed thread runtime (per-core execution with per-thread HTTP clients), download-to-file support, and remove the legacy middleware and scheduler infrastructure. After this PR, vnext has full feature parity with the development branch.

This is the final PR in the redesign series:

4. ~~Adaptive concurrency controller~~ (#135)
5. ~~Collapse two-phase work model, telemetry, stall detection~~ (#136)
6. **This PR**: Managed thread runtime, download-to-file, legacy cleanup

## Why

PR 2 introduced the `ExecutionRuntime` trait and `TokioMultiThreadRuntime` — work dispatched via `tokio::spawn` onto the default multi-threaded runtime. This works but has limitations: tasks land on arbitrary worker threads, connections migrate between threads (cross-thread wakeup overhead), and there's no control over thread-to-core or thread-to-NIC affinity.

`ManagedThreadRuntime` is the second implementation. Each CPU gets a dedicated OS thread running a tokio current-thread runtime. Tasks stay on the thread they're dispatched to. Each thread owns its own HTTP connection pool bound to its reactor — no cross-thread connection migration, no shared pool contention. This is the foundation for NUMA-aware execution and per-NIC connection binding (structurally supported, deferred until benchmarked).

Download-to-file was the other missing capability. Previously, downloads always streamed through `Body` and the consumer wrote to disk. Now the slot buffer can flush completed chunks directly to disk via positioned writes, and `ManagedDownloadHandle` manages the temp-file lifecycle.

The legacy tower-based middleware (hedge, concurrency limit) and old runtime infrastructure (scheduler, token bucket) are fully replaced by the new scheduler and adaptive controller from PRs 1-5.

## How to review

Suggested reading order:

1. **`runtime/mod.rs`** — start here. The `ExecutionRuntime` trait gains `components()` and the `dispatch` signature changes from single-item to batch (`SubmissionGuard`). `RuntimeComponents` carries runtime-provided HTTP client and direct I/O flag.
2. **`runtime/topology.rs`** — `Cpu`, `NumaNode`, `Topology`. Small, self-contained.
3. **`runtime/managed.rs`** — the new runtime. Read `ManagedThreadRuntimeBuilder::build()` for construction, `dispatch()` for the hot path, `execute_work()` for per-task execution. `ShufflingDnsResolver` and `DispatchRouter` are helpers at the top.
4. **`runtime/sync/submission.rs`** — `SubmissionQueue`. Read the module doc, then the `enter()`/`push()`/`submit()` protocol. Tests at the bottom cover edge cases.
5. **`client.rs`** + **`config.rs`** — construction order change. The runtime creates the S3 client now.
6. **`operation/download/body.rs`** — `Sink`, `try_flush()`, `finalize()`. The slot buffer gains file-write capability.
7. **`operation/download/handle.rs`** — `ManagedDownloadHandle`, `write_to_path()`, `write_to_file()`.
8. **`io/fs.rs`** — `write_all_at` (`pwritev` on unix, `seek_write` on windows).
9. Skim the deletions: `middleware/`, `runtime/scheduler.rs`, `runtime/token_bucket.rs`.

### Managed thread runtime

`ManagedThreadRuntimeBuilder::build()` spawns one OS thread per CPU in the topology. Each thread runs `tokio::runtime::Builder::new_current_thread()` with a `CancellationToken`-gated `block_on`. Work arrives via `tokio::runtime::Handle::spawn()` from the dispatch path.

Each managed thread gets its own `HyperClientBuilder` → TLS connector → connection pool. A shared `http_client_fn` closure dispatches to the correct per-thread client using a `MANAGED_THREAD_CPU` thread-local set during thread startup. The S3 client is constructed with this shared `http_client_fn`, so every SDK call routes to the calling thread's connection pool transparently.

`ShufflingDnsResolver` wraps the SDK's DNS resolver and shuffles returned IPs via `fastrand::shuffle`. Without this, hyper connects to the first resolved IP, clustering all connections on one S3 frontend.

`DispatchRouter` uses power-of-two-random-choices: pick two random threads, select the one with lower `in_flight` count. O(1) regardless of thread count.

Shutdown: cancel the shared token → each thread's `block_on` returns → join all handles.

### Submission queue

`SubmissionQueue` replaces single-item dispatch with batch dispatch between the scheduler and runtime. Fixed-capacity array with lock-free slot claiming (`AtomicUsize` fetch_add on tail) and `parking_lot::Mutex`+`Condvar` for round coordination. The scheduler's `generate_work` loop and follow-on dispatch path push work into the queue. The last producer to call `submit()` becomes the consumer and receives a `SubmissionGuard` with contiguous batch access. The runtime's `dispatch()` drains this guard.

### S3 client construction

Construction order changes: runtime first → `runtime.components()` → inject HTTP client into S3 client builder → build client → create scheduler with runtime. `S3ClientConfig` wraps `aws_sdk_s3::config::Builder`. Users who call `.client(my_s3_client)` bypass runtime HTTP optimization — that's the escape hatch.

`RuntimeComponents` carries:
- `http_client`: per-thread HTTP client (managed) or `None` (tokio-mt uses SDK default)
- `direct_io`: file I/O runs inline on the calling thread (managed) vs `spawn_blocking` (tokio-mt)

### Download-to-file

`Sink` attaches to the slot buffer at construction. After each `slot.fill()`, `try_flush()` scans for contiguous runs of FILLED slots and coalesces them into single `pwritev` calls. Two-phase flush: write bytes to disk, then mark slots FLUSHED and advance the consumed counter — the consumer sees data as consumed only after it's on disk. `finalize()` drains all remaining slots at transfer completion.

`ManagedDownloadHandle` wraps `DownloadHandle` with file lifecycle:
- `write_to_path(path)`: creates a temp file (`.s3tmp` suffix) in the same directory, preallocates via `posix_fallocate`/`set_len`, downloads into it, atomic-renames on success. Cleans up the temp file on failure or drop.
- `write_to_file(file)`: writes to a caller-provided `File`.

`io::fs::write_all_at` provides the platform-specific positioned write (`pwritev` on unix via `nix`, sequential `seek_write` on windows).

### Runtime

- `ExecutionRuntime` trait: `dispatch` signature changes from `dispatch(ScheduledWork)` to `dispatch(&mut SubmissionGuard<ScheduledWork>)`. Added `components() -> &RuntimeComponents`.
- `TokioMultiThreadRuntime`: updated for batch dispatch (drains guard into worker pool) and `RuntimeComponents` (defaults: no custom HTTP client, `direct_io = false`). Lazy worker scaling via `compare_exchange` when concurrency target grows.
- New: `ManagedThreadRuntime`, `ManagedThreadRuntimeBuilder`, `ShufflingDnsResolver`, `DispatchRouter`
- New: `Topology`, `Cpu`, `NumaNode`
- New: `SubmissionQueue`, `Submission`, `SubmissionGuard`
- New: `RuntimeComponents` (HTTP client, direct I/O flag)

### Download-to-file

- New: `Sink` on slot buffer with `try_flush()`, `flush_all()`, `finalize()`, `preallocate()`
- New: `ManagedDownloadHandle` with `write_to_path()`, `write_to_file()`
- New: `io::fs::write_all_at` (`pwritev` on unix, `seek_write` on windows)
- `ChunkOutput` gains `offset` field (byte offset in object)
- `Body` gains `close()` method

### Client and config

- `S3ClientConfig` newtype wrapping `aws_sdk_s3::config::Builder`
- `Handle` gains `s3_client` field (built by runtime), removes `config.client()` pattern
- Construction order: runtime → components → S3 client → scheduler
- `StreamContext` gains `direct_io` flag and `io_counters` reference

### Legacy removal

- Deleted `middleware/` (tower-based hedge and concurrency limit services)
- Deleted `runtime/scheduler.rs` (legacy task scheduler)
- Deleted `runtime/token_bucket.rs` (legacy rate limiter)
- Removed `async-trait`, `tower`, `aws-smithy-async` dependencies

### Other

- `disable_payload_signing()` added to PutObject send path
- Integration tests against mock S3 server (download write_to_path, upload, ranged download)
- `examples/cp.rs` updated with `write_to_path` support, iteration loop, JSON output
- Clippy fixes across the codebase
- All dependencies bumped to latest versions

## What changed

```
runtime/mod.rs                       - ExecutionRuntime gains components() + batch dispatch
runtime/managed.rs                   - New: ManagedThreadRuntime, per-thread HTTP, DNS shuffling
runtime/topology.rs                  - New: Cpu, NumaNode, Topology
runtime/sync/mod.rs                  - New: sync primitives module
runtime/sync/submission.rs           - New: SubmissionQueue with opportunistic batching
runtime/tokio_mt.rs                  - Updated for batch dispatch + RuntimeComponents
client.rs                            - S3ClientConfig, construction order, Handle.s3_client
config.rs                            - S3ClientConfig newtype, S3ClientSource enum
config/loader.rs                     - Builder takes S3ClientConfig
io/fs.rs                             - New: write_all_at (pwritev/seek_write)
io/mod.rs                            - Added fs module
io/part_reader.rs                    - StreamContext gains direct_io + io_counters
io/stream.rs                         - StreamContext constructor change
io/adapters.rs                       - StreamContext constructor change (tests)
operation/download/body.rs           - Sink, try_flush, finalize, preallocate, close
operation/download/handle.rs         - ManagedDownloadHandle, write_to_path, write_to_file
operation/download/transfer.rs       - Preallocate, try_flush after fill, finalize on complete
operation/download/builders.rs       - Download-to-file builder support
operation/download/chunk_meta.rs     - ChunkOutput.offset field
operation/upload/transfer.rs         - disable_payload_signing on PutObject, direct_io plumbing
scheduler/scheduler.rs               - SubmissionQueue integration, batch dispatch
scheduler/mod.rs                     - Re-exports updated
scheduler/concurrency.rs             - Replaces concurrency/mod.rs
scheduler/transfer.rs                - New: mock transfer module
scheduler/transfer/mock.rs           - New: MockTransfer for tests
scheduler/transfer/test_util.rs      - New: test Handle construction helpers
middleware.rs                        - Deleted
middleware/                          - Deleted (hedge, limit services)
runtime/scheduler.rs                 - Deleted (legacy scheduler)
runtime/token_bucket.rs              - Deleted (legacy rate limiter)
scheduler/test_util.rs               - Deleted (replaced by scheduler/transfer/)
scheduler/concurrency/mod.rs         - Deleted (replaced by scheduler/concurrency.rs)
Cargo.toml                           - +crossbeam-utils, parking_lot, nix, fastrand,
                                       aws-smithy-http-client, aws-smithy-dns
                                     - -async-trait, tower, aws-smithy-async
                                     - All deps bumped to latest
.github/workflows/ci.yml            - Toolchain updated, -Dwarnings relaxed temporarily
examples/cp.rs                       - write_to_path, iterations, JSON output
tests/download_test.rs               - test_download_write_to_path
integration-tests/                   - New: mock server integration tests
```

## Future
* NUMA awareness/pinning
* Coordinated connection pool (NIC to NUMA awareness)
* `SubmissionQueue` is structured to batch but we don't take advantage of it yet. It's meant to allow us to extend dispatch by grouping/routing intelligently (e.g. combine FD reads/writes, io_uring, etc).
* DNS - Shuffling is better than defaulting to the same (first) IP but this isn't likely what we want.


aws-sdk-s3-transfer-manager/src/operation/download/handle.rs
```diff
@@ -197,24 +234,144 @@ impl DownloadHandle {
...
+        let ctx = self.inner.transfer.ctx();
+        if ctx.is_active() {
+            ctx.set_cancelled();
+            self.inner.transfer.writer().notify_consumer();
+            ctx.handle
+                .scheduler
+                .cancel_transfer(self.inner.transfer.id());
+        }
+        if let Some(temp) = &self.temp_path {
+            let _ = std::fs::remove_file(temp);
```
landonxjames, line 367: Sync file cleanup feels wrong here. Should a task be spawned to handle it and avoid blocking?

aws-sdk-s3-transfer-manager/src/runtime/managed.rs
```diff
@@ -0,0 +1,490 @@
...
+
+        // Build an http_client_fn that dispatches to the per-thread HTTP client
+        // based on which managed thread is calling.
+        let per_thread_clients: Arc<Vec<SharedHttpClient>> =
+            Arc::new(threads.iter().map(|th| th.http_client.clone()).collect());
+
+        let shared_http_client = http_client_fn(move |settings, components| {
+            let cpu_index = MANAGED_THREAD_CPU
+                .with(|c| c.get())
+                .expect("http_client_fn called from non-managed thread");
```
landonxjames, line 255: Should this return an Error instead of panicing? Seems potentially recoverable

aws-sdk-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -174,28 +200,24 @@ impl BytesPartReader {
...
+        // order is sequential so kernel readahead would help, but concurrent execution creates
+        // some scatter. Benchmark before adding.
+        // TODO(vnext): does this need to be async now?
+        let file = Arc::new(File::open(&body.path)?);
         let offset = body.offset;
         let content_length = body.length;
-        let file = Arc::new(File::open(&body.path).map_err(|e| {
-            crate::error::Error::new(
-                crate::error::ErrorKind::IOError,
-                format!("failed to open {}: {e}", body.path.display()),
```
landonxjames, line 196: Probably worth adding this context back, helpful to know which file failed to open

aws-sdk-s3-transfer-manager/src/runtime/sync/submission.rs
```diff
@@ -0,0 +1,544 @@
...
+//!  │   0              tail (atomic)                          │
+//!  └─────────────────────────────────────────────────────────┘
+//! ```
+
+use std::cell::UnsafeCell;
+use std::mem::MaybeUninit;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+use crossbeam_utils::CachePadded;
+use parking_lot::{Condvar, Mutex};
```
landonxjames, line 46: Should we be using `parking_lot::Mutex` in more places instead of the `std::Mutex`? Guess we would need benchmarks to determine which is best in each application?

Edit: I guess what I'm looking for is the thought process between using `parking_lot` here and `std` elsewhere.

landonxjames (approved):
LGTM, few questions, no blockers

aws-sdk-s3-transfer-manager/src/runtime/sync/submission.rs
aajtodd: Yeah I was going to circle back and update them all as part of cleanup most likely (there's a handful of things that need circle backed to and updated now that we're further along). Though I think it breaks loom so I need to look closer.

aws-sdk-s3-transfer-manager/src/runtime/managed.rs
```diff
@@ -0,0 +1,490 @@
...
+                    .map(|n| n.get())
+                    .unwrap_or(1),
+            )
+        });
+        ManagedThreadRuntime::new(self.scheduler, topology, self.pin_threads)
+    }
+}
+
+impl ExecutionRuntime for ManagedThreadRuntime {
+    fn dispatch(&self, batch: &mut SubmissionGuard<'_, ScheduledWork>) {
```
vcjana, line 320: could there be a race with Drop here? If the managed thread's runtime is already gone, spawn silently drops the task and `in_flight` never decrements. will checking `is_cancelled()` before the loop help?

vcjana (approved):

aws-sdk-s3-transfer-manager/src/runtime/managed.rs
aajtodd: I think anything that breaks our assumptions/contracts we assert to be true should be an expect. I don't expect we ever end up here from an unmanaged thread because it shouldn't be possible. Also though the API here doesn't allow for an Error.

aws-sdk-s3-transfer-manager/examples/cp.rs
```diff
@@ -28,6 +24,20 @@ use jemallocator::Jemalloc;
 #[global_allocator]
 static GLOBAL: Jemalloc = Jemalloc;
 
+// #[cfg(not(target_env = "msvc"))]
+// use mimalloc::MiMalloc;
+//
+// #[cfg(not(target_env = "msvc"))]
+// #[global_allocator]
+// static GLOBAL: MiMalloc = MiMalloc;
```
ysaito1001: TODO for cleanup or will be used later?

aws-sdk-s3-transfer-manager/src/runtime/managed.rs
```diff
@@ -0,0 +1,490 @@
...
+                        // Create per-thread HTTP client on this thread's runtime.
+                        // The TLS connector and connection pool bind to this thread's reactor.
+                        let http_client = aws_smithy_http_client::Builder::new()
+                            .tls_provider(aws_smithy_http_client::tls::Provider::Rustls(
+                                aws_smithy_http_client::tls::rustls_provider::CryptoMode::AwsLc,
+                            ))
+                            // .tls_provider(aws_smithy_http_client::tls::Provider::S2nTls)
+                            .build_with_resolver(resolver);
+
+                        let _ = tx.send((rt.handle().clone(), http_client));
```
ysaito1001, line 222: Should we say something if `.send` fails?

aws-sdk-s3-transfer-manager/src/runtime/managed.rs
ysaito1001: Since `Drop` takes an exclusive reference `&mut self`, I _suspect_ no race is happening?

aws-sdk-s3-transfer-manager/src/operation/download/body.rs
```diff
@@ -94,7 +139,136 @@ impl SlotBuffer {
...
+        let mut i = consumed;
+        while i < claimed {
+            let idx = (i % self.capacity) as usize;
+            if self.slots[idx].state.load(AtomicOrdering::Acquire) != SLOT_FILLED {
+                i += 1;
+                continue;
+            }
+
+            // Start a new run with the first FILLED slot
+            let first_chunk = unsafe { (*self.slots[idx].data.get()).take().unwrap() };
```
ysaito1001, line 209: Small block but maybe consider adding `Safety` comments for `unsafe` (and other occurrences of bare `unsafe` blocks in this file, line 226 and 230)?

ysaito1001 (approved):
Love `How to review` in the PR description.

aws-sdk-s3-transfer-manager/src/runtime/managed.rs
aajtodd: Race with Drop of what? As long as we're in Dispatch the runtime should be active.

vcjana: i was thinking about a scenario where `shutdown_token.cancel()` fires on one thread while dispatch is mid-loop on another, but Drop takes `&mut self` so that can't happen while any &self borrow exists. The Arc keeps the runtime alive. my mistake.

Issue #92, opened 2025-01-21
UploadFilter should evaluate directories, not just files


graebm:
Currently, the UploadFilter only evaluates files, and the default filter does NOT exclude hidden directories ([source](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/ec067193d403098489c5f33536ec39def617e233/aws-s3-transfer-manager/src/types.rs#L179-L187)).

If a user wanted to exclude files in hidden directories, their filter would have to crawl up the filepath for ANY segment starting with ".". This filter would be run on every single file beneath the directory they didn't want to upload.

A more intuitive experience would be to run the UploadFilter on each directory before recursing into it (user just needs to check if name starts with "."). And if `false` is returned, don't recurse into the directory at all. This is also far more efficient.

This change will likely require transfer manager to write its own directory walking code (currently uses `walkdir`), which isn't hard, but just bringing it up

PR #142, opened 2026-05-04
vnext walking primitives, scheduler support for children, and transfer metrics

aajtodd:
# Walking primitives, scheduler child-cancel, and transfer metrics

## Series

8. **This PR**: Walking primitives, scheduler child-cancel, and transfer metrics
9. `upload_objects` state machine (upcoming)
10. `download_objects` state machine (upcoming)

## Summary

Introduces the building blocks for directory transfer operations (`upload_objects` / `download_objects`) on the new scheduler. This includes local filesystem and S3 walking primitives, scheduler support for parent-child transfer relationships, scaffolding for the `UploadObjectsTransfer` state machine, and a public transfer metrics/status API on handles.

None of this is wired into production paths yet. The existing `upload_objects` / `download_objects` implementations continue to use the legacy worker pool. The next two PRs will implement the actual state machines and swap out the legacy code.

## Why

The current `upload_objects` / `download_objects` use a bespoke worker pool with `async_channel` that operates outside the scheduler. This means they don't participate in fair-share scheduling, adaptive concurrency, or the cancellation model. Moving them onto the scheduler requires:

1. A walker abstraction that yields entries one at a time (async, paginated for S3)
2. The scheduler to understand parent-child relationships (cancel parent = cancel children)
3. Per-transfer metrics so the parent can observe child completion
4. Type scaffolding for the new state machine

## How to review

~4000 lines, but most of it is the two walker modules which are self-contained with extensive tests.

1. `src/io/walk/error.rs` (147 lines) - `WalkError` and `WalkErrorKind`. Per-entry vs fatal error distinction.

2. `src/io/walk/fs.rs` (1667 lines, 44 tests) - Local filesystem walker. Config/instance split via builder. Cycle detection in `read_dir`. Sort support for sync compatibility.

3. `src/io/walk/s3.rs` (1005 lines, 19 tests) - S3 listing walker. Same config/instance pattern. Pagination, common prefix recursion, resume primitives.

4. `src/scheduler/scheduler.rs` - Child-cancel cascade. Small addition (~30 lines of logic, 4 tests).

5. `src/transfer.rs` + `src/types.rs` - `MetricsState`, `TransferContext`, and the public `TransferMetrics`/`TransferStatus` types.

6. `src/operation/upload_objects/transfer.rs` - Scaffolding only (stub `Transfer` impl, types).

7. `integration-tests/src/metrics.rs` - 6 integration tests for the metrics API.

### Walking

Both walkers follow a config/instance split: a `Builder` produces an immutable config struct (`FsWalker` / `S3Walker`), and `.walk(context)` produces a stateful walk instance (`FsWalk` / `S3Walk`) that yields entries via `async fn next()`.

**FsWalker** does breadth-first traversal per directory level (depth-first across levels). Each `read_dir` call produces files + subdirectories for one directory. When `sort` is enabled, entries within each directory are sorted lexicographically (needed for sync's merge-join). Cycle detection tracks directory handles on the current descent path; a symlink whose target matches an ancestor is reported as a per-entry error and skipped.

Defaults match the SEP: `follow_symlinks=false`, `recursive=false`. The CLI layer can set `follow_symlinks(true)` and `sort(true)` to match Python CLI behavior.

**S3Walker** paginates through `ListObjectsV2` responses per the SEP (null delimiter for flat listing, handle pagination). When a delimiter is explicitly configured, common prefixes are pushed onto a stack and processed depth-first after the current prefix is exhausted. This matches the SEP specification and is consistent with Java SDK v2's implementation. Additional configuration beyond the SEP (`start_after`, `continuation_token`, `expected_bucket_owner`, `request_payer`) supports resume and cross-account use cases needed by the CLI.

### Scheduler child-cancel

`Scheduler::cancel_transfer` now cascades to depth-1 children (transfers whose `TransferId.parent == Some(cancelled_id)`). Parent is cancelled first, then each child. This is the mechanism that lets cancelling an `upload_objects` transfer cancel all its in-flight child uploads.

### Transfer metrics

`MetricsState` is a per-transfer atomic counter store (network_tx/rx, disk_read/write, total_bytes, timestamps). `TransferContext::record_io` and `StreamContext::record_io` both write to per-transfer metrics and forward to per-client telemetry (the `Telemetry` struct on `Handle`, used by the adaptive concurrency controller).

Public API: `handle.status()` returns `TransferStatus` (Active/Completed/Failed/Cancelled), `handle.metrics()` returns a `TransferMetrics` snapshot. `join()` returns metrics on the output type.

## Changes

**`src/io/walk/` (new)**
- `error.rs`: `WalkError`, `WalkErrorKind` with fatal/per-entry classification
- `fs.rs`: `FsWalker`, `FsWalkerBuilder`, `FsWalkContext`, `FsWalk`, `DirEntry`
- `s3.rs`: `S3Walker`, `S3WalkerBuilder`, `S3WalkContext`, `S3Walk`

**`src/scheduler/scheduler.rs`**
- `cancel_transfer` cascades to depth-1 children
- 4 new unit tests (cascade, single-parent regression, sibling isolation, nonexistent ID)

**`src/transfer.rs` (new)**
- `MetricsState`: per-transfer atomic counters
- `TransferContext`: holds `Arc<Handle>` + `Arc<MetricsState>`, provides `record_io`, `set_total_bytes`, `transfer_status`, `transfer_metrics`

**`src/types.rs`**
- `TransferStatus` enum (`#[non_exhaustive]`)
- `TransferMetrics` struct (`#[non_exhaustive]`, pub fields)

**`src/operation/upload/handle.rs`, `download/handle.rs`**
- `status()` and `metrics()` methods

**`src/operation/upload/output.rs`, `download/output.rs`**
- `TransferMetrics` field on output types (populated by `join()`)

**`src/operation/upload_objects/transfer.rs` (new)**
- Stub `Transfer` impl, `UploadObjectsWork`, `UploadObjectsState`, `ChildMeta`
- `pipeline_depth` on `UploadObjectsInput`
- `source_path` on `FailedUpload`

**`src/io/stream.rs`, `src/io/part_reader.rs`**
- `StreamContext` holds `Arc<Telemetry>` for per-client forwarding
- `PartReaderBuilder` accepts `.telemetry()`

**`src/client.rs`**
- `Handle.telemetry` is now `Arc<Telemetry>` (enables sharing with StreamContext)

**`integration-tests/src/metrics.rs` (new)**
- 6 tests: upload metrics, upload-from-file, download metrics, download-to-file, upload abort, download abort

## Future

- `upload_objects` state machine (next PR): consumes `FsWalk`, orchestrates child uploads via scheduler
- `download_objects` state machine: consumes `S3Walk`, orchestrates child downloads
- Wire-up PRs: replace legacy worker pool, un-ignore existing tests
- Walker extensions (deferred): `dir_filter` for subtree pruning, `normalize_unicode`, `same_file_system`

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/io/mod.rs
+++ b/aws-sdk-s3-transfer-manager/src/io/mod.rs
@@ -16,6 +16,8 @@ mod stream;
 pub mod error;
 pub(crate) mod fs;
 mod size_hint;
+/// Walker types for traversing filesystems and S3 buckets.
+pub mod walk;
 
 // re-exports
 pub use self::aggregated_bytes::AggregatedBytes;
--- a/aws-sdk-s3-transfer-manager/src/io/walk/error.rs
+++ b/aws-sdk-s3-transfer-manager/src/io/walk/error.rs
@@ -7,40 +7,52 @@ use std::path::{Path, PathBuf};
 
 /// Classifies a [`WalkError`].
 ///
-/// Errors are either fatal (the walk cannot continue and no further entries
-/// will be produced) or non-fatal (the affected entry is skipped but the
-/// walk continues). Use [`WalkError::is_fatal`] to distinguish them.
+/// Each kind has a deterministic fatality (see [`WalkErrorKind::is_fatal`]).
+/// Errors whose kind is fatal terminate the walk; non-fatal kinds report
+/// the affected entry and the walk continues.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum WalkErrorKind {
-    /// Generic I/O failure reading a directory or entry.
-    Io,
-    /// Permission denied on a directory or entry.
-    PermissionDenied,
-    /// Path expected to be a directory but isn't (e.g. source is a file).
+    /// Source root cannot be opened for reading (I/O or permission error
+    /// on the initial path). Fatal: the walk never gets off the ground.
+    SourceUnreadable,
+    /// Source root exists but is not a directory. Fatal.
     NotADirectory,
-    /// Symlink encountered with no valid target.
+    /// S3 service error from `ListObjectsV2` or related call. Fatal.
+    Service,
+    /// I/O error reading a subdirectory or entry during the walk.
+    /// Non-fatal: the affected entry is skipped and the walk continues.
+    Io,
+    /// Permission denied on a subdirectory or entry during the walk.
+    /// Non-fatal.
+    PermissionDenied,
+    /// Symlink encountered with no valid target. Non-fatal.
     BrokenSymlink,
     /// Symlink whose target is a directory already on the current descent
     /// path (a cycle). Non-cyclic duplicate symlinks (two different symlinks
     /// to the same target) are not reported as cycles and are traversed
-    /// normally.
+    /// normally. Non-fatal.
     SymlinkCycle,
-    /// S3 service error from `ListObjectsV2` or related call.
-    Service,
-    /// Other error that doesn't fit the categories above.
-    Other,
+}
+
+impl WalkErrorKind {
+    /// Whether an error of this kind terminates the walk.
+    pub fn is_fatal(&self) -> bool {
+        matches!(
+            self,
+            WalkErrorKind::SourceUnreadable | WalkErrorKind::NotADirectory | WalkErrorKind::Service
+        )
+    }
 }
 
 /// An error encountered during a directory walk.
 ///
-/// Wraps an optional path, a [`WalkErrorKind`] classifier, a fatal flag,
-/// and a source error. Check [`is_fatal`](Self::is_fatal) to determine
-/// whether the walk will continue producing entries after this error.
+/// Wraps an optional path, a [`WalkErrorKind`] classifier, and a source
+/// error. Fatality is determined by [`kind`](Self::kind); see
+/// [`is_fatal`](Self::is_fatal).
 #[derive(Debug)]
 pub struct WalkError {
     path: Option<PathBuf>,
     kind: WalkErrorKind,
-    fatal: bool,
     source: Box<dyn std::error::Error + Send + Sync>,
 }
 
@@ -48,9 +60,9 @@ impl WalkError {
     /// The path associated with this error, if any.
     ///
     /// For non-fatal errors this is typically the entry that failed (a file
-    /// or symlink). For fatal errors it may be the directory that could not
-    /// be read. `None` for errors not tied to a specific path (e.g. S3
-    /// service errors).
+    /// or symlink). For fatal errors it may be the source root that could
+    /// not be opened. `None` for errors not tied to a specific path (e.g.
+    /// S3 service errors).
     pub fn path(&self) -> Option<&Path> {
         self.path.as_deref()
     }
@@ -64,25 +76,25 @@ impl WalkError {
     ///
     /// When `true`, no further entries will be produced by the walk.
     /// When `false`, the walk continues and may produce more entries.
+    /// Equivalent to `self.kind().is_fatal()`.
     pub fn is_fatal(&self) -> bool {
-        self.fatal
+        self.kind.is_fatal()
     }
 
     pub(crate) fn new(
         path: Option<PathBuf>,
         kind: WalkErrorKind,
-        fatal: bool,
         source: Box<dyn std::error::Error + Send + Sync>,
     ) -> Self {
-        Self {
-            path,
-            kind,
-            fatal,
-            source,
-        }
+        Self { path, kind, source }
     }
 
-    /// Classify an `io::Error` into a [`WalkErrorKind`].
+    /// Classify an `io::Error` into a non-root [`WalkErrorKind`].
+    ///
+    /// For root-level I/O failures construct
+    /// [`WalkErrorKind::SourceUnreadable`] directly instead; this helper
+    /// is only appropriate for errors encountered on subdirectories or
+    /// entries reached after the walk has started.
     pub(crate) fn classify_io(err: &std::io::Error) -> WalkErrorKind {
         match err.kind() {
             std::io::ErrorKind::PermissionDenied => WalkErrorKind::PermissionDenied,
@@ -136,5 +148,16 @@ mod tests {
     fn test_classify_io_other_maps_to_io() {
         let err = std::io::Error::from(std::io::ErrorKind::ConnectionRefused);
         assert_eq!(WalkError::classify_io(&err), WalkErrorKind::Io);
+    }
+
+    #[test]
+    fn test_is_fatal_by_kind() {
+        assert!(WalkErrorKind::SourceUnreadable.is_fatal());
+        assert!(WalkErrorKind::NotADirectory.is_fatal());
+        assert!(WalkErrorKind::Service.is_fatal());
+        assert!(!WalkErrorKind::Io.is_fatal());
+        assert!(!WalkErrorKind::PermissionDenied.is_fatal());
+        assert!(!WalkErrorKind::BrokenSymlink.is_fatal());
+        assert!(!WalkErrorKind::SymlinkCycle.is_fatal());
     }
 }
--- a/aws-sdk-s3-transfer-manager/src/io/walk/error.rs
+++ b/aws-sdk-s3-transfer-manager/src/io/walk/error.rs
@@ -5,12 +5,11 @@
 
 use std::path::{Path, PathBuf};
 
-/// Classifies a [`WalkError`] for consumer-side branching.
+/// Classifies a [`WalkError`].
 ///
-/// Separates fatal failures (walk cannot continue) from per-entry
-/// failures (individual entry failed, walk continues). State machines
-/// use [`WalkError::is_fatal`] to decide whether to abort the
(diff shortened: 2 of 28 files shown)


aws-sdk-s3-transfer-manager/src/io/walk/error.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// When `true`, no further entries will be produced by the walk.
+    /// When `false`, the walk continues and may produce more entries.
+    pub fn is_fatal(&self) -> bool {
+        self.fatal
+    }
+
+    pub(crate) fn new(
+        path: Option<PathBuf>,
+        kind: WalkErrorKind,
+        fatal: bool,
```
landonxjames: I feel like the value of `fatal` should be derived from `WalkErrorKind` rather than set explicitly? Unless `WalkerErrorKind::Foo` can be both fatal and not fatal depending on circumstances not captured in the state of the struct itself?

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
```diff
@@ -0,0 +1,1704 @@
...
+impl FsWalk {
+    /// Return the next entry from the walk.
+    ///
+    /// Returns:
+    /// - `Some(Ok(entry))` for a file that passed the filter.
+    /// - `Some(Err(err))` for a fatal or non-fatal error. Fatal errors
+    ///   are followed by `None` on the next call; non-fatal errors are
+    ///   followed by more results as the walk continues.
+    /// - `None` when the walk is complete.
+    pub async fn next(&mut self) -> Option<Result<DirEntry, WalkError>> {
```
landonxjames, line 438: Does this need to be `async`? It doesn't seem to `await` anything

And if we make it non-async we `impl Iterator` for `FsWalk` so we get all the associated utilities instead of having pub next?

aws-sdk-s3-transfer-manager/src/io/walk/s3.rs
```diff
@@ -0,0 +1,1017 @@
...
+
+impl S3Walk {
+    /// Return the next object from the walk.
+    ///
+    /// Returns:
+    /// - `Some(Ok(object))` for an object that passed the filter.
+    /// - `Some(Err(err))` when a `ListObjectsV2` request fails. The walk
+    ///   terminates after an error; subsequent calls return `None`.
+    /// - `None` when the walk is complete.
+    pub async fn next(&mut self) -> Option<Result<Object, WalkError>> {
```
landonxjames, line 344: Similarly feel like `impl Iterator` is justified here

aws-sdk-s3-transfer-manager/src/operation/upload_objects/input.rs
```diff
@@ -35,6 +35,12 @@ pub struct UploadObjectsInput {
 
     /// The failure policy to use when any individual object upload fails.
     pub failure_policy: FailedTransferPolicy,
+
+    /// Maximum number of concurrent child upload transfers.
+    ///
+    /// Controls how many individual object uploads run simultaneously.
+    /// Defaults to 100.
+    pub pipeline_depth: usize,
```
landonxjames, line 43: Seems like this pattern was already in place, but feel like this should have a builder instead of all `pub` fields

aws-sdk-s3-transfer-manager/src/scheduler/scheduler.rs
```diff
@@ -142,35 +142,57 @@ impl Scheduler {
...
-            let purged = self.handle().runtime.remove_pending_for_transfer(id);
-            self.0.dispatched.fetch_sub(purged, Ordering::Relaxed);
-            desc.work_purged(purged);
-            desc.notify_idle();
-            tracing::debug!(target: telemetry::TARGET_SCHEDULING, id = %id, purged, "transfer cancelled");
-            true
-        } else {
-            false
+        let (target, children) = {
+            let mut transfers = self.0.transfers.write().unwrap();
```
landonxjames, line 157: Holding the write lock while scanning all of the transfers feels like it could cause a lot of contention in high concurrency use?

aws-sdk-s3-transfer-manager/src/operation/upload_objects/transfer.rs
```diff
@@ -0,0 +1,259 @@
...
+        UploadObjectsTransfer::ctx(self)
+    }
+
+    fn poll_work(&self) -> PollWork {
+        if !self.inner.ctx.is_active() {
+            let mut state = self.inner.state.lock().expect("lock poisoned");
+            *state = UploadObjectsState::Complete;
+            self.inner.ctx.signal_terminal();
+        }
+        PollWork::Done
```
landonxjames, line 171: Probably worth a `TODO` explaining that this still needs to be implemented since it always does `signal_terminal` right now

landonxjames (approved):
LGTM, few questions, but no blockers

aws-sdk-s3-transfer-manager/src/io/mod.rs
```diff
@@ -16,6 +16,8 @@ mod stream;
 pub mod error;
 pub(crate) mod fs;
 mod size_hint;
+/// Walker types for traversing filesystems and S3 buckets.
+pub mod walk;
```
landonxjames, line 20: Should `walk` be `pub`? I don't know that we won't to support its use outside of the TM and be responsible for its stability.

aws-sdk-s3-transfer-manager/src/io/walk/s3.rs
```diff
@@ -0,0 +1,1017 @@
...
+        }
+
+        let common_prefixes = output
+            .common_prefixes
+            .unwrap_or_default()
+            .into_iter()
+            .filter_map(|cp| cp.prefix)
+            .collect();
+
+        Ok(ListPageResult {
```
DmitriyMusatkin, line 480: note: there is a slight subtlety here with how regular s3 treats empty folder markers.  
ex. console uses empty objects with delim at the end to represent "folder" concept. 
those are technically not real objects and we will need to differentiate those special objects later in the code when creating correspond fs files.
express treats it differently and either no folder gets and object or all of them do

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
aajtodd: We'll validate this shape when we finish the state machines and fix tokio-MT interaction; likely stays async
  for forward-compat with spawn_blocking / io_uring, but let's revisit once real usage is in place.

aws-sdk-s3-transfer-manager/src/io/walk/s3.rs
aajtodd: Iterator isn't async

aws-sdk-s3-transfer-manager/src/operation/upload_objects/input.rs
aajtodd: Deferring, matches existing, we can cleanup whole crate if we're making a different choice

aws-sdk-s3-transfer-manager/src/scheduler/scheduler.rs
aajtodd: Defer until benchmarked as a problem. Cancellation isn't on the hot path usually.

aws-sdk-s3-transfer-manager/src/operation/upload_objects/transfer.rs
aajtodd: This is coming in next PR, just scaffolding for now

aws-sdk-s3-transfer-manager/src/io/mod.rs
aajtodd: Yes, we want it to be pub. We're committed to it's stability and behaviors anyway.

aws-sdk-s3-transfer-manager/src/io/walk/s3.rs
aajtodd: Noted, will look to add a test for this when we get to the state machines.

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
```diff
@@ -0,0 +1,1708 @@
...
+    /// The filter does not apply to directories. Directory traversal is
+    /// controlled by `max_depth` and `follow_symlinks`.
+    ///
+    /// **The filter does not prune subtrees.** A file-level filter that
+    /// rejects everything under a given subdirectory still causes the
+    /// walker to descend into and stat every file in that subtree. For
+    /// performance on trees with large excludable subtrees (e.g. `.git/`,
+    /// `node_modules/`), compose multiple walks rooted at smaller subtrees.
+    #[must_use]
+    pub fn filter(mut self, f: impl Fn(&DirEntry) -> bool + Send + Sync + 'static) -> Self {
```
ysaito1001, line 342: Related to https://github.com/awslabs/aws-s3-transfer-manager-rs/issues/92, this `filter` looks like it acts on individual `DirEntry`s, but now that we own walking dir logic, hopefully supporting directory filter isn't too hard?

aajtodd: Good callout will look at addressing in next PR

PR #145, opened 2026-05-14
Call shutdown on tokio::fs::File before drop in download_objects

AlJohri:
## Issue

`download_objects` reports completion before files are fully written to disk. Callers that stat the downloaded file immediately after `send().await?.join().await?` returns observe a short read by a small, varying amount; a retry usually succeeds because by then the writes have completed.

## Root cause

`src/operation/download_objects/worker.rs` opened a `tokio::fs::File`, ran `write_all` in a loop, and let `dest` drop without `flush().await` or `shutdown().await`:

```rust
let mut dest = fs::File::create(key_path).await?;
while let Some(chunk) = body.next().await {
    let chunk = chunk?;
    for segment in chunk.data.into_segments() {
        dest.write_all(segment.as_ref()).await?;
    }
}
Ok(())
```

Per the tokio docs ([fs module](https://docs.rs/tokio/latest/tokio/fs/index.html), [File buffering](https://docs.rs/tokio/latest/tokio/fs/struct.File.html#in-the-presence-of-buffering)):

> It is important to use flush when writing to a Tokio File. This is because calls to write will return before the write has finished, and flush will wait for the write to finish. (The write will happen even if you don't flush; it will just happen later.) This is different from std::fs::File, and is due to the fact that File uses spawn_blocking behind the scenes.

The bytes do eventually land — the `spawn_blocking` write keeps a clone of the inner `Arc<std::fs::File>`, so the fd stays open until the blocking task completes — but they have not landed by the time the worker's `Ok(())` returns. The result is that `download_objects` reports success *before* the file is fully written.

## Fix

Add `dest.shutdown().await?;` before the worker's final `Ok(())`. `shutdown` flushes the internal buffer and synchronously waits for the in-flight `spawn_blocking` write to complete, which is the contract the caller assumes.

The same one-line bug exists in `examples/cp.rs::write_body`; fixed there too so the example doesn't propagate the pattern.

## Regression test

To make the fix verifiable, the file-write portion of `download_single_obj` is extracted into a `write_body_to_open_file(body, dest)` helper. The new test `write_body_to_open_file_returns_only_after_bytes_are_durable` calls this helper directly and asserts the durability contract deterministically:

- Build a runtime with `max_blocking_threads(1)`.
- Pre-open the destination via `std::fs::File::create` + `tokio::fs::File::from_std` so the file-open path doesn't need the pool.
- Saturate the single blocking thread with a parked task.
- Spawn the helper. With `shutdown` present, the helper must wait for the queued `spawn_blocking` write to drain — `writer.is_finished()` stays false until we release the saturator. Without `shutdown`, the helper returns immediately while the pool is still saturated, and the assertion fires.

Verified locally: with `dest.shutdown().await?` present, the test passes. With the line removed, the test fails with:

```
write_body_to_open_file returned while the blocking pool is saturated — `dest.shutdown().await?` is missing or non-load-bearing
```

## Production impact

Hit repeatedly in a downstream consumer that performs a `HEAD`-vs-`stat` size check immediately after `transfer_manager.send().await?.join().await?` returns. The check would observe short files at multiple call sites within a single work session — short by ~1–4 KB each time, retries cleared each one. With the fix applied, the size check never trips.

## Test plan

- [x] `cargo test --package aws-sdk-s3-transfer-manager` — all existing tests pass plus the new regression test.
- [x] Removing `dest.shutdown().await?` from `write_body_to_open_file` causes the new regression test to fail deterministically.
- [ ] Maintainer review of helper-extraction shape (happy to inline or reshape if preferred).

### DIFF
--- a/aws-sdk-s3-transfer-manager/examples/cp.rs
+++ b/aws-sdk-s3-transfer-manager/examples/cp.rs
@@ -340,6 +340,10 @@ async fn write_body(body: &mut Body, mut dest: fs::File) -> Result<(), BoxError>
         }
         tracing::trace!("chunk had {segment_cnt} segments");
     }
+    // Flush tokio's internal write buffer before the file is dropped;
+    // without this the file may still be partially written when this
+    // function returns. See `tokio::fs::File` docs on buffering.
+    dest.shutdown().await?;
     Ok(())
 }
 
--- a/aws-sdk-s3-transfer-manager/src/operation/download_objects/worker.rs
+++ b/aws-sdk-s3-transfer-manager/src/operation/download_objects/worker.rs
@@ -219,10 +219,7 @@ async fn write_body_to_file(body: &mut Body, key_path: &Path) -> Result<(), erro
 /// be in flight, and callers that stat the file immediately can observe
 /// a short read. See the tokio docs:
 /// <https://docs.rs/tokio/latest/tokio/fs/struct.File.html#in-the-presence-of-buffering>.
-async fn write_body_to_open_file(
-    body: &mut Body,
-    mut dest: fs::File,
-) -> Result<(), error::Error> {
+async fn write_body_to_open_file(body: &mut Body, mut dest: fs::File) -> Result<(), error::Error> {
     while let Some(chunk) = body.next().await {
         let chunk = chunk?;
         for segment in chunk.data.into_segments() {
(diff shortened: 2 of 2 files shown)


aajtodd:
Thanks for the PR. This looks good, just needs to get through CI. Looks like fmt is blocking it currently.

AlJohri:
@aajtodd looks like CI is failing for a different reason now, maybe flaky?

```
Run aws-actions/configure-aws-credentials@v4
Error: Credentials could not be loaded, please check your action inputs: Could not load credentials from any providers
```

aajtodd:
Hey thanks again for the PR. We cut the 0.2.0 release a few weeks ago which was a significant re-architecture of the internals. Would encourage testing it out and if you still have any issues or feedback please don't hesitate to open an issue.

PR #146, opened 2026-05-18
vnext scheduler hierarchical CFS support

aajtodd:
# Hierarchical CFS scheduler for composite transfers

## Series
10. **This PR**: Hierarchical CFS scheduler
11. `upload_objects` state machine + file body (upcoming)
12. `download_objects` state machine (upcoming)

## Summary

Extends the scheduler from flat per-transfer fairness to a two-level hierarchy where transfers are grouped. A composite transfer (e.g. `upload_objects`) and all its children form a single group that competes as one entity in the root tree, while children compete fairly within the group's inner tree.

Existing flat transfers (upload, download) are unaffected. A top-level transfer with no children is a group of one; the two-level pop degenerates to the previous single-level behavior.

## Why

The flat scheduler treats every transfer equally. When `upload_objects` spawns 500 child uploads, those children individually compete with a concurrent single-file download. The download gets 1/501 of the scheduler's attention instead of 1/2. Hierarchical grouping fixes this: the `upload_objects` group and the download each get roughly 50% of dispatch share, regardless of how many children the group contains.

This also introduces the claim protocol and submission queue that prevent the re-entrancy deadlock discovered during upload_objects development.

## How to review

~3500 lines across 9 files. Start with the design doc update, then the ready set (the core data structure), then the scheduler (the dispatch loop), then supporting infrastructure.

1. **`docs/design/scheduler.md`** (+285/-13) — Design doc updates. New sections: Composite Transfers, Concurrency and Threading, Cost Model, Correctness Invariants. Read this first for the mental model.

2. **`src/scheduler/ready_set.rs`** (+1190/-181) — The two-level data structure. `GroupState` (atomic coordination extracted for loom testability), `GroupQueue` (per-group SkipMap), `ReadySet` (root SkipMap of groups). 15 unit tests + 7 loom tests.

3. **`src/scheduler/scheduler.rs`** (+1009/-46) — `generate_work` claim-based dispatch loop, `enqueue_transfer` with group resolution, re-entrancy guard, cancellation cascade. 32 unit tests.

4. **`src/scheduler/descriptor.rs`** (+256/-15) — `ClaimState` (prevents double-dispatch via CAS), `group_vruntime` Arc sharing, priority-scaled cost accounting. 2 loom tests.

5. **`src/runtime/sync/submission.rs`** (+239/-7) — `SubmissionQueue`: bounded MPSC between `generate_work` and the runtime. Back-pressure when full. 18 unit tests.

6. **`src/scheduler/transfer/mock.rs`** (+514) — Test infrastructure: `CountedWork`, `BlockingWork`, `ChildMockTransfer`, `DispatchCounter`.

7. **`src/transfer.rs`** (+283/-14) — `WakeFlag` (loom-verified edge-triggered wake primitive), `TransferContext::new_child` (parent linkage), parent-wake on `signal_terminal`, expanded `Transfer` trait docs. 1 loom test.

8. **`src/operation/download/body.rs`**, **`src/operation/download/transfer.rs`** — 1-line doc fixes each ("blocks" to "waits for").

### Two-level dispatch

The ready set is two nested SkipMaps:

```
root:  (group_vruntime, group_id) -> GroupQueue
group: (member_vruntime, member_id) -> TransferDescriptor
```

`pop()` takes the lowest-vruntime group from the root, then the lowest-vruntime member within that group. After popping, the group is re-inserted at its updated vruntime if it still has members. A CAS on `GroupState::in_root` prevents duplicate root entries when pop's re-insert races with a concurrent child insertion transitioning the group from empty to non-empty.

Group-entity accounting: when a member is popped, the group's `group_vruntime` advances by `vruntime_delta_for_priority(member.priority)`. Higher-priority groups advance more slowly and win more dispatch share at the root.

### Claim protocol

Before calling `poll_work`, the scheduler claims the descriptor via `ClaimState`. A claimed descriptor cannot be dispatched again until released. This enforces single-poll exclusivity: at most one thread is inside `poll_work` for a given transfer at a time.

On `PollWork::Pending` the claim is released and the descriptor leaves the ready set until woken. On `PollWork::Ready` the descriptor stays claimed through execution; the runtime re-inserts it after `execute` completes.

A `wake_requested` flag on the descriptor handles wakes that arrive during the release window (between claim release and potential re-insert). `generate_work` checks this flag after releasing and re-inserts if set.

### Wake primitive

Edge-triggered, mutex-serialized. The poller's condition check and `set_pending` call both happen under the transfer's state lock. The mutator follows `lock → mutate → unlock → try_wake`. Because both sides hold the same lock, they are serialized: either the mutator's change is visible to the poller (returns Ready), or the poller sets pending first and the mutator's `try_wake` observes it. Verified under loom.

### Re-entrancy guard

A composite transfer's `poll_work` may call `scheduler.enqueue_transfer` to spawn children. A thread-local `IN_GENERATE_WORK` flag prevents the inner `enqueue_transfer` from recursing into `generate_work`. The outer frame picks up newly-inserted transfers on its next loop iteration.

### Parent-child linkage

`TransferContext::new_child(handle, parent_id)` creates a context linked to a parent. When the child calls `signal_terminal`, it wakes the parent so the parent's state machine can reap it. `scheduler.cancel_transfer(parent_id)` cascades to all children in the parent's group.

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/scheduler/ready_set.rs
+++ b/aws-sdk-s3-transfer-manager/src/scheduler/ready_set.rs
@@ -349,20 +349,79 @@ impl ReadySet {
     pub(super) fn reinsert_under_claim(&self, descriptor: TransferDescriptor) {
         let tid = descriptor.id();
         let group_id = tid.parent.unwrap_or(tid.id);
-        let by_group = self.by_group.read().unwrap();
-        if let Some(group) = by_group.get(&group_id) {
-            let group = Arc::clone(group);
-            drop(by_group);
-            let prev_count = group.insert(descriptor);
-            if prev_count == 0 && group.state.try_enter_root() {
-                let gv = group.group_vruntime();
-                let key = GroupKey {
-                    group_vruntime: gv,
-                    group_id,
-                };
-                self.groups.insert(key, group);
-            }
+        // The Ready path: descriptor was popped from this group, so its
+        // group_vruntime is already current; preserve it (don't reset to
+        // root_min).
+        if self
+            .try_insert_into_existing_group(group_id, descriptor, false)
+            .is_err()
+        {
+            tracing::trace!(
+                target: crate::telemetry::TARGET_SCHEDULING,
+                id = %tid,
+                "reinsert_under_claim: parent group gone (cancel during Ready); descriptor already terminal"
+            );
         }
+    }
+
+    /// Helper for the inner ready_set lock-ordering protocol: consistency
+    /// between `by_group` and `groups`.
+    ///
+    /// Lock-ordering invariant: `by_group.read` is held across both
+    /// `group.insert(descriptor)` (into the inner queue) AND
+    /// `self.groups.insert(...)` (into the root tree). A concurrent
+    /// `remove_group` (which needs `by_group.write`) blocks until this
+    /// method returns. Without this ordering, `remove_group` could clear
+    /// `by_group` and `groups` between our drop-of-read and our write,
+    /// leaving a phantom entry in `groups` for a group_id no longer
+    /// present in `by_group`. The phantom misleads pop and stalls
+    /// scheduling for that descriptor.
+    ///
+    /// Verified by `loom_tests::child_insert_phantom_group`.
+    ///
+    /// On `Err(descriptor)`: group is gone (parent was cancelled
+    /// concurrently). Caller decides whether this is an error
+    /// (`insert_child` returns OrphanedChild) or benign (Ready-path
+    /// reinsert: descriptor will be cleaned up via terminal flag).
+    /// Importantly, `descriptor` is returned UNCONSUMED on this path,
+    /// so callers can run cleanup or fall back to a different code path.
+    ///
+    /// `reset_group_vruntime`: when true, the group's vruntime is reset
+    /// to the current root floor (used by `insert_top_level` and
+    /// `insert_child`, which are wake/spawn paths). When false, the
+    /// existing group vruntime is preserved (used by
+    /// `reinsert_under_claim`, where the descriptor was popped from
+    /// this group and the vruntime is current).
+    fn try_insert_into_existing_group(
+        &self,
+        group_id: u64,
+        descriptor: TransferDescriptor,
+        reset_group_vruntime: bool,
+    ) -> Result<(), TransferDescriptor> {
+        let by_group = self.by_group.read().unwrap();
+        let group = match by_group.get(&group_id) {
+            Some(g) => Arc::clone(g),
+            None => return Err(descriptor),
+        };
+        // From here, by_group.read is held until end-of-scope. Concurrent
+        // remove_group blocks waiting for by_group.write.
+        let prev_count = group.insert(descriptor);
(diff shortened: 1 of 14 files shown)


aws-sdk-s3-transfer-manager/src/scheduler/ready_set.rs
```diff
@@ -82,287 +283,914 @@ impl ReadySet {
...
+        let by_group = self.by_group.read().unwrap();
+        let group = match by_group.get(&parent_id) {
+            Some(g) => Arc::clone(g),
+            None => {
+                // Release the claim we took since we're rejecting this insert
+                descriptor.release_claim();
+                return Err(OrphanedChild);
+            }
+        };
+        drop(by_group);
```
landonxjames: Should this `drop` happen after the `group.insert` on the line below? If a `remove_group` happens concurrently between the `drop` and the `insert` what is the expected behavior?

aws-sdk-s3-transfer-manager/src/scheduler/scheduler.rs
```diff
@@ -314,23 +462,56 @@ impl Scheduler {
...
+            }));
+
+            match poll_result {
+                Ok(PollWork::Ready(item)) => {
                     generated += 1;
                     desc.work_generated();
-                    self.0.ready_set.insert(desc.clone());
+                    // Re-insert under the still-held claim - bypasses the
+                    // CAS gate that `insert` would otherwise apply.
+                    claim.hold();
```
landonxjames, line 614: Avoiding the CAS makes sense but also feels like it violates the invariant mentioned in the module level docs:
```
//!   CAS-gated on this claim, and the claim is held through `pop` and `poll_work`
//!   until the scheduler has finished handling the outcome.
```
since the claim is actually held across multiple `poll_work`s even though it is ready

landonxjames (approved):
Looks good, couple of questions, no blockers

ysaito1001 (approved):

aws-sdk-s3-transfer-manager/src/scheduler/scheduler.rs
```diff
@@ -274,7 +483,11 @@ impl Scheduler {
 
         let is_idle = desc.work_finished();
         if is_idle {
-            self.0.transfers.write().unwrap().remove(&desc.id());
+            let tid = desc.id();
+            self.0.transfers.write().unwrap().remove(&tid);
+            if tid.parent.is_none() {
+                self.0.ready_set.remove_group(tid.id);
+            }
```
vcjana: Should this use `remove_transfer_atomic` instead? The separate `remove` + `remove_group `
here doesn't hold `transfers.write` across both, unlike `cancel_transfer`

vcjana (approved):

aws-sdk-s3-transfer-manager/src/scheduler/scheduler.rs
aajtodd: Yes it should!

aws-sdk-s3-transfer-manager/src/scheduler/scheduler.rs
aajtodd: Updated docs

PR #147, opened 2026-05-26
vnext upload_objects state machine on hierarchical CFS

aajtodd:
# `upload_objects` state machine on hierarchical CFS

## Series
10. ~~Hierarchical CFS scheduler~~ (#146)
11. **This PR**: `upload_objects` state machine + file body
12. `download_objects` state machine (upcoming)

## Summary

Replaces the `upload_objects` worker pool (JoinSet + async_channel) with a state machine running as a composite transfer on the hierarchical CFS scheduler. Each directory entry becomes a child `Upload` transfer competing within the parent's scheduler group. The parent drives walker enumeration, child spawning, and result aggregation through the standard `poll_work` / `execute` contract.

Introduces `DirectFileBody` / `OffloadedFileBody` for streaming file upload via pread, and extends the filesystem walker with subtree splitting for parallel enumeration.

Deletes the legacy `worker.rs` (736 lines).

## Motivation

The legacy worker pool was a separate concurrency fabric outside the scheduler. It could not participate in hierarchical fair-share, could not receive backpressure from the adaptive controller, and required bespoke shutdown logic for cancellation. 

This PR routes `upload_objects` through the scheduler and makes the counter protocols structural via RAII tokens.

## `poll_work` structure

```
poll_work(self) -> PollWork
│
├─ Phase 1 [state.lock HELD]
│  ├─ drain_terminal_children  →  return JoinChildren(ReapingBatch)
│  ├─ !active                  →  signal_terminal, return Done
│  └─ claim_spawn_batch        →  (Vec<ClaimedEntry>, Reservation)
│
├─ Phase 2 [state.lock RELEASED]
│  └─ for each claimed entry:
│       Upload::orchestrate_child(handle, input, parent_id)
│       // calls scheduler.enqueue_transfer internally
│
└─ Phase 3 [state.lock RE-ACQUIRED]
   ├─ merge_spawned(outcomes, reservation.consume())
   ├─ claim_subtrees
   ├─ check_terminal
   └─ dispatch_walk  →  return AdvanceWalker(WalkSlot)
```

The state lock is released for phase 2 because `orchestrate_child` calls `scheduler.enqueue_transfer`, which acquires scheduler-internal locks. Holding the state lock across that call would block any completion-path thread that needs to lock state (e.g. a child completing → `signal_terminal` → `scheduler.wake(parent)` → `generate_work` → pops parent → parent's next `poll_work` blocks on state lock still held by the spawning thread). The `Reservation` token ensures the counter is recovered if phase 2 panics.

## Subtree splitting

The walker supports parallel enumeration by splitting pending subdirectories into independent walkers:

```
Filesystem:
  /data/
  ├── file1.txt
  ├── logs/
  │   ├── app.log
  │   └── err.log
  ├── images/
  │   └── photo.jpg
  └── reports/
      └── q1.csv

Step 1: Initial walker enumerates /data/, yields file1.txt,
         discovers logs/, images/, reports/ as pending subdirs.

  state.walks = { 0: FsWalk(/data/) }
                      ↑ has pending_dirs: [logs/, images/, reports/]

Step 2: claim_subtrees splits pending subdirs into new walkers:

  state.walks = { 0: FsWalk(/data/),      ← continues yielding /data/ files
                  1: FsWalk(/data/logs/),   ← independent walker
                  2: FsWalk(/data/images/),
                  3: FsWalk(/data/reports/) }

Step 3: dispatch_walk picks one walker per poll_work pass.
         Each executes as a separate AdvanceWalker work item
         on a different managed thread.

Bounded by MAX_PARALLEL_WALKS (16). Deep trees with many
subdirectories are enumerated in parallel without unbounded
fan-out.
```

## Counter protocol tokens

Three counters track work that has crossed the state lock boundary but not yet returned:

| Counter | Token | Increment | Decrement |
|---|---|---|---|
| `children_reserved` | `Reservation` | `claim_spawn_batch` | `merge_spawned` |
| `reaping_in_flight` | `ReapingBatch` | `drain_terminal_children` | `execute_join_children` |
| `in_flight_walks` | `WalkSlot` | `dispatch_walk` | `execute_advance_walker` |

Each token's `consume()` decrements under the state lock. `Drop` recovers via `saturating_sub` and emits a warning. `check_terminal` waits for all three counters to reach zero before signalling completion.

The consuming method's signature requires the token, making the protocol obligation visible at the type level.

## File body

`DirectFileBody` uses pread (positional read, no seek state, no file-level mutex) on managed threads where blocking I/O is the execution model. `OffloadedFileBody` wraps the same logic in `spawn_blocking` for shared runtimes. The runtime's `direct_io()` flag selects the variant at construction time.

Both implement `http_body::Body` and stream file content in 1 MiB chunks directly into the SDK request pipeline.

## How to review

~6700 lines across 28 files. Recommended reading order:

### Core state machine

**`src/operation/upload_objects/transfer.rs`** (+2086) — The heart of the PR. Contains `poll_work` (three-phase structure above), `execute_advance_walker` (walker advancement with entry processing), `execute_join_children` (batched child handle consumption via `join_all`), and the three RAII token types (`Reservation`, `ReapingBatch`, `WalkSlot`). Also contains `claim_spawn_batch` (key derivation, `UploadInput` construction under the lock), `merge_spawned` (orchestration result processing), `claim_subtrees` (parallel walker fan-out), and `check_terminal` (completion gate on all three counters). 21 unit tests exercise token Drop semantics and state machine paths directly.

### Walker extension

**`src/io/walk/fs.rs`** (+819) — `try_claim_subtree` on `FsWalk`: atomically removes a pending subdirectory and returns a new independent `FsWalk` rooted at that subdirectory. The parent walker continues yielding files from its current directory. Cycle detection state is shared across all walkers from the same root (they share the `Arc<HandleSet>`).

### File body

**`src/operation/upload/file_body.rs`** (+469) — `DirectFileBody` and `OffloadedFileBody`. Both hold a shared `Arc<File>` (one open fd per transfer, not per part) and a byte range. `poll_frame` reads the next chunk via pread at the current offset. No seek, no mutex — multiple parts of the same file can be read concurrently from different body instances.

### Upload path changes

**`src/operation/upload.rs`** (+33) — `Upload::orchestrate_child`: constructs a child `UploadTransfer` linked to a parent via `TransferContext::new_child`, enqueues it on the scheduler. This is the bridge between the parent state machine and the singular upload path.

**`src/operation/upload/transfer.rs`** (+179) — PutObject path now uses `DirectFileBody`/`OffloadedFileBody` instead of reading the entire file into memory. A per-operation `read_timeout` (30 s) caps time-to-first-response-byte; on timeout the SDK retries with a rebuilt body from the source file.

**`src/operation/upload/handle.rs`** (+65) — `UploadHandle::join()` no longer panics on `Cancelled` status.

### Public API reshape

**`src/operation/upload_objects/{input,output,handle,builders}.rs`** — `UploadObjectsInput` takes `walker: Option<FsWalker>` replacing `recursive`/`follow_symlinks`/`filter` (users who need custom walk behavior build a `FsWalker` directly). `UploadObjectsOutput` gains `metrics: TransferMetrics`. `UploadObjectsHandle` exposes `status()`, `metrics()`, `scheduling()`.

**`src/types.rs`** (+114) — `FailedUpload` gains `source_path: Option<PathBuf>` for diagnostics.

### Bug fixes

**`src/operation/download/handle.rs`** (+106) — Three bugs fixed:
- `DownloadHandleInner::join()` panicked on `Cancelled` — now returns `Err(OperationCancelled)`
- `ManagedDownloadHandle::join()` renamed temp→dest on cancel (silent data corruption) — now returns Err and removes the temp file
- `ManagedDownloadHandle` docs falsely claimed fsync — corrected

### Infrastructure

**`src/runtime/managed.rs`** (+55) — Managed thread count exposed for `max_concurrent_uploads` default derivation.

**`.github/workflows/ci.yml`** (+48) — Dedicated `dial9-build` job: builds the `cp` example with `--all-features` and `--cfg tokio_unstable`, scoped to that job only.

### Tests

**`tests/upload_objects_test.rs`** (+1868) — 27 integration tests + 2 ignored stress tests. Notable patterns: counter-based mocks (`succeeds_then_fails`) for deterministic abort timing, atomic max-tracking for concurrency assertions, watch-channel synchronization for drop-timing tests.

**`integration-tests/src/upload_objects.rs`** (+423) — E2E tests against mock S3 server.

### Deletions

**`src/operation/upload_objects/worker.rs`** (-736) — Legacy worker pool removed.


aws-sdk-s3-transfer-manager/src/operation/upload_objects/input.rs
```diff
@@ -3,12 +3,27 @@
...
 use std::path::{Path, PathBuf};
 
-/// Input type for uploading multiple objects
+/// Default per-request memory cap on concurrently-materialized child upload transfers.
+///
+/// Acts as a backstop on memory growth; the scheduler's hierarchical
+/// CFS handles fairness and rate. Users can override via
+/// [`UploadObjectsInputBuilder::max_concurrent_uploads`] or the
+/// corresponding fluent builder method.
+const DEFAULT_MAX_CONCURRENT_UPLOADS: usize = 10000;
```
landonxjames: This seems high based on past discussions. Maybe something more like 512/1024 since it seems like users could still set a higher concurrency limit if they wanted to?

aws-sdk-s3-transfer-manager/src/operation/upload_objects/transfer.rs
```diff
@@ -5,154 +5,1119 @@
...
+        abort_result
+    }
+
+    fn claim_subtrees(&self, state: &mut State) {
+        loop {
+            if state.walks.len() + state.in_flight_walks >= MAX_PARALLEL_WALKS {
+                break;
+            }
+            let mut claimed_any = false;
+            let walk_ids: Vec<u64> = state.walks.keys().copied().collect();
```
landonxjames, line 750: The keys are going to be randomly ordered here, which won't be incorrect, but it might be easier to test/reason about if the ordering was deterministic?

aws-sdk-s3-transfer-manager/src/operation/upload_objects/handle.rs
```diff
@@ -3,100 +3,137 @@
...
+    /// This will:
+    /// 1. Cancel the transfer in the scheduler
+    /// 2. Wait for any in-flight child uploads to complete or be cancelled
+    ///
+    /// When this method returns, all work for this transfer has been stopped.
+    /// No further child uploads will be initiated.
+    pub async fn abort(self) {
+        let ctx = self.transfer.ctx();
+        ctx.handle.scheduler.cancel_transfer(ctx.id);
+        ctx.handle.scheduler.wait_for_idle(ctx.id).await;
```
landonxjames: Is `wait_for_idle` a no-op here? I think `cancel_transfer` removes the descriptor from `scheduler.0.transfers`, then `wait_for_idle` looks it up there, finds nothing, and returns immediately

landonxjames (approved):
LGTM, few questions

aws-sdk-s3-transfer-manager/src/operation/upload_objects/input.rs
aajtodd: This is the default max concurrent child transfers and you only hit it if there is enough capacity. The`ConcurrencyController` still caps in-flight work so we don't want it too low but could drop down to 1/2/4k perhaps. Needs benchmarked across different instance sizes.

aws-sdk-s3-transfer-manager/src/operation/upload_objects/handle.rs
aajtodd: hmm indeed it does nothing now, let me look what we want to do here.

aws-sdk-s3-transfer-manager/src/operation/upload_objects/transfer.rs
aajtodd: Yeah can update to a BTreeMap for deterministic ordering.

aws-sdk-s3-transfer-manager/src/operation/upload_objects/input.rs
aajtodd: Lowered to 4096, can adjust further with benchmarks.

aws-sdk-s3-transfer-manager/examples/cp.rs
```diff
@@ -73,6 +73,14 @@ pub struct Args {
     /// Path to manifest JSON for concurrent transfers
     #[arg(long)]
     manifest: Option<String>,
+
+    /// Directory to write dial9 runtime traces (enables tracing when set)
+    #[arg(long)]
+    trace_dir: Option<String>,
+
+    /// Enable CPU profiling in dial9 traces (Linux only, requires --trace-dir)
+    #[arg(long, default_value_t = false, action = clap::ArgAction::SetTrue)]
```
ysaito1001, line 82: nit: can we include that condition in
```
requires = "trace-dir"
```
for clap?

vcjana (approved):

aws-sdk-s3-transfer-manager/src/operation/upload_objects/transfer.rs
```diff
@@ -5,154 +5,1119 @@
...
+            .field("consumed", &self.consumed)
+            .field("walk_present", &self.walk.is_some())
+            .finish_non_exhaustive()
+    }
+}
+
+impl WalkSlot {
+    /// Take the walker out for advancement. Must be called exactly
+    /// once before `consume`.
+    fn take_walk(&mut self) -> FsWalk {
```
ysaito1001, line 262: This is an aptly-named method 👍

ysaito1001 (approved):
Solid work, the three-phase lock design with RAII counter tokens is well thought out.

aws-sdk-s3-transfer-manager/examples/cp.rs
aajtodd: Good idea, I'll get it on the next PR for download objects.

PR #148, opened 2026-06-02
`download_objects` state machine on hierarchical CFS

aajtodd:
# `download_objects` state machine on hierarchical CFS

## Series

10. ~~Hierarchical CFS scheduler~~ (#146)
11. ~~`upload_objects` state machine + file body~~ (#147)
12. **This PR**: `download_objects` state machine

## Summary

Replaces the `download_objects` worker pool (JoinSet + async_channel) with a state machine running as a composite transfer on the hierarchical CFS scheduler, mirroring `upload_objects` (#147). The parent walks S3 via `S3Walk`; each listed object becomes a child `Download` transfer competing within the parent's scheduler group. The parent drives listing, child spawning, and result aggregation through the standard `poll_work` / `execute` contract.

Also reworks the spawn/reap loop in **both** `download_objects` and `upload_objects` to keep the in-flight pipeline full under bursty completion, fixes a `try_wake` terminal race in both directory state machines, and standardizes the transfer-id log field across the crate.

Deletes the legacy `worker.rs` (-585) and `list_objects.rs` (-363).

## `poll_work` structure

```
poll_work(self) -> PollWork
│
├─ 1. spawn  [lock; released around orchestration]
│     claim_spawn_batch → orchestrate child Downloads → merge_spawned
│     (a side effect, not a returned item: composes with the slot below)
│
├─ 2. walk   [lock]
│     dispatch_walk → return AdvanceWalker(S3Walk)
│     (preferred over reap for the single returned slot)
│
└─ 3. reap   [lock]
      drain_terminal_children → return JoinChildren(batch)
```

`poll_work` returns at most one work item. Spawning is a side effect that composes with whatever is returned, so the pipeline tops up every cycle. Walk is preferred over reap for the returned slot: a steady completion stream would otherwise reap forever and never list the next page. The state lock is released around child orchestration because `Download::orchestrate_with_sink` calls `scheduler.enqueue_transfer`, which takes scheduler-internal locks; holding the state lock across that would block any completion-path thread that needs to lock state.

## Keep the pipeline full (`download_objects` + `upload_objects`)

The concurrency cap conflated two budgets that need to be separate:

```
                 registered children
        ┌──────────────────────────────────┐
        │  active (transferring)  │ terminal│
        │  network + disk work    │ unreaped│
        └─────────────────────────┴─────────┘
         └──── in-flight budget ───┘
              gates SPAWN              both hold buffers
                                       until reaped
         └──────────── memory budget ──────────┘
                       gates LISTING
```

- **in-flight budget** — children actively transferring (network + disk concurrency). Gates spawning.
- **memory budget** — all registered children, including terminal-but-unreaped ones (buffers free at reap, not completion). Gates listing.

Counting terminal-unreaped children against spawn capacity ties the spawn rate to reap latency. The fix, applied to both modules:

1. Gate spawn on **active** (non-terminal) children + reserved, so a completion burst frees spawn capacity immediately rather than after the reap.
2. Asymmetric bounded work generation. `SPAWN_BATCH_SIZE = 32` (hoisted to shared `operation.rs`): each spawn registers a schedulable CFS entity, so the batch must stay small or one parent floods the ready set. `REAP_BATCH_SIZE = 256` (per module): a reap injects no schedulable entity (it joins already-terminal handles), so it can batch large; the bound only caps worker-thread hold time on the serial join. `drain_terminal_children` was previously unbounded — one burst could monopolize a thread.
3. Reorder `poll_work` to spawn → walk → reap (above).

`upload_objects` additionally: `dispatch_walk` now returns `Option<PollWork>`, and `debug_assert_capacity` asserts the in-flight budget (`active + reserved <= max`) rather than total registered.

## `try_wake` terminal race

`execute_advance_walker` / `execute_join_children` signaled terminal indirectly via `try_wake()` → re-poll → `check_terminal`. That path is unreliable from an execute callback: the wake pending flag is only armed when `poll_work` returns `Pending`, so when the last in-flight work drains inside an execute callback no flag is armed and the parent's `join()` hangs. Concurrent `JoinChildren` executions race the flag and lose the wake.

Fix: an execute callback that drains the last in-flight work owns the terminal transition — it calls `check_terminal` directly rather than deferring to a re-poll. This matches how singular upload/download signal terminal from `execute`. Applied to both directory modules; un-ignored `test_failed_get_object_should_cancel_the_operation`.

## How to review

~2200 net lines across 31 files. Recommended reading order:

### Core state machine

**`src/operation/download_objects/transfer.rs`** (+1401) — The heart of the PR. `poll_work` (three steps above), `execute_advance_walker` (advance `S3Walk`, push entries, fatal vs per-entry error handling per failure policy), `execute_join_children` (batched child `join()`, metrics rollup, Abort-policy handling), `claim_spawn_batch` (in-flight gate, key derivation, child build under the lock), `dispatch_walk` (memory-budget gate on listing), `drain_terminal_children` (bounded reap), and `check_terminal` (completion gate on children + reserved + reaping). Key derivation includes the path-traversal guard and prefix stripping. 18 unit tests cover the state-machine paths, key derivation, path traversal, and cancellation.

### Shared spawn constant

**`src/operation.rs`** (+67) — `SPAWN_BATCH_SIZE` hoisted here as `pub(crate)` and shared by both directory modules; doc states why the spawn batch is bounded (schedulable-entity injection).

### upload_objects symmetric change

**`src/operation/upload_objects/transfer.rs`** (313 changed) — Same keep-pipeline-full rework and `try_wake` fix. Spawn gates on a new `active_children()` helper; `dispatch_walk` converted to `Option<PollWork>` (its four "why parked" debug traces preserved); local `REAP_BATCH_SIZE`; `debug_assert_capacity` redefined to the in-flight budget. The `poll_work` phase structure is unchanged from #147 — only the gate, the reap bound, and step order moved.

### Public API reshape (matches upload_objects)

**`src/operation/download_objects/{input,output,handle,builders}.rs`** — `DownloadObjectsInput` takes `walker: Option<S3Walker>` replacing `recursive` / `follow_symlinks` / `filter` (users needing custom walk behavior build an `S3Walker`; `None` selects a default that filters 0-byte folder markers). `DownloadObjectsOutput` replaces `total_bytes_transferred` with `metrics: TransferMetrics` (matches singular `DownloadOutput`). `DownloadObjectsHandle` exposes `status()`, `metrics()`, `scheduling()` and a non-panicking `join()`.

**`src/types.rs`** (-44) — `DownloadFilter` removed from the public API.

### Download path bridge

**`src/operation/download.rs`** (+16) — child download initiation linked to the parent via `TransferContext::new_child`.

### Telemetry

**`src/scheduler/scheduler.rs`, `src/transfer.rs`, `src/operation/**`** — transfer-id log field standardized to `tid = %<id>` (Display renders `id-parent` for children) across the crate; was four different field names. download_objects gained trace events at every state transition.

### Tests

**`tests/download_objects_test.rs`** (401 changed) — integration tests migrated to the state machine, wrapped in `tokio::time::timeout` so hangs fail fast.

**`tests/e2e_transfer_test.rs`** (+12) — `test_objects_transfer` stays `#[ignore]` (needs real S3); un-ignored in the e2e expansion pass.

### Deletions

**`src/operation/download_objects/worker.rs`** (-585), **`list_objects.rs`** (-363) — legacy worker pool and listing removed.


aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
```diff
@@ -0,0 +1,1439 @@
...
+                && !state.walk_in_flight
+            {
+                tracing::debug!(
+                    target: crate::telemetry::TARGET_TRANSFER,
+                    tid = %self.inner.ctx.id,
+                    successful = state.successful_downloads,
+                    failed = state.failed.len(),
+                    "download_objects terminal (cancelled/failed), signaling",
+                );
+                self.inner.ctx.signal_terminal();
```
landonxjames, line 595: Should this also call `.set_completed()`? The equivalent method for [upload_objects](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/bb59ff873dde0efca3ef5a9b1f80e879a76f029c/aws-sdk-s3-transfer-manager/src/operation/upload_objects/transfer.rs#L793-L794) calls it.

aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
```diff
@@ -0,0 +1,1439 @@
...
+            && state.reaping_in_flight == 0
+        {
+            tracing::debug!(
+                target: crate::telemetry::TARGET_TRANSFER,
+                tid = %self.inner.ctx.id,
+                successful = state.successful_downloads,
+                failed = state.failed.len(),
+                "download_objects terminal (walk exhausted), signaling",
+            );
+            self.inner.ctx.signal_terminal();
```
landonxjames, line 620: Same comment here

aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
```diff
@@ -0,0 +1,1439 @@
...
+        drop(state);
+        // Wake: new entries in pending_entries unblock poll_work from Pending.
+        self.inner.ctx.try_wake();
+        WorkOutcome::Success { data: None }
+    }
+
+    async fn execute_join_children(&self, batch: Vec<ChildTransfer>) -> WorkOutcome {
+        for child in batch {
+            let key = child.key;
+            match child.handle.join().await {
```
landonxjames, line 794: I think we need to record metrics before doing the join like in [upload_objects](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/bb59ff873dde0efca3ef5a9b1f80e879a76f029c/aws-sdk-s3-transfer-manager/src/operation/upload_objects/transfer.rs#L1026-L1031)?

aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
```diff
@@ -0,0 +1,1439 @@
...
+        )?;
+
+        // Create parent directories, deduped via created_dirs so we don't
+        // re-stat the whole ancestor chain for every object sharing a prefix
+        // (create_dir_all stats each ancestor; for N small files in a few
+        // dirs that's the dominant per-child syscall cost).
+        if let Some(parent) = dest_path.parent() {
+            let known = self.inner.created_dirs.lock().contains(parent);
+            if !known {
+                std::fs::create_dir_all(parent).map_err(|e| {
```
landonxjames, line 492: Pretty sure this ends up being called in `poll_work`, the documentation of which says it will [never contain blocking IO](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/bb59ff873dde0efca3ef5a9b1f80e879a76f029c/aws-sdk-s3-transfer-manager/src/scheduler/scheduler.rs#L53-L56). Could we defer this to `execute`?

aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
```diff
@@ -0,0 +1,1439 @@
...
+    pending_entries: std::collections::VecDeque<Object>,
+
+    /// Active child downloads keyed by their TransferId.
+    children: HashMap<TransferId, ChildTransfer>,
+
+    /// Entries claimed from pending_entries but not yet inserted into children.
+    children_reserved: usize,
+
+    /// JoinChildren batches currently executing.
+    reaping_in_flight: usize,
```
landonxjames, line 219: These are treated as raw values here and manually updated, but the similar values in `upload_objects` are wrapped in custom types like [ReapingBatch](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/bb59ff873dde0efca3ef5a9b1f80e879a76f029c/aws-sdk-s3-transfer-manager/src/operation/upload_objects/transfer.rs#L155-L164) with custom `Drop` impls to avoid having to manually manage the counter. 

That seems safer and like a pattern we should probably repeat here to keep things consistent.

landonxjames (approved):
Couple of small notes, nothing blocking

aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
```diff
@@ -0,0 +1,1439 @@
...
+/// serial join.
+const REAP_BATCH_SIZE: usize = 256;
+
+/// Maximum entries to drain from the walker per AdvanceWalker work item.
+/// Prevents a single work item from holding the executor across multiple
+/// ListObjectsV2 pages. Matches S3's default MaxKeys.
+const MAX_ENTRIES_PER_WALK: usize = 1000;
+
+/// Default S3 key delimiter.
+const DEFAULT_DELIMITER: &str = "/";
```
ysaito1001, line 43: nit: can we reuse `DEFAULT_DELIMITER` in `operation.rs`?

ysaito1001 (approved):
🚢

aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
aajtodd: This one differs, it's already under !active check so it's already terminal/transitioned.

aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
aajtodd: Fixed this one.

aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
aajtodd: Yup, fixed + test.

aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
aajtodd: It does I'm going to leave it for now because it's a little blurry whether this is bad in practice (at least on managed runtime where we already own the thread)

PR #149, opened 2026-06-08
vnext checksum and integrity checks

aajtodd:
## Summary

Make transfer manager downloads validate object checksums, report the result on the download output, and add the test infrastructure to keep it honest: a mock and real-S3 integrity contract suite, mock fault injection, and mock checksum fidelity fixes. Checksums are the largest remaining gap before merging `s3-tm-vnext` into `main`. Branches off `s3-tm-vnext` (`15ec23b`, after #148).

## Why

Download validation did not work. A download is sliced into ranged GETs at a fixed part size (Auto = 5 MiB) that ignores the object's stored layout. S3 returns a per-part checksum only when a ranged GET matches a stored part boundary, so an object uploaded multipart (Auto upload part = 8 MiB) and downloaded with defaults produced ranges that never aligned, and nothing was validated. The output also carried no checksum information, so a caller could not tell whether anything was checked.

Fixing this surfaced two latent bugs. First, `execute_get_range` hand-built each chunk GET with only bucket/key/range/if_match, dropping the other 17 fields the `From<DownloadInput>` conversion sets. Besides `checksum_mode` (the reason nothing validated), this dropped the SSE-C customer key (encrypted multipart range GETs would 400), `version_id` (a range chunk could read a different version than discovery pinned), `request_payer`, `expected_bucket_owner`, and the response-header overrides. Second, the error paths decremented the in-flight range count before transitioning to terminal, so a concurrent `poll_work` woken in between could observe `ranges_in_flight == 0` and `complete()` the transfer over a `ChecksumMismatch` on the last chunk, reporting success.

## How to review

Commit by commit. Each is self-contained and builds on the last.

1. [ensure inputs are propagated](https://github.com/awslabs/aws-s3-transfer-manager-rs/commit/1fa8674422db42596f331f2467a69e1feb8c01be) — range-GET field forwarding. `execute_get_range` builds from `From<DownloadInput>` then overrides range + etag, so every chunk carries the same fields as discovery. Smallest commit, independent of checksums; start here.
2. [improve checksum fidelity](https://github.com/awslabs/aws-s3-transfer-manager-rs/commit/4adc45dd472c0d23cdb09fc052d23e1d17585457) — mock-server composite and ranged-GET checksum behavior. Composite was hashing the base64 text of the part checksums (now decodes to raw bytes) and omitting the `-N` suffix; GET was range-blind (now returns the part checksum on an aligned range, none on an unaligned one, the object value on a whole-object GET). Mock only.
3. [bootstrap fault injection](https://github.com/awslabs/aws-s3-transfer-manager-rs/commit/284bb0d7e1617ff875c9547d9beba2d09844ef57) — `faults.rs`: a key-scoped ordered fault queue, `skip` warm-up orthogonal to `Occurrence` (Once/NTimes/Always), no RNG, every fire logged. Two checksum variants (`WrongStoredChecksum`, `CorruptBody`) applied in `get_object`. A general test primitive, not checksum-specific.
4. [define integrity checks](https://github.com/awslabs/aws-s3-transfer-manager-rs/commit/4c52f565f35a4924ec19ccc6ee369fab402e0167) — `IntegrityChecks` / `ChecksumValidation` / `NotValidatedReason` in `crate::types`, exposed via `DownloadOutput::integrity_checks()`, computed in `build_integrity_checks`. Read the type docs for the semantics.
5. [fix download terminal race](https://github.com/awslabs/aws-s3-transfer-manager-rs/commit/4962a411df4e36142c1d8cc2439494a6c9fcc334) — error paths go terminal under the lock before any wake. See the comments at `fail_range` and the two body-error sites.
6. [fill in integrity tests](https://github.com/awslabs/aws-s3-transfer-manager-rs/commit/59b391a7b5f4b5e7dcb83e2660c7227e6a613de3) — `harness.rs` (`Target = Backend × BucketKind`, one body per cell) and `integrity.rs` (assert outcomes: bytes round-trip; every guarantee has a paired tamper-must-error test).
7. [auto align multipart size](https://github.com/awslabs/aws-s3-transfer-manager-rs/commit/460b80ba3f78b0a1cb209db5a2602cec910166cd) — the validate-by-default mechanism. `discover_obj` re-issues discovery with `partNumber=1` to learn the stored part size and threads it through `effective_part_size` → `Transferring.part_size` → the range slicer. Also drops the now-dead `async-channel`/`blocking` deps.
8. [wire up express and prod s3 into harness](https://github.com/awslabs/aws-s3-transfer-manager-rs/commit/b7e80a015d44135ad4893ed30ad343218e7e6f0a) — real-S3 and S3 Express targets behind `--cfg e2e_test`, un-ignores `test_objects_transfer`, runs the suite in CI's e2e step.

### Range alignment

When validation is on, there is no user range, the object is multipart (ETag carries `-N`), and the user has not pinned an explicit part size, `discover_obj` issues a second discovery GET with `partNumber=1`. Its `Content-Range` reports the exact first-part length, which is the stored part size. Total size and part count from a ranged response cannot give this: a multipart object's parts are (typically) uniform except a possibly-smaller last part, so `total / N` is ambiguous, and only a `partNumber` GET returns the stored size exactly. That size becomes `effective_part_size`, carried into `Transferring.part_size`, so every subsequent range aligns to a stored boundary (including the ragged tail) and S3 returns each part's checksum for the SDK to validate. It costs one extra headers-mostly request, only on this path. An explicit user part size is respected as set and validates only if it matches the upload.

### Integrity result

`DownloadOutput::integrity_checks()` returns the object's reported checksum values and a `ChecksumValidation` verdict. Values are surfaced only when the whole object arrived in the discovery chunk; a multipart object's discovery chunk carries part 1's checksum, not the object's, so its value members stay `None`. `Validated { algorithm }` means the entire delivered object was covered and matched; anything less is `NotValidated { reason }`. A checksum mismatch is not represented here, it is an `Err` from `join()`.

The verdict is currently conservative: `NotValidated { Unavailable }` when validation is enabled, `NotValidated { Disabled }` when it is not. Reporting `Validated` requires a per-response validation outcome from the SDK, which validates lazily as the body is consumed and records nothing observable at the operation boundary. The types are `#[non_exhaustive]` so adding the positive verdict later is not a breaking change. The negative guarantee holds regardless: a tampered or wrong-checksum download fails `join()`, and on the file-sink path it fails before the temp-to-final rename, so a corrupt object is never published.

### Mock fidelity

The mock is a dev-loop convenience, not an authority. The composite checksum value, ranged-GET checksum behavior (aligned / unaligned / whole-object / ragged tail), and single-PUT ranged behavior were each probed against real S3 first, and the mock changes encode the verified behavior. The contract suite runs against the mock in CI and against real S3 (general-purpose and Express) under `--cfg e2e_test`.

### Transfer manager

- `execute_get_range` builds the chunk GET from `From<DownloadInput>` (was a four-field hand-built request).
- `discover_obj` takes `validation_enabled` and aligns multipart download ranges to the stored part size via a `partNumber=1` re-issue. Adds `discovery.effective_part_size`, `DownloadState::Transferring.part_size`, `parts_from_etag`.
- Error paths (`fail_range`, body read, flush) go terminal before any wake.
- New `IntegrityChecks` / `ChecksumValidation` / `NotValidatedReason` in `crate::types`; `DownloadOutput::integrity_checks()`.
- `Handle::user_set_part_size()`. Removed dead `target_part_size_bytes()`.
- Dropped `async-channel` and `blocking` (dead since the scheduler migration).
- Un-ignored `test_objects_transfer`; its stale FIXME (objects paths on the old scheduler) was resolved by #147/#148.

### s3-mock-server

- `faults.rs`: key-scoped fault registry with deterministic firing; `insert_fault` / `clear_fault` on `S3MockServer`; `WrongStoredChecksum` and `CorruptBody` in `get_object`.
- Composite checksum computed over raw decoded part bytes with the `-N` suffix (both backends).
- Range-aware GET checksums: per-part checksum on an aligned range, none on unaligned, object value on whole-object. Adds `ObjectMetadata.parts`, `part_range`, `parts_count`, and `partNumber` GET support.

### Tests and CI

- `integration-tests/harness.rs`: `Target = Backend × BucketKind`.
- `integration-tests/integrity.rs`: object-integrity contract suite (mock and real-S3 wrappers).
- `download_test.rs`: integrity-checks unit coverage.
- CI e2e step also runs `cargo test -p integration-tests` under `--cfg e2e_test`.

## What changed

```
operation/download/transfer.rs       - range-GET field forwarding, terminal-before-wake,
                                        build_integrity_checks, part_size in Transferring
operation/download/discovery.rs      - validation_enabled, partNumber=1 realignment,
                                        effective_part_size, parts_from_etag
operation/download/context.rs        - Transferring.part_size
operation/download/output.rs         - DownloadOutput::integrity_checks()
operation/download/handle.rs         - join() populates integrity_checks
types.rs                             - IntegrityChecks, ChecksumValidation, NotValidatedReason
client.rs                            - Handle::user_set_part_size()
tests/download_test.rs               - integrity-checks coverage
tests/e2e_transfer_test.rs           - un-ignore test_objects_transfer
Cargo.toml                           - drop async-channel, blocking
s3-mock-server/src/faults.rs         - New: fault registry (key-scoped, deterministic)
s3-mock-server/src/s3s.rs            - apply faults; partNumber GET; range-aware checksums
s3-mock-server/src/server.rs         - insert_fault/clear_fault, registry plumbing
s3-mock-server/src/storage/*.rs      - composite over raw bytes + -N; parts retained on MPU
s3-mock-server/src/types.rs          - Express bucket support
s3-mock-server/tests/{checksums,faults}.rs - fidelity + fault tests
integration-tests/src/harness.rs     - New: Target = Backend x BucketKind
integration-tests/src/integrity.rs   - New: object-integrity contract suite
.github/workflows/ci.yml             - run integration-tests in the e2e step
```

## Future

- Validate a large single-PUT object split into parallel ranges. S3 returns no checksum for a sub-range of a single stored part, so this needs a transfer-manager-computed whole-object checksum over the delivered bytes plus a HeadObject for the expected value. An `#[ignore]`d intent test states the goal.
- Report `Validated { algorithm }` once the SDK exposes a per-response validation outcome.
- A typed `ErrorKind::ChecksumMismatch` carrying algorithm and expected/computed, folded into the error-hierarchy work.
- Fold the legacy `e2e_transfer_test.rs` cases into the harness. S3 Express is real-S3-only until the mock models directory buckets.

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
+++ b/aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
@@ -564,19 +564,10 @@ impl DownloadObjectsTransfer {
                         error: err,
                     });
                     if self.inner.failure_policy == FailedTransferPolicy::Abort {
-                        self.inner.ctx.set_failed(Error::new(
+                        self.inner.ctx.set_failed_and_signal(Error::new(
                             ErrorKind::ChildOperationFailed,
                             format!("download failed for key '{key}'"),
                         ));
-                        // set_failed makes the transfer terminal but does not
-                        // signal; signal here so the terminal transition is
-                        // observable before returning to the scheduler. A
-                        // terminal transfer that returns idle without signaling
-                        // can be removed by the scheduler with its completion
-                        // channel never fired, hanging the owning handle. Any
-                        // surviving children are cancelled when the handle's
-                        // join()/drop calls cancel_transfer(parent_id).
-                        self.inner.ctx.signal_terminal();
                     }
                 }
             }
@@ -707,6 +698,7 @@ impl DownloadObjectsTransfer {
             if let Err(err) = self.validate_destination() {
                 let mut state = self.inner.state.lock();
                 state.walk_in_flight = false;
+                // set_failed alone: check_terminal arbitrates the terminal signal once in-flight work (children/walks/reaps) drains.
                 self.inner.ctx.set_failed(err);
                 if self.check_terminal(&state).is_some() {
                     return WorkOutcome::Success { data: None };
@@ -829,19 +821,10 @@ impl DownloadObjectsTransfer {
                         error: err,
                     });
                     if self.inner.failure_policy == FailedTransferPolicy::Abort {
-                        self.inner.ctx.set_failed(Error::new(
+                        self.inner.ctx.set_failed_and_signal(Error::new(
                             ErrorKind::ChildOperationFailed,
                             format!("download failed for key '{key}'"),
                         ));
-                        // set_failed makes the transfer terminal but does not
-                        // signal; signal here so the terminal transition is
-                        // observable before returning to the scheduler. A
-                        // terminal transfer that returns idle without signaling
-                        // can be removed by the scheduler with its completion
-                        // channel never fired, hanging the owning handle. Any
-                        // surviving children are cancelled when the handle's
-                        // join()/drop calls cancel_transfer(parent_id).
-                        self.inner.ctx.signal_terminal();
                     }
                 }
             }
--- a/aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
+++ b/aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
@@ -568,6 +568,15 @@ impl DownloadObjectsTransfer {
                             ErrorKind::ChildOperationFailed,
                             format!("download failed for key '{key}'"),
                         ));
+                        // set_failed makes the transfer terminal but does not
+                        // signal; signal here so the terminal transition is
+                        // observable before returning to the scheduler. A
+                        // terminal transfer that returns idle without signaling
+                        // can be removed by the scheduler with its completion
+                        // channel never fired, hanging the owning handle. Any
+                        // surviving children are cancelled when the handle's
+                        // join()/drop calls cancel_transfer(parent_id).
+                        self.inner.ctx.signal_terminal();
                     }
                 }
             }
@@ -824,6 +833,15 @@ impl DownloadObjectsTransfer {
                             ErrorKind::ChildOperationFailed,
                             format!("download failed for key '{key}'"),
                         ));
+                        // set_failed makes the transfer terminal but does not
+                        // signal; signal here so the terminal transition is
+                        // observable before returning to the scheduler. A
+                        // terminal transfer that returns idle without signaling
+                        // can be removed by the scheduler with its completion
+                        // channel never fired, hanging the owning handle. Any
+                        // surviving children are cancelled when the handle's
+                        // join()/drop calls cancel_transfer(parent_id).
+                        self.inner.ctx.signal_terminal();
                     }
                 }
             }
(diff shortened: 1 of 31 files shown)


aws-sdk-s3-transfer-manager/src/types.rs
```diff
@@ -253,3 +253,123 @@ impl Bucket {
...
+#[non_exhaustive]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ChecksumValidation {
+    /// Every delivered byte was validated against a checksum and matched.
+    Validated {
+        /// The algorithm used to validate.
+        algorithm: aws_sdk_s3::types::ChecksumAlgorithm,
+    },
+    /// No whole-object validation occurred. `reason` explains why.
+    NotValidated {
```
ysaito1001, line 333: Do these variants themselves need `#[non_exhaustive]` so more fields can be added to the future within them?

aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
```diff
@@ -568,6 +568,15 @@ impl DownloadObjectsTransfer {
...
                         ));
+                        // set_failed makes the transfer terminal but does not
+                        // signal; signal here so the terminal transition is
+                        // observable before returning to the scheduler. A
+                        // terminal transfer that returns idle without signaling
+                        // can be removed by the scheduler with its completion
+                        // channel never fired, hanging the owning handle. Any
+                        // surviving children are cancelled when the handle's
+                        // join()/drop calls cancel_transfer(parent_id).
+                        self.inner.ctx.signal_terminal();
```
ysaito1001: Long comment seems to indicate calling `signal_terminal` is crucial. It's also done at line 844 (new revision).

Should we bundle `self.inner.ctx.set_failed` and `self.inner.ctx.signal_terminal` into one helper to make it atomic unless some call sites can legitimately call `set_failed` without `signal_terminal`?

integration-tests/src/integrity.rs
```diff
@@ -0,0 +1,979 @@
...
+
+// tamper -> error on the FILE path (validate before rename) ---------------------
+//
+// The in-memory tamper test above drains the body; this exercises the managed
+// file-sink path (slot buffer -> pwrite -> temp file -> rename), where the
+// integrity contract is "validate before rename": a checksum failure must error
+// AND leave no file at the destination (the temp is cleaned up). A corrupt body
+// landing at the destination would be silent on-disk corruption.
+
+async fn tampered_single_part_file_errors(target: Target) {
```
ysaito1001, line 586: Cool test to have 💡

ysaito1001 (approved):
Like the idea of commit-by-commit guide. LGTM.

aws-sdk-s3-transfer-manager/src/types.rs
aajtodd: Ya the enum-level attribute only covers the variant list, not fields within a variant. Added `#[non_exhaustive]` to both.

aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs
aajtodd: Bundled into `set_failed_and_signal` and migrated the immediate-abort sites. Kept them separable for the drain-gated paths (validate-destination, fatal walker), where the signal must wait for in-flight work to drain. Lifted the rationale onto the methods.

PR #150, opened 2026-06-17
Download body-read retry, error-hierarchy rework, and per-bucket retry partitioning

aajtodd:
## Motivation

Three pieces of download/upload resilience.

**Body-read retry.** The TM downloads via parallel ranged GETs and streams each
chunk's body after the SDK's GetObject orchestration has returned `Ok`. A failure
reading that body — connection reset, truncated body, short read, stall — happens
outside the SDK's retry scope, so the SDK cannot recover it. `fail_range` was
terminal, so one transient body failure on any chunk failed the whole transfer.

**Error-hierarchy rework.** The classifier has to distinguish a transient
transport failure from a checksum mismatch, but both were `ErrorKind::IOError`,
forcing a source-chain downcast to tell them apart. Separately, the error surface
dropped request IDs and S3 metadata (`// TODO - extract request id`), classified the
same logical failure as different kinds on different code paths, and modeled "chunk"
as an error kind rather than as metadata on an error.

**Per-bucket retry partitioning.** The SDK shares one retry token bucket per
`RetryPartition`, and all operations on a client default to a single region-wide
`s3-{region}` partition. The TM fans out hundreds of concurrent requests; a throttle
storm on one bucket then drains the shared retry budget for unrelated buckets on the
same client. This keys the partition per bucket which matches CRT per/host.

### Retry loop

`LatencyTracker::guarded` runs one attempt under the adaptive latency deadline. On
timeout it drops the in-flight future (which tears down the connection) and returns
`GuardError::DeadlineExceeded`; otherwise it returns the inner result as
`GuardError::Inner`. It holds no retry policy. The `retry` module owns the loop and
asks a classifier, per failure, whether to re-issue:

```rust
match tracker.guarded(build()).await {
    Ok(v) => return Ok(v),
    Err(ge) => match classify(&ge) {
        RetryDecision::NoRetry => return Err(into_error(ge)),
        RetryDecision::Retry { .. } if attempt >= MAX_ATTEMPTS => return Err(into_error(ge)),
        RetryDecision::Retry { after } => { /* optional backoff; rebuild and re-issue */ }
    },
}
```

`classify_body_retry` (download) retries a deadline timeout and a body-stream
`IOError`, which are the failures the SDK does not see. It does not retry an
`IntegrityError` (a checksum mismatch is not re-fetched, so corruption is not masked)
or an SDK-derived send error (the SDK already retried it). Throttle and 5xx are not
handled here; they belong to the SDK token bucket. The retry closure rebuilds an
identical request — same range, etag, version — so a re-issue reads the same bytes.

The upload part-send path uses the same loop with `retry_deadline_only`: retry on
deadline only, since the SDK already retries UploadPart over its rewindable in-memory
body. PutObject uses SDK retry alone — a single PutObject body is variable-size (below
the part size, and possibly a user-supplied range), so the part-calibrated deadline
does not apply to it.

Body-read retry covers the range GETs in `execute_get_range` and the discovery body in
`execute_read_discovery_body`. The discovery GET and its body read are co-located into
one retryable unit, preserving the partNumber=1 alignment used for per-part checksum
validation.

### Error hierarchy

`ErrorKind` gains `ServiceError` and `IntegrityError(IntegrityError)`. "Chunk" moves
from a kind to metadata, `ChunkRef { seq, range }`. The common `Error` stays small:
service metadata, chunk ref, and per-object failure lists sit behind an
`Option<Box<ErrorExtra>>`, boxed to keep `Result` small enough for `result_large_err`.
Accessors:

```rust
err.operation_name();        // operation that failed (captured before type erasure)
err.code();                  // S3 error code
err.request_id();            // x-amz-request-id
err.extended_request_id();   // x-amz-id-2
err.is_not_found();          // NoSuchKey / 404
err.chunk();                 // ChunkRef { seq, range }
err.failed_uploads();        // per-object failures on a bulk Err
err.failed_downloads();
```

`From<SdkError<E, HttpResponse>>` is implemented per operation (8 ops) so the
operation name, which is not recoverable after type erasure, is captured at the
conversion. `From<WalkError>` is kind-aware: a listing `Service` error recovers the
underlying `SdkError<ListObjectsV2Error>` to a `ServiceError` with full metadata;
`SourceUnreadable`/`NotADirectory` become `InputInvalid`. The checksum-mismatch
downcast now happens once, at the body-read boundary in `error::body_read_error`,
producing an `IntegrityError` with base64 expected/computed values; the classifier
keys on the kind and no longer walks the source chain.

### Per-bucket retry partition

`bucket_retry_partition(bucket)` returns `RetryPartition::new("s3-tm-{bucket}")`,
attached via `config_override` at each send site (download discovery, range GET, HEAD;
upload PutObject, UploadPart, Create/Complete/Abort MPU). The SDK's
`TokenBucketProvider` interceptor reads `RetryPartition` from the config bag per
operation, so the override re-keys the global token-bucket map for that operation
without rebuilding the client. The `_objects` operations inherit it through the
singular send sites. Capacity stays at 500 (CRT parity).

Per-bucket is the coarsest key that does not couple independent workloads, and its
cardinality is bounded by the number of buckets a process touches. A per-prefix key
would instead guess at S3's internal partition boundaries and grow unbounded in the
SDK's never-evicting partition map.

## How to review

`retry.rs` is the core: the loop, `classify_body_retry`, `retry_deadline_only`, and
`bucket_retry_partition`. `metrics/latency.rs` shows `guarded` reduced to a
single-attempt deadline primitive. `error.rs` is the largest change — the `ErrorKind`
reshape, the `ErrorExtra` side-channel, the per-operation and `WalkError` conversions,
and the accessors. The send-site wiring in `operation/download/{transfer,discovery}.rs`
and `operation/upload/{transfer,handle}.rs` is mechanical: each call routes through
the retry loop and attaches the partition override.

The mock-server changes are test infrastructure. `s3-mock-server/src/faults.rs` and
`socket_fault.rs` add the fault types used by the tests — body-stream faults
(`TruncateBody`, `StallBody`, `ShortBody`), a per-connection `ConnectionReset`, and a
send-time `ServiceError { status }` — each producing a distinct real wire error so the
classifier is exercised against genuine errors rather than synthetic ones. The tests
live in `integration-tests/src/retry.rs` and the `retry_tests` module in
`operation/upload.rs`.

## Testing

The classifier arms and the loop (timeout-retry, exhaustion, no-retry-on-inner) are
unit-tested in `retry.rs`. `integration-tests/src/retry.rs` drives the mock-server
faults through the TM and reads behavior by contrast: a fault as `Once` recovers, as
`Always` exhausts to failure — across `TruncateBody`, `ConnectionReset`, `ShortBody`,
`StallBody`, on the range and discovery paths. A standing guard pins that a corrupt
body is not retried.

`test_upload_part_deadline_retry` stalls the first UploadPart past the warm adaptive
deadline with SDK retry disabled, so only the TM layer can recover it; neutering
`retry_deadline_only` fails it. The error hierarchy adds Display and `is_not_found`
unit tests and an integrity backwards-compat test that drives a real `CorruptBody`
mismatch and asserts the `IntegrityError` fields populate, so an SDK error-shape change
fails the test rather than silently degrading classification.
`requests_carry_per_bucket_retry_partition` records the `RetryPartition` the SDK
resolves per request through an interceptor and asserts distinct `s3-tm-{bucket}` keys;
a forced shared partition fails it. (The partition key is asserted directly because the
runtime drain-and-observe effect is not stable to assert — the adaptive token bucket
recovers between retries, so it cannot be held empty without races.)

TM lib 426, integration-tests 55 (+1 ignored), download 17, download_objects 10,
upload_objects 27 (+2 ignored), upload_checksum 25, mock-server 42+39+12+14. `fmt` and
`clippy -Dwarnings` clean across all three crates.

## Changes

```
aws-sdk-s3-transfer-manager/
  src/
    retry.rs                          NEW — retry loop, classifiers, partition helper
    error.rs                          ErrorKind reshape, metadata side-channel, accessors
    metrics/latency.rs                guarded → single-attempt deadline primitive
    operation/
      download/{transfer,discovery,input}.rs   body-read + discovery retry; fluent GET helper
      upload/{transfer,handle}.rs              retry_deadline_only; per-bucket config_override
    examples/cp.rs                    print operation/code/request-id on failure
s3-mock-server/
  src/
    faults.rs                         body-stream + ServiceError fault types
    socket_fault.rs                   NEW — per-connection transport faults
    s3s.rs, server.rs                 fault application + registry
integration-tests/
  src/retry.rs                        NEW — fault-driven retry + partition tests
```


aws-sdk-s3-transfer-manager/src/client.rs
```diff
@@ -17,23 +17,6 @@ use std::time::Duration;
...
-///
-/// The transfer manager owns stall detection for the bodies it drives (an
-/// adaptive, transfer-aware deadline in the download path), and it applies
-/// backpressure: a chunk body may be intentionally left unpolled while the slot
-/// buffer is full. The SDK's body-level minimum-throughput guard cannot tell
-/// that apart from a dead connection and would abort a healthy, backpressured
-/// chunk. A client supplied directly via `client()` is left untouched (the
-/// caller owns its configuration); this applies only to clients the transfer
-/// manager builds itself.
-fn without_stalled_stream_protection(
```
landonxjames, line 31: Did we want to remove this? Feels like the TM should manage SSP instead of relying on the SDK's impl

aws-sdk-s3-transfer-manager/src/error.rs
```diff
@@ -168,40 +351,257 @@ where
...
 
-impl<E, R> From<aws_sdk_s3::error::SdkError<E, R>> for Error
+/// Walks the source chain for a smithy checksum-mismatch error, returning its
+/// expected and computed checksums base64-encoded.
+fn checksum_mismatch_values(err: &(dyn std::error::Error + 'static)) -> Option<(String, String)> {
+    use aws_smithy_checksums::body::validate::Error as ChecksumError;
+    let mut source: Option<&(dyn std::error::Error + 'static)> = Some(err);
+    while let Some(e) = source {
+        if let Some(ChecksumError::ChecksumMismatch { expected, actual }) =
+            e.downcast_ref::<ChecksumError>()
```
landonxjames, line 409: This feels sketchy. Since the TM directly relies on a version of `aws-smithy-checksums` if we bumped the minor version, released `aws-sdk-s3` relying on that new version, but didn't release the TM with that version bump at the same time I think this downcast would fail. Because the S3 crate would be relying on `0.65::ChecksumError` and the TM would be looking for `0.64::ChecksumError` and the `TypeId` wouldn't match. Similar to the issue that came up with https://github.com/smithy-lang/smithy-rs/pull/4691

aws-sdk-s3-transfer-manager/src/retry.rs
```diff
@@ -0,0 +1,266 @@
...
+pub(crate) fn classify_body_retry(ge: &GuardError<Error>) -> RetryDecision {
+    let err = match ge {
+        GuardError::DeadlineExceeded(_) => return RetryDecision::Retry { after: None },
+        GuardError::Inner(e) => e,
+    };
+
+    match err.kind() {
+        // A checksum mismatch must never be re-fetched and masked.
+        ErrorKind::IntegrityError(_) => RetryDecision::NoRetry,
+        ErrorKind::IOError => RetryDecision::Retry { after: None },
```
landonxjames, line 132: IOErrors seem like they should have some kind of backoff/jitter to prevent retry storms?

landonxjames (approved):
Looks good, couple of questions, the downcast issue probably should be fixed before merge.

aws-sdk-s3-transfer-manager/src/error.rs
ysaito1001: Ah, good to raise it. Downcasting is sort of stepping its private parts, which can be fragile as Landon pointed out. I wonder if any of the following should be pursued 
- this classification were offered upstream
- `aws-smithy-checksums` goes 1.x
- move `ChecksumError` to a stable crate

so we don't carry a potential semver hazard 🤔

ysaito1001 (approved):
LGTM, semver hazard on `ChecksumError` should be clarified before GA (not necessarily need to be addressed in this PR)

aws-sdk-s3-transfer-manager/src/error.rs
aajtodd: Created https://github.com/smithy-lang/smithy-rs/issues/4718. Not going to address in this PR but agreed on all points.

aws-sdk-s3-transfer-manager/src/retry.rs
aajtodd: I thought about this a bit but I don't think so. This retry loop is outside of the SDK and only handles errors around downloading the body which at that point S3 has already accepted the request and started processing so we aren't being throttled and is most likely a transient issue. If there is need to backoff I would expect to hit it on the SDK token bucket for the new request.

aws-sdk-s3-transfer-manager/src/client.rs
aajtodd: Removed again.

PR #151, opened 2026-06-23
 fix scheduler work-generation lost wake (EpochGate)

aajtodd:
## Motivation

The HLL integration tests time out on macOS and Windows CI runners. The hang is a
pre-existing lost wake in `Scheduler::generate_work`, present since #147; it surfaces
on those runners because they have fewer cores, which widens the interleaving window
the bug needs. It does not reproduce locally.

`generate_work` admits one work-generation runner at a time through a gate. The gate's
job is to let one thread drive ready-set pops while any concurrent `wake`/`enqueue`/
`on_completion` that cannot acquire the runner role still gets its just-published work
drained — without two runners popping in parallel, which would break the
pop-poll-reinsert cycle that priority ordering depends on.

The original gate recorded a pending request in a single `NOTIFIED` bit. That bit is a
saturating 1-bit counter: two concurrent requests collapse into one. The runner
services the bit once, sees it clear, and retires — stranding the second request's
work with no runner and no edge left to re-drive it. The transfer then hangs forever
waiting on work that was published but never generated.

The window is real but narrow, which is why it only bites under the extra preemption of
a few-core runner. The completion path makes it worse: a completing thread decrements
`dispatched`, then finds the gate already `NOTIFIED`, so it publishes no gate edge and
establishes no happens-before with the runner. The runner can then retire on a stale
view of capacity even though a slot just freed.

### EpochGate

The gate packs two fields into one `AtomicUsize`: bit 0 is the `RUNNING` flag, bits
1.. are a monotonic request *epoch*. A caller that finds a runner active bumps the
epoch instead of setting a flag, so concurrent requests are counted, not coalesced.

```
 ┌──────────────── usize ─────────────────┐
 │  request epoch (bits 1..)         │ R  │   R = RUNNING
 └───────────────────────────────────┴────┘
```

The runner captures the epoch when it acquires, re-syncs it after each pass, and
retires only through a CAS that requires the epoch to be unchanged — never a blind
store:

```rust
fn release(&self, observed: usize) -> Release {
    let expected = (observed << 1) | RUNNING;
    let idle     =  observed << 1;          // clear RUNNING, same epoch
    match self.packed.compare_exchange(expected, idle, SeqCst, SeqCst) {
        Ok(_)       => Release::Idle,        // no request since `observed`; retire
        Err(actual) => Release::Pending(actual >> 1), // epoch moved; keep role, drain again
    }
}
```

A request that bumps the epoch makes the retire CAS fail, so the runner observes it and
drains another pass. The bumping caller publishes its work before bumping, so the
runner's failing CAS reads-from the bump and the next pass sees the work. No request is
lost regardless of how many race, because each increments the epoch rather than setting
a shared bit.

The packed atomic is isolated as `GateState` (the bit-level state machine) under the
`GenerateWorkGate` role policy. `try_acquire` / `try_release` are unchanged at the call
site.

### Stale-capacity recheck at retire

`has_capacity` reads `dispatched` with a plain load. That is correct in the dispatch
hot loop, where the runner just observed its own writes, but not at the retire
decision, where a coalesced wake can hide a concurrent completion's slot release. The
retire path uses a read-modify-write instead:

```rust
fn recheck_capacity(&self) -> bool {
    self.0.dispatched.fetch_add(0, Ordering::SeqCst) < target
}
```

The RMW reads the latest value in `dispatched`'s modification order, so it observes the
completion's decrement; `SeqCst` places this read and the gate's retire CAS in one
total order, so the runner and the completing thread agree on which of them drives the
freed slot. It is used only at the retire check, not per-pop, so its cost is off the
hot path. The epoch does not subsume this: the epoch closes the lost-request race, the
RMW closes the stale-capacity race, and both are needed.

### Bounded generation passes

A runner that keeps finding the epoch bumped would drain passes indefinitely while
sustained wakes arrive. On a managed runtime each scheduler thread is a current-thread
tokio runtime, so a thread pinned in `generate_work` cannot poll its own in-flight
`execute` futures — the source of the fairness regression seen at high concurrency.

`MAX_GENERATION_PASSES = 3` bounds the proactive work. After the bound, if an in-flight
completion is guaranteed to re-drive generation (`dispatched >= 1`), the runner retires
and lets the runner role rotate to whichever thread completes next — the completion
edge is a free baton, no new primitive or blocking waiter. This honors the scheduler's
`on_completion = O(target × poll_work)` cost contract.

The `dispatched == 0` corner is the exception: there is no in-flight item, so no
completion edge exists to re-drive. The bound does not apply there and the runner keeps
draining, self-bounded by the finite published work. This corner is exactly what
strands a 1-bit gate (the producer publishes while the last item completes), and it is
the case the epoch is required for.

The retire is safe across the bound handoff. A completion racing in decrements
`dispatched` before bumping the epoch via `try_acquire`; the RMW observes that
decrement, and the epoch-gated retire CAS fails if any request bumped past the observed
epoch, so a racing request is always seen, never stranded.

## How to review

`gate.rs` is the core: `GateState` (the packed atomic and its transitions),
`GenerateWorkGate` / `GenerateWorkPermit` (the role policy and panic-safe `Drop`
fallback), and the module doc stating the no-lost-wake guarantee. `scheduler.rs` holds
the retire loop, `recheck_capacity`, and `MAX_GENERATION_PASSES`.

The protocol is model-checked under loom (`cfg(s3_tm_loom)`, `LOOM_MAX_PREEMPTIONS=3`),
driving the production gate directly. Eight tests cover the no-lost-wake property across
the request-count and capacity dimensions:

- `..._no_lost_wake`, `..._no_lost_wake_three_callers` — single and three concurrent
  requests against an idle gate.
- `..._no_lost_wake_at_capacity`, `..._at_capacity_with_producer` — a completion frees a
  slot while a producer publishes; the `with_producer` case is the one the RMW recheck
  is required for.
- `..._bounded_handoff`, `..._bounded_handoff_corner` — the bounded-pass retire hands
  off to the completion edge without dropping a request, including the `dispatched == 0`
  corner.
- `..._corner_unbounded` — the exact configuration that strands a 1-bit gate (`k = MAX`);
  passes with the epoch.
- `..._no_lost_wake_drive_path` — the drive-after-acquire path.

Each was teeth-checked by mutating the production code to reintroduce the bug (a
blind clear-`RUNNING` in `release`) and confirming the model fails. The bit-level
transitions also have eight fast `#[cfg(test)]` unit tests on `GateState`
(acquire-from-idle, record-and-bump, release-at-same/stale-epoch, multi-bump,
`force_idle` preserves epoch), so an off-by-one in the epoch shift is caught by
`cargo test` without running loom.

`test_two_composites_fair_share_*` was re-stabilized. CFS fairness is a steady-state
property; the prior sample at N=500 measured a startup transient (the composite whose
children win the initial ready-set race burst ahead before vruntime accounting pulls
them level). Skew decays ~35% at N=50 → ~4% at N=500 → <2% at N=1200. The test now
samples at N>=1200 with 10% tolerance; the worst observed skew there across hundreds of
runs was ~4%.

## Changes

```
aws-sdk-s3-transfer-manager/src/scheduler/
  gate.rs          EpochGate: GateState packed atomic + role policy; 8 unit + 8 loom tests
  scheduler.rs     bounded retire loop, recheck_capacity (RMW), MAX_GENERATION_PASSES;
                   fairness test sampled at steady state
  concurrency.rs   document the ConcurrencyController::target() >= 1 contract the
                   retire-at-capacity path relies on
```

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/scheduler/gate.rs
+++ b/aws-sdk-s3-transfer-manager/src/scheduler/gate.rs
@@ -11,35 +11,143 @@
 
 use crate::runtime::sync::sync::atomic::{AtomicUsize, Ordering};
 
-/// No runner active and no request pending.
-const IDLE: usize = 0;
-/// A runner holds the permit and is draining generation passes.
+/// Bit 0 of the packed state: set while a runner holds the role.
 const RUNNING: usize = 1;
-/// A runner is active AND a request arrived while it ran; the runner must
-/// drain once more before retiring.
-const NOTIFIED: usize = 2;
+/// Increment applied to the packed state to bump the request epoch (bits 1..).
+const EPOCH_STEP: usize = 2;
 
-/// Admits one work-generation runner at a time and coalesces concurrent
-/// requests so none is lost.
+/// Outcome of [`GateState::acquire`].
+enum Acquire {
+    /// This caller became the sole runner; it observed the gate at `epoch` and
+    /// must drive passes until [`GateState::release`] reports `Idle`.
+    Runner { epoch: usize },
+    /// A runner was already active; this caller bumped the epoch to record its
+    /// request and the active runner will drain an extra pass for it.
+    Recorded,
+}
+
+/// Outcome of [`GateState::release`].
+enum Release {
+    /// No request arrived since the observed epoch; the gate returned to idle
+    /// and the runner role is released. Stop draining.
+    Idle,
+    /// A request bumped the epoch during the pass; the role is retained. Carries
+    /// the new epoch to re-sync, then drain again.
+    Pending(usize),
+}
+
+/// The packed runner/epoch atomic, with the bit-level state machine isolated
+/// from the [`GenerateWorkGate`] role policy that wraps it.
+///
+/// One `usize` holds both: bit 0 is the `RUNNING` flag, bits 1.. are a monotonic
+/// request *epoch*.
+///
+/// ```text
+///  ┌──────────────── usize ─────────────────┐
+///  │  request epoch (bits 1..)         │ R  │   R = RUNNING (bit 0)
+///  └───────────────────────────────────┴────┘
+/// ```
+///
+/// A caller that finds a runner active bumps the epoch rather than setting a
+/// shared flag, so two concurrent requests are *counted*, not coalesced into one
+/// bit. (A single saturating bit loses the second of two concurrent requests: the
+/// runner services it once and releases, stranding the other — the failure that
+/// motivated the epoch.)
+///
+/// All transitions are `SeqCst`. The epoch occupies all but the low bit and may
+/// wrap; this is benign for an eventcount. A lost wake would require the epoch
+/// to advance by exactly a multiple of `2^(usize::BITS-1)` between a pass ending
+/// and its release CAS — i.e. that many concurrent requests in that window —
+/// which cannot occur.
+struct GateState {
+    packed: AtomicUsize,
+}
+
+impl GateState {
+    fn new() -> Self {
+        Self {
+            packed: AtomicUsize::new(0),
+        }
+    }
+
+    /// Become the sole runner (epoch unchanged), or record a request by bumping
+    /// the epoch. The CAS-retry loop terminates: a retry happens only when some
+    /// other thread's CAS succeeded, so the system made progress.
+    fn acquire(&self) -> Acquire {
+        loop {
+            let s = self.packed.load(Ordering::SeqCst);
+            if s & RUNNING == 0 {
+                // Idle: claim the runner role, leaving the epoch unchanged.
+                if self
+                    .packed
+                    .compare_exchange_weak(s, s | RUNNING, Ordering::SeqCst, Ordering::SeqCst)
+                    .is_ok()
+                {
+                    return Acquire::Runner { epoch: s >> 1 };
+                }
+            } else {
+                // A runner is active: record a request by bumping the epoch.
+                if self
+                    .packed
+                    .compare_exchange_weak(s, s + EPOCH_STEP, Ordering::SeqCst, Ordering::SeqCst)
+                    .is_ok()
+                {
+                    return Acquire::Recorded;
+                }
+            }
+            // CAS failed (weak spurious failure, or the state changed under us);
+            // re-observe and retry.
+        }
+    }
+
+    /// Release the runner role iff no request arrived since `observed`; otherwise
+    /// return the new epoch so the runner re-syncs and drains again. The release
+    /// is a CAS gated on the epoch — never a blind store — so a request that
+    /// raced the release is always observed, not stomped.
+    fn release(&self, observed: usize) -> Release {
+        let expected = (observed << 1) | RUNNING;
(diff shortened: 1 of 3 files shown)


aws-sdk-s3-transfer-manager/src/scheduler/gate.rs
```diff
@@ -11,35 +11,143 @@
...
+        {
+            Ok(_) => Release::Idle,
+            Err(actual) => Release::Pending(actual >> 1),
+        }
+    }
+
+    /// Clear `RUNNING` (preserving the epoch). Panic-path only; see
+    /// [`GenerateWorkPermit`]'s `Drop`.
+    fn force_idle(&self) {
+        let s = self.packed.load(Ordering::SeqCst);
```
landonxjames: Instead of two operations I think this could just be a single atomic  `fetch_and`

aws-sdk-s3-transfer-manager/src/scheduler/gate.rs
```diff
@@ -218,6 +443,79 @@ mod loom_tests {
...
+                        }
+                    }
+                    // At capacity or bound-handoff: retire and retry the CAS
+                    // release rather than spin a pass that would dispatch nothing.
+                }
+            }
+        }
+
+        /// Free one in-flight slot then re-drive — the exact order of
+        /// `Scheduler::on_completion` (`dispatched.fetch_sub` @436, then
```
landonxjames: Line numbers probably unnecessary/will become outdated on next change to this file

aws-sdk-s3-transfer-manager/src/scheduler/scheduler.rs
```diff
@@ -527,7 +537,37 @@ impl Scheduler {
...
+    }
+
+    /// `has_capacity`, but reading `dispatched` with a read-modify-write so the
+    /// runner cannot retire on a stale view of capacity.
+    ///
+    /// The plain-load `has_capacity` is correct in the dispatch hot loop, where
+    /// the runner just observed its own `dispatched` writes. It is NOT safe at
+    /// the retire decision: when a runner re-arms off a *coalesced* notification
+    /// (a `wake`/`enqueue` CAS'd `RUNNING → NOTIFIED`), a concurrent
+    /// `on_completion` may have decremented `dispatched` and then found the gate
```
landonxjames: Isn't this describing the case the EpochGate protects against?

landonxjames (approved):
Looks good, couple of questions/small suggestions

aws-sdk-s3-transfer-manager/src/scheduler/gate.rs
aajtodd: Yup good catch. Fixed.

aws-sdk-s3-transfer-manager/src/scheduler/gate.rs
aajtodd: Yup removed.

aws-sdk-s3-transfer-manager/src/scheduler/scheduler.rs
aajtodd: It is comment is out of date, fixed.

PR #153, opened 2026-07-02
download: occupancy-bounded flow control over a paged receive buffer

aajtodd:
## Summary

Replaces the download delivery path's fixed-capacity slot ring with a paged receive buffer and an occupancy-bounded read-ahead gate. The buffer grows to absorb head-of-line backlog and reclaims from the front as delivery advances; issuance is gated on resident occupancy (`issued - released < window`) rather than on a fixed ring capacity. This fixes a disk-download wedge, decouples the memory bound from the delivery buffer, and adds a public read-ahead control surface.

## Why

The base delivery path is `SlotBuffer`: a fixed-size ring of `capacity` slots indexed by `seq % capacity`. Three limits motivate the change.

**One fixed structure serves three roles.** The ring's capacity is simultaneously the issuance permit (`claimed - consumed < capacity` gates the next ranged GET), the delivery buffer, and the memory bound. They cannot move independently: the buffer cannot grow to absorb a burst of out-of-order arrivals without also loosening the memory bound, and the memory bound cannot tighten without also throttling issuance.

**One cursor cannot express two positions.** The ring tracks a single `consumed` cursor — the in-order delivery position. That is the right bound for a streaming consumer, but a download to disk does not consume in order: it drains filled parts to their file offsets as they arrive, out of order, and never advances an in-order cursor. Resident occupancy (how much buffer is live) and in-order delivery position (how far the stream has been read) are different quantities, and one cursor cannot be both.

**The disk-download wedge.** Because issuance was gated on the in-order cursor, a disk download larger than the ring latched shut. It issued a ring's worth of parts, drained them to disk (freeing the buffer), but the in-order cursor never advanced past the first not-yet-delivered part — so `claimed - consumed` stayed at capacity and issuance never resumed. Observed as a multi-GiB disk download that received a ring's worth, then stalled with zero further bytes written and falling RSS.

### PagedRecvBuffer

`PagedRecvBuffer<T>` assigns each payload a sequence number at claim time, accepts completed payloads into their slots in any order from many producers, and delivers them to a single consumer. It is a `VecDeque<Arc<Segment<T>>>` under one `Mutex`; each segment is a fixed run of slots covering a contiguous span of sequences. The deque grows at the tail and reclaims from the front, so resident memory tracks the gap between the oldest undelivered and newest claimed sequence, not the total count.

```text
  reclaim ◀── front                        tail ──▶ grow
  ┌─────────┐    ┌─────────┐    ┌─────────┐
  │  seg 0  │─nx▶│  seg 1  │─nx▶│  seg 2  │
  │ x x x x │    │ x x x . │    │ F . _ _ │
  │ 0 1 2 3 │    │ 4 5 6 7 │    │ 8 9     │
  │ base=0  │    │ base=4  │    │ base=8  │
  └─────────┘    └───────▲─┘    └─────────┘
                         │
                    cursor=7   (issued=10)

  x = delivered   F = filled, awaiting delivery
  . = claimed, in flight   _ = not yet claimed   nx = next pointer
```

The two hot paths are lock-free: `fill` writes a claimed slot and publishes it with a `Release` store; `poll_next` reads the cursor slot under `Acquire` and hops to the successor segment via an intrusive `next` pointer without taking the lock. Only the issuer paths — `claim` (assign a sequence, maybe grow, opportunistic reclaim) and segment reclaim — take the lock, in short critical sections that never span a fill, a delivery, or an I/O.

### The occupancy gate

Issuance of a speculative part requires `issued - released < window`. `issued` counts parts claimed for issuance; `released` counts parts whose payload memory has been freed. `issued - released` is resident occupancy in parts — an exact quantity, not an estimate. The gate bounds how far issuance runs ahead of consumption and nothing more:

- Fast consumer: `released` keeps pace, occupancy stays low, the gate does not bind.
- Slow consumer: occupancy fills to the window, the gate closes, issuance waits. Resident memory is bounded at `window` parts.
- Blocked consumer (a stalled or not-yet-filled part): occupancy pins at the window, leaving a full window of slack to make progress around the hole rather than collapsing to a single outstanding part.

The window is a fixed per-transfer cap held in an `AtomicU64`, sized above the bandwidth-delay product of a 100 Gbps link so it does not bind a consumer that keeps pace. It is atomic (not a constant) so it can be lowered with a lock-free store while a transfer runs.

### Two surfaces, one occupancy counter

The buffer is consumed through exactly one of two surfaces per instance:

- **Stream** (`RecvBufferConsumer::poll_next`) — one payload at a time, in strict sequence order, emptying each slot as it passes. Its single non-cloneable handle makes "single consumer" a type fact.
- **Block** (`take_drain_run`) — hands out a contiguous filled run of a segment as an owned `SegmentWrite` token, for bulk consumption (disk write) that does not need in-order delivery. Runs may be taken and completed out of order and concurrently.

`released` is incremented by *both* surfaces — one per in-order stream delivery, and by a run's filled count on each block drain. `consumed`, the in-order delivery cursor, is separate and drives only stream-path reclaim. Because the gate reads `issued - released`, a disk download paces to its drain rate instead of latching shut on the in-order cursor — the fix for the wedge. Each range completion repokes the issuer, so a drain that frees occupancy wakes a gated poll.

### The drain primitive

The block surface drains through `take_drain_run`, backed by two per-segment cursors: `claim_cursor` (leading slots claimed for draining, advanced only under the lock, so a run is handed out at most once) and `drained_count` (slots written to the sink and freed; when it reaches the segment length the segment is reclaimable). A drain claims the contiguous filled run beyond `claim_cursor` when it reaches the drain batch, when the segment fills (so a sub-batch tail residue still drains), or when the call is terminal. The batch is a minimum trigger coupled to the read-ahead window (`min(seg_size, window)`), so a window smaller than a segment drains in smaller runs instead of never sealing a segment. Dropping a token without completing runs the same drain, so an aborted consumer still frees its run and cannot stall reclaim.

### Positioned disk writes behind a seam

Disk delivery reads a claimed run's payloads in place and writes them at their absolute object offsets via `pwritev`, coalescing offset-contiguous payloads into one positioned write. The write target sits behind a `SinkWrite` trait (`FileSink` for the file path), so the drain orchestration is exercised against an in-memory capture in tests and an alternative write strategy can replace the file write without touching the buffer or the drain logic.

### Public read-ahead control

`crate::types::ReadAhead { Auto, Parts(n) }` is exposed at three scopes, precedence dynamic > request > client:

- `Config::builder().read_ahead(..)` — client default.
- `download().read_ahead(..)` — per-request override.
- `handle.io_ctl().set_read_ahead(..)` — dynamic, on a running transfer.

`Parts(n)` caps read-ahead at `n` parts of speculation beyond the part the consumer is waiting on (`n + 1` resident), so `Parts(0)` is demand paging. `Auto` uses the fixed per-transfer cap. The knob resolves to a window in one place, `window_parts_for`, used at construction and by the dynamic setter. I/O controls live on a new `DownloadIoCtl`, separate from `SchedulingCtl` (data movement vs scheduling).

## How to review

Start with `recv_buffer.rs` — it is the core and stands alone. Its module doc walks the structure, the two surfaces, and the six memory-ordering/safety rules; the concurrency is model-checked (loom) and the sequencing is covered by deterministic unit tests. Read it before the wiring.

Then `transfer.rs::poll_work` — the occupancy gate (`issued - released < window`) and the completion repoke. The gate condition and the `released` accounting are what fix the wedge.

Then `body.rs` — the `SinkWrite`/`FileSink` seam and the drain orchestration (run coalescing, offset translation). `read_ahead.rs` is the window controller and the knob-to-window resolution; `types.rs`/`input.rs`/`builders.rs`/`config.rs`/`handle.rs` are the public `ReadAhead` surface and are mechanical.

## Testing

`recv_buffer` is covered at three levels: deterministic unit tests for the sequencing logic (exhaustive fill-order permutations, head-of-line blocking, the segment hop, drain-run partition and gaps, terminal partial-tail drain, exactly-once free, out-of-order eager reclaim, and the block-vs-stream counter split); loom models for the concurrency edges (the fill/poll handoff, the hop, reclaim against an outstanding write and against a drop, multi-producer out-of-order fill, and drainers racing and partitioning a segment); and miri (stacked borrows) over the unit suite. A concurrent stress test drives fill-then-drain at batch 1 and asserts every slot drains exactly once, guarding the fill-probe's saturating subtraction against a regression.

Download integration tests against the mock S3 server cover the behaviors this change targets: a disk download exceeding the read-ahead window (the wedge), a disk download with a window below the segment size (the sub-segment drain regime), and a slow-consumer transfer that must pace rather than wedge.

## Changes

```
aws-sdk-s3-transfer-manager/
  src/
    operation/download/
      recv_buffer.rs    NEW — paged receive buffer, stream + block surfaces, drain primitive
      read_ahead.rs     NEW — occupancy window controller, ReadAhead knob resolution
      body.rs           slot ring → PagedRecvBuffer; SinkWrite/FileSink seam; positioned writes
      transfer.rs       occupancy gate in poll_work; completion repoke
      handle.rs         DownloadIoCtl (runtime read-ahead control)
      input.rs          per-request read_ahead override
      builders.rs       read_ahead fluent setter
      context.rs        Transferring.issued (issuance cursor)
      download.rs        wiring
    types.rs            ReadAhead { Auto, Parts(n) }
    config.rs           client-default read_ahead
    runtime/sync/{std,loom}.rs   atomics used by the buffer under the loom shim
  examples/cp.rs        jemalloc receive-path config for the download benchmark
integration-tests/
  src/download.rs       disk-exceeds-window, sub-segment-window, slow-consumer tests
```

## Notes for review

- The occupancy window is a fixed per-transfer cap. Cross-transfer arbitration against a global memory budget, and an `Auto`-mode resident-size adaptation, layer on top of the `AtomicU64` window and are not part of this change.
- The disk drain executes its positioned write inline on the execute path; moving the blocking file operations (temp-file create, rename) off the caller's runtime is tracked separately and is orthogonal to this change.

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/operation/download/body.rs
+++ b/aws-sdk-s3-transfer-manager/src/operation/download/body.rs
@@ -35,20 +35,55 @@ impl WakeNotify {
     async fn notified(&self) {}
 }
 
-/// File sink for download-to-file writes.
+/// Positioned-write target for download-to-file. Abstracts the file so the drain
+/// orchestration (run coalescing, offset translation) can be exercised against an
+/// in-memory capture, and so an alternative write strategy (e.g. O_DIRECT/io_uring)
+/// can replace the file write without touching the buffer or the drain logic.
 ///
-/// Positioned writes land at `chunk.offset - object_range_start` in the file.
-struct Sink {
+/// `write_all_at` writes the whole buffer at an absolute file position; `preallocate`
+/// is a best-effort size hint. Implementations are shared across the issuer and the
+/// drain task, hence `Send + Sync`.
+pub(crate) trait SinkWrite: Send + Sync + std::fmt::Debug {
+    /// Write the entire buffer at `pos` bytes into the target.
+    fn write_all_at(
+        &self,
+        buf: &mut bytes_utils::SegmentedBuf<bytes::Bytes>,
+        pos: u64,
+    ) -> std::io::Result<()>;
+
+    /// Best-effort preallocation of `len` bytes. Default no-op.
+    fn preallocate(&self, _len: u64) {}
+}
+
+/// File-backed [`SinkWrite`]: positioned writes via `pwritev`.
+struct FileSink {
     file: std::fs::File,
-    /// Start of the S3 byte range for this transfer.
-    object_range_start: u64,
-    /// Whether the transfer manager created this file (vs caller-provided).
+    /// Whether the transfer manager created this file (vs caller-provided). Only an
+    /// owned file is preallocated.
     owns_file: bool,
 }
 
-impl std::fmt::Debug for Sink {
+impl std::fmt::Debug for FileSink {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("Sink").finish_non_exhaustive()
+        f.debug_struct("FileSink").finish_non_exhaustive()
+    }
+}
+
+impl SinkWrite for FileSink {
+    fn write_all_at(
+        &self,
+        buf: &mut bytes_utils::SegmentedBuf<bytes::Bytes>,
+        pos: u64,
+    ) -> std::io::Result<()> {
+        crate::io::fs::write_all_at(&self.file, buf, pos)
+    }
+
+    fn preallocate(&self, len: u64) {
+        if self.owns_file {
+            if let Err(e) = crate::io::fs::preallocate(&self.file, len) {
+                tracing::warn!(error = %e, "failed to preallocate file space");
+            }
+        }
     }
 }
 
@@ -56,7 +91,12 @@ impl std::fmt::Debug for Sink {
 #[derive(Debug)]
 enum Mode {
     Stream,
-    Disk(Sink),
+    /// Positioned writes land at `chunk.offset - object_range_start` in the sink.
+    Disk {
+        sink: Box<dyn SinkWrite>,
+        /// Start of the S3 byte range for this transfer.
+        object_range_start: u64,
+    },
 }
 
 /// Producer handle to the download body buffer. Held at `self.inner.writer`
@@ -121,7 +161,7 @@ impl Drop for BodySlot {
 
 impl BodyWriter {
     pub(crate) fn has_sink(&self) -> bool {
-        matches!(&*self.mode, Mode::Disk(_))
+        matches!(&*self.mode, Mode::Disk { .. })
     }
 
     /// Claim the next slot. The read-ahead gate lives in poll_work (tracks
@@ -141,32 +181,53 @@ impl BodyWriter {
         self.buffer.released()
     }
 
-    /// Disk mode: drain every sealed (FULL) segment to disk. Called on the
-    /// `SealedSegment` edge from execute tasks. Stream mode: no-op.
-    pub(crate) fn drain_sealed(&self) -> Result<(), std::io::Error> {
-        if let Mode::Disk(sink) = &*self.mode {
-            while let Some(sw) = self.buffer.take_completed_segment() {
-                write_segment(sink, &sw)?;
+    /// Disk mode: drain every batch-ready run to disk. Called on the `DrainReady`
+    /// edge from execute tasks. A non-terminal drain claims only runs that reach the
+    /// drain batch, coalescing each into one positioned write. Stream mode: no-op.
+    pub(crate) fn drain(&self, terminal: bool) -> Result<(), std::io::Error> {
+        if let Mode::Disk {
+            sink,
+            object_range_start,
+        } = &*self.mode
+        {
+            while let Some(sw) = self.buffer.take_drain_run(terminal) {
+                write_run(sink.as_ref(), *object_range_start, &sw)?;
                 sw.complete();
             }
         }
         Ok(())
     }
 
-    /// Terminal drain: sealed segments then the partial tail. Called once
-    /// from `complete()` / `on_terminal()`.
+    /// Terminal drain: flush every remaining filled run, including a partial final
+    /// segment below the drain batch. Called once from `complete()` / `on_terminal()`.
     pub(crate) fn finalize(&self) -> Result<(), std::io::Error> {
-        if let Mode::Disk(sink) = &*self.mode {
-            while let Some(sw) = self.buffer.take_completed_segment() {
-                write_segment(sink, &sw)?;
-                sw.complete();
-            }
-            while let Some(sw) = self.buffer.take_tail() {
-                write_segment(sink, &sw)?;
-                sw.complete();
-            }
+        if !matches!(&*self.mode, Mode::Disk { .. }) {
+            return Ok(());
         }
+        // Report the terminal flush: how many parts this last pass wrote (those left
+        // resident below the drain batch) and the total drained over the transfer.
+        // Fires once per disk transfer — the lifecycle marker that the tail actually
+        // reached disk; its absence on a hung transfer localizes the stall to the
+        // terminal drain.
+        let before = self.released();
+        self.drain(true)?;
+        let total = self.released();
+        tracing::debug!(
+            target: crate::telemetry::TARGET_TRANSFER,
+            terminal_parts = total - before,
+            parts_total = total,
+            "terminal drain complete; tail flushed to disk",
+        );
         Ok(())
+    }
+
+    /// Couple the drain batch to the read-ahead window: a window narrower than a
+    /// segment drains in smaller runs so the block surface never waits for a run it
+    /// cannot accumulate. Capped at the segment size (the coalescing target). Called
+    /// at construction and whenever the window changes.
+    pub(crate) fn sync_drain_batch(&self, window: u64) {
+        let batch = std::cmp::min(SEG_SIZE as u64, window.max(1)) as usize;
+        self.buffer.set_drain_batch(batch);
     }
 
     /// Wake the consumer (terminal transition).
@@ -176,20 +237,22 @@ impl BodyWriter {
 
     /// Best-effort pre-allocation of disk space.
     pub(crate) fn preallocate(&self, len: u64) {
-        if let Mode::Disk(sink) = &*self.mode {
-            if sink.owns_file {
-                if let Err(e) = crate::io::fs::preallocate(&sink.file, len) {
-                    tracing::warn!(error = %e, "failed to preallocate file space");
-                }
-            }
+        if let Mode::Disk { sink, .. } = &*self.mode {
+            sink.preallocate(len);
         }
     }
 }
 
-/// Gather a segment's filled payloads and write them to disk at the
-/// correct file positions. Reads payloads in place via `Slot::get()`;
-/// `complete()` (called by the caller after this returns) frees them.
-fn write_segment(sink: &Sink, sw: &SegmentWrite<ChunkOutput>) -> std::io::Result<()> {
+/// Gather a claimed run's filled payloads and write them to the sink at the correct
+/// positions, coalescing offset-contiguous payloads into one positioned write. A
+/// payload at object offset `o` lands at sink position `o - object_range_start`. Reads
+/// payloads in place via `Slot::get()`; `complete()` (called by the caller after this
+/// returns) frees them.
+fn write_run(
+    sink: &dyn SinkWrite,
+    object_range_start: u64,
+    sw: &SegmentWrite<ChunkOutput>,
+) -> std::io::Result<()> {
     let mut run_start_file_pos: Option<u64> = None;
     let mut run_end: u64 = 0;
     let mut combined = bytes_utils::SegmentedBuf::<bytes::Bytes>::new();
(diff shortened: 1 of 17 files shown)


aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
```diff
@@ -189,15 +217,42 @@ impl DownloadTransfer {
...
+                        // poll in steady state.
+                        tracing::trace!(
+                            target: crate::telemetry::TARGET_TRANSFER,
+                            issued = *issued,
+                            released,
+                            in_flight = *issued - released,
+                            window,
+                            "read-ahead gate closed: issuance paused until the consumer drains",
+                        );
+                        self.inner.ctx.set_pending();
```
landonxjames, line 278: This gate reads `released()` and calls `set_pending()` while holding the state lock. The wake flag protocol documented in `transfer.rs` requires the mutator to serialize against that with the same lock (`lock -> mutate -> unlock -> try_wake`). The disk path follow this pattern (`decrement_in_flight` mutates under the lock and then wakes). The stream path does not (`Body::next -> poll_next` advances `released` and calls `try_wake` with no state lock) (see `body.rs:445` / `recv_buffer.rs:906`).

The broken interleaving pattern looks like `poll_work` reads stale `released`. The consumer frees the last resident part and calls `try_wake`, which sees `pending = false` and no-ops, `poll_work` then `set_pending` and parks. Since `try_wake` only calls `scheduler.wake` when pending was already set, the schedulers release and recheck can't recover it, nothing repolls. Reachable when the gate closes with zero in-flight parts (fast network filled the whole window) and the consumer drains occupancy to zero inside this critical section. Deleting the estimator's per poll tick removed the fallback that used to mask this. The existing loom models don't cover the gate, so I included some tests below: `t1`, a model of the gate vs. a concurrent stream consumer using the production memory orderings, fails with `LOST WAKE: poller parked with the gate open and no repoll scheduled`.

Possible Fix: The gate and consumer are the store buffer litmus pattern (poller writes `pending`, reads `released`, consumer writes `released`, reads `pending`), so re-reading `released` after `set_pending` can still return stale, the loom model `t2` (double-check) fails the same way. The fix is to supply the missing happens before. Route the consumer's occupancy-release and `try_wake` through the same lock the gate checks under (the `lock -> mutate -> unlock -> try_wake` pattern in loom model `t3` passes), or make the four store-buffer ops SeqCst.

Tests to prove this out (mostly AI generated because I've never used loom before and was confusing myself trying to do it by hand):

```sh
LOOM_MAX_PREEMPTIONS=3 RUSTFLAGS="--cfg s3_tm_loom" \
  cargo test --lib --release gate_lost_wake_loom -- --test-threads=1
```

```
t1_current_logic_loses_wake ............ FAILED  ("LOST WAKE: poller parked ...")
t2_double_check_alone_still_loses_wake . FAILED  ("... re-read of `released` was stale")
t3_shared_lock_protocol_is_safe ........ ok
```

```rust
/// Loom model of the read-ahead gate (`poll_work`) racing a concurrent stream
/// consumer (`Body::next → RecvBufferConsumer::poll_next → try_wake`), using the
/// production primitives and memory orderings:
///   - `released` : buffer occupancy counter (fetch_add AcqRel / load Acquire).
///   - `pending`  : the `wake_flag` (store true Release = set_pending, swap false
///                  AcqRel = take_pending).
///   - `woken`    : whether try_wake scheduled a re-poll — production calls
///                  `scheduler.wake` ONLY when take_pending() returns true.
/// issued/window = 1 so one consumer release opens the gate; zero in-flight parts,
/// so `decrement_in_flight` gives no backup wake — the wedge conditions.
#[cfg(all(test, s3_tm_loom))]
mod gate_lost_wake_loom {
    use crate::runtime::sync::sync::atomic::{AtomicBool, AtomicU64, Ordering};
    use crate::runtime::sync::sync::{Arc, Mutex};
    use crate::runtime::sync::thread;

    const ISSUED: u64 = 1;
    const WINDOW: u64 = 1;

    /// t1 — current logic: gate arms `set_pending` with no re-check; consumer
    /// advances `released` + `try_wake` with no shared lock. EXPECTED TO FAIL.
    #[test]
    fn t1_current_logic_loses_wake() {
        loom::model(|| {
            let released = Arc::new(AtomicU64::new(0));
            let pending = Arc::new(AtomicBool::new(false));
            let woken = Arc::new(AtomicBool::new(false));

            let (r_c, p_c, w_c) = (released.clone(), pending.clone(), woken.clone());
            let consumer = thread::spawn(move || {
                r_c.fetch_add(1, Ordering::AcqRel); // poll_next: released++
                if p_c.swap(false, Ordering::AcqRel) {
                    w_c.store(true, Ordering::Release); // take_pending() → scheduler.wake
                }
            });

            let r = released.load(Ordering::Acquire); // P1
            let parked = if ISSUED - r >= WINDOW {
                pending.store(true, Ordering::Release); // P2: set_pending, park
                true
            } else {
                false
            };
            consumer.join().unwrap();

            let gate_open = ISSUED - released.load(Ordering::Acquire) < WINDOW;
            if parked && gate_open {
                assert!(
                    woken.load(Ordering::Acquire),
                    "LOST WAKE: poller parked with the gate open and no re-poll scheduled"
                );
            }
        });
    }

    /// t2 — plain double-check after arming is NOT enough: nothing orders the
    /// re-read of `released` against the consumer's fetch_add, so it can still be
    /// stale. EXPECTED TO FAIL (disproves "double-check is the minimal fix").
    #[test]
    fn t2_double_check_alone_still_loses_wake() {
        loom::model(|| {
            let released = Arc::new(AtomicU64::new(0));
            let pending = Arc::new(AtomicBool::new(false));
            let woken = Arc::new(AtomicBool::new(false));

            let (r_c, p_c, w_c) = (released.clone(), pending.clone(), woken.clone());
            let consumer = thread::spawn(move || {
                r_c.fetch_add(1, Ordering::AcqRel);
                if p_c.swap(false, Ordering::AcqRel) {
                    w_c.store(true, Ordering::Release);
                }
            });

            let r = released.load(Ordering::Acquire); // P1
            let parked = if ISSUED - r >= WINDOW {
                pending.store(true, Ordering::Release); // P2: arm
                let r2 = released.load(Ordering::Acquire); // double-check (can still be stale)
                if ISSUED - r2 < WINDOW {
                    pending.swap(false, Ordering::AcqRel);
                    false
                } else {
                    true
                }
            } else {
                false
            };
            consumer.join().unwrap();

            let gate_open = ISSUED - released.load(Ordering::Acquire) < WINDOW;
            if parked && gate_open {
                assert!(
                    woken.load(Ordering::Acquire),
                    "LOST WAKE despite double-check: the re-read of `released` was stale"
                );
            }
        });
    }

    /// t3 — the fix: consumer's mutate + wake go through the same lock the gate
    /// checks under (`lock → mutate → unlock → try_wake`), supplying the missing
    /// happens-before. EXPECTED TO PASS.
    #[test]
    fn t3_shared_lock_protocol_is_safe() {
        loom::model(|| {
            let lock = Arc::new(Mutex::new(()));
            let released = Arc::new(AtomicU64::new(0));
            let pending = Arc::new(AtomicBool::new(false));
            let woken = Arc::new(AtomicBool::new(false));

            let (l_c, r_c, p_c, w_c) =
                (lock.clone(), released.clone(), pending.clone(), woken.clone());
            let consumer = thread::spawn(move || {
                {
                    let _g = l_c.lock(); // shim Mutex: lock() returns the guard
                    r_c.fetch_add(1, Ordering::AcqRel); // mutate under the lock
                }
                if p_c.swap(false, Ordering::AcqRel) {
                    w_c.store(true, Ordering::Release); // try_wake after unlock
                }
            });

            let parked = {
                let _g = lock.lock();
                let r = released.load(Ordering::Acquire);
                if ISSUED - r >= WINDOW {
                    pending.store(true, Ordering::Release);
                    true
                } else {
                    false
                }
            };
            consumer.join().unwrap();

            let gate_open = ISSUED - released.load(Ordering::Acquire) < WINDOW;
            if parked && gate_open {
                assert!(
                    woken.load(Ordering::Acquire),
                    "LOST WAKE under the lock protocol (should be unreachable)"
                );
            }
        });
    }
}
```

aws-sdk-s3-transfer-manager/src/operation/download/body.rs
```diff
@@ -587,30 +436,41 @@ impl Body {
...
-                // Wake the transfer so it can generate more work now that a slot freed up
-                transfer.ctx().try_wake();
-                Some(Ok(chunk))
+        loop {
+            // Register interest BEFORE checking state (lost-wake safety).
+            let notified = self.notify.notified();
+            if let Some(chunk) = self.consumer.poll_next() {
+                // Delivery advances `consumed`, freeing read-ahead occupancy; wake
+                // the issuer so the reopened window admits more work.
+                self.transfer.ctx().try_wake();
```
landonxjames: This is the release side of the race from `transfer.rs`. `poll_next` advanced `released` and here we `try_wake`, both outside the state lock that `poll_work` holds while reading `released` and calling `set_pending`. Because the two sides don't share a lock, the consumer's `released++ / try_wake` can slip between the gate's `released` read and `set_pending`, losing the wake. Route this wake through the same lock pattern the disk path uses (`decrement_in_flight` takes the state lock before `try_wake`). See the loom model in `transfer.rs` comment (`t2` fails, `t3` with the lock protocol passes).

aws-sdk-s3-transfer-manager/src/operation/download/recv_buffer.rs
```diff
@@ -0,0 +1,2375 @@
...
+//! and the per-slot `FILLED` state, not their relative order, gates a safe read; the
+//! fill probe's `filled - claim_cursor` is therefore a saturating subtraction, never a
+//! bare one.
+//!
+//! # Surface exclusion
+//!
+//! An instance is driven through one consumption surface or the other, never both: the
+//! stream cursor empties slots as it passes; the block surface reads runs in place and
+//! frees them. A block-only caller drops the `RecvBufferConsumer`. This exclusion is a
+//! caller obligation, not type-enforced — `PagedRecvBuffer` is `Clone` and exposes the
```
landonxjames, line 145: This feels like a dangerous footgun that would be easy for a future contributor to trigger. Is there any way we could enforce this at the type level? Maybe a token that is consumed when either action is performed so we can't start the other (now dangerous) operation since it seems like we would only do either one once?

landonxjames (approved):
LGTM. One interleaving bug and one comment about the safety of an API, bug might be blocking.

aws-sdk-s3-transfer-manager/src/operation/download/recv_buffer.rs
aajtodd: Yeah we can look at it I had already sort of pulled at this a bit but didn't finish it. In part because you can't reach it from public API. I'll take another look

aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
aajtodd: Good catch!

aajtodd: Fixed in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/153/changes/2814443963187dbd8a5f41bdf1249b4c294ce006

PR #154, opened 2026-07-09
vnext: Memory Budget

aajtodd:
## Summary

Adds a global memory budget that bounds total resident transfer data across all concurrent transfers, and wires the download path onto it. Prior flow control bounded memory only per-transfer, through each download's read-ahead window; N concurrent transfers could each fill a window, so aggregate resident memory scaled with concurrency and had no ceiling. The budget is a fungible, chunk-accounted admission gate: a transfer reserves before it buffers, and at the limit transfers backpressure — park and resume — rather than fail or over-allocate. It sizes itself from detected machine RAM by default and is observable through the client.

## Why

Read-ahead (#153) gave each download an occupancy window: issuance runs ahead of consumption up to `window` resident parts, then waits. That bounds one transfer. It does not bound a fleet of them — `download_objects` runs thousands of child downloads, and nothing capped their combined footprint. The window is also per-transfer by construction: it cannot express "all transfers together stay under X bytes," because no single transfer knows what the others hold.

A global bound has to be fungible (any transfer draws from one shared pool, since demand is unpredictable and shifts between transfers), part-size-agnostic (parts vary in size within and across transfers), and it has to backpressure rather than fail — a download that cannot reserve should wait for memory to free, not error. It also has to compose with the existing per-transfer window without the two gates fighting.

### The budget primitive

`MemoryBudget` (`runtime/memory.rs`) tracks fixed-size chunks logically reserved across all transfers behind one lock. A request for `n` bytes costs `ceil(n / chunk)` chunks (`chunk` = 8 MiB nominal), so non-uniform part sizes account correctly without the budget knowing any transfer's part size.

```text
reserve(bytes) ─┬─ fits & queue empty ──▶ Reserve::Ready(Reservation)   ── RAII; drop returns chunks
                └─ else ────────────────▶ Reserve::Pending(WaitTicket)  ── parked FIFO; notify on grant

Reservation::drop ──▶ return chunks ──▶ drain waiters front-to-back (strict FIFO, no skip-ahead)
```

- `reserve(bytes, notify)` grants immediately (`Ready`) when the wait queue is empty and the request fits; otherwise it enqueues a waiter and returns `Pending(WaitTicket)`, invoking `notify` when a later release grants it. `try_reserve(bytes)` is the allocation-free fast path: it grants under exactly the same condition but registers no waker and never enqueues, for a caller that can re-drive itself.
- `Reservation` is an RAII ticket; its drop returns the chunks and drains the wait queue. `WaitTicket` is the parked-request handle; `take()` yields the granted reservation once the notify fires, and dropping it cancels a still-queued waiter or releases an already-granted one.
- **Arrival-order admission.** A fresh request never jumps a queued waiter even if it would fit, and the drain stops at the first front waiter that does not fit rather than skipping past it. Together these guarantee a queued request of `need <= capacity` sees `in_use` only fall until its need is met — it cannot be starved by a stream of smaller requests. The cost is bounded head-of-line blocking under mixed part sizes; uniform part sizes never trigger it.
- **Idle-only forced grant.** A request larger than the entire capacity is granted only when `in_use == 0`. Strict FIFO guarantees the front waiter eventually reaches that point, so an oversized request (an un-sliced upload; downloads slice to part size and never hit it) makes progress instead of parking forever.

The one non-obvious invariant is lock ordering. Two lock flavors exist — the budget lock and per-waiter slot locks — and the release path (`budget → slot`, to store a grant) and `WaitTicket::drop` (`slot → budget`, to cancel or return) must never nest, or they deadlock. `WaitTicket::drop` extracts the slot state under the slot lock, releases it, then acts. This is model-checked under loom.

### Sizing and configuration

The `Auto` default sizes the budget from detected physical RAM — cgroup-aware on Linux, `sysctl` on macOS, `GlobalMemoryStatusEx` on Windows — as `round_pow2(0.25 * RAM)` clamped to `[512 MiB, 32 GiB]`. The clamp auto-tapers the effective fraction on large boxes and keeps a small box usable; an undetectable environment falls back to a fixed default. `MemoryBudgetConfig::{Fraction, Limit}` override the proportion or set an explicit ceiling. `Config::builder().memory_budget(..)` sets it; `Client::memory_budget()` returns a `MemoryBudgetSnapshot` (capacity, reserved bytes, live waiters, cumulative park count) for observability.

### Wiring the download path

A download reserves against the budget as it issues work, and the reservation rides the buffered bytes until they leave memory:

- **Two gates, window first.** Read-ahead issuance takes the min of two gates: the per-transfer occupancy window (`gate.try_issue`), then the budget (`reserve`). The window is checked first so a window-blocked transfer never holds a slice of the fungible budget it cannot yet use, which would starve other transfers.
- **Reservation lifetime = byte residency.** A reservation is taken when a slot is claimed, moves into the `ChunkOutput` at `fill`, and drops when the bytes it accounts for leave memory: on the stream surface when the consumer drops the delivered chunk, on the disk surface when the drain copies the bytes out. It is held as `Arc<Reservation>` so a cloned `ChunkOutput` (sharing the same backing bytes) is charged once and released on the last clone.
- **Budget-blocked read-ahead parks without failing.** When the budget queues a read-ahead claim, the transfer holds the claimed slot in `DownloadState::Transferring.pending` (a `PendingClaim`) and returns `PollWork::Pending`; the budget's notify re-readies it. The gate's `issued` bump stays, so the parked transfer holds exactly the slot it will fill.
- **Terminal completion runs in `execute`, not `poll_work`.** This is the one non-mechanical restructure in the diff, and it exists so budget release is never gated on a re-poll. A chunk's reservation releases when its bytes drain to disk; a chunk below the drain batch drains only at completion. If completion runs in `poll_work`, reclaiming that budget needs a scheduler re-poll — but under a saturated budget the transfers holding budget are the ones needing the re-poll, and the re-poll is gated by the scheduler capacity (`dispatched < target`) they hold (discovery parks inside `execute`, keeping its slot). That is a circular wait, and it hung the `256 KiB × 10k` concurrent disk benchmark. So `decrement_in_flight` detects terminal completion under the state lock and claims the `Terminal` transition there (a concurrently-woken `poll_work` cannot double-complete), and `execute` finalizes inline — tail drain plus reservation release through the normal drop path, symmetric with `fail()`. `poll_work`'s completion arm now handles only the zero-byte object.
- **`enter_terminal` is the sole path to `Terminal`.** It extracts any budget-parked `PendingClaim` under the state lock so the `WaitTicket` drop (which takes the budget lock) happens only after the state lock is released, keeping the budget lock from nesting under the state lock. `#[must_use]`, so the claim cannot be silently dropped in place.

## How to review

Start with **`runtime/memory.rs`** — the primitive stands alone. Its module doc, the lock-ordering note above `Inner`, and `can_grant` / `drain` / `WaitTicket::drop` are the core; the concurrency is model-checked (loom) and the arrival-order/forced-grant/cancel semantics are covered by unit tests. Read it before the wiring.

Then **`operation/download/transfer.rs`** — `reserve_claim` (the two-gate issuance and the `try_reserve` fast path), `reserve_chunk` (the discovery reserve), and the terminal-completion rework: `decrement_in_flight` returning whether it claimed terminal, and `finalize_completion` doing the inline tail drain. The gate ordering and the terminal claim are where correctness lives.

Then **`operation/download/context.rs`** (`PendingClaim`, `enter_terminal`) and **`operation/download/body.rs`** (the reservation moving `BodySlot → ChunkOutput` at `fill`, and dropping with the delivered bytes).

**`runtime/platform.rs`** (RAM detection), **`types.rs`** / **`config.rs`** / **`client.rs`** (the `MemoryBudgetConfig` surface and `Client::memory_budget()`) are mechanical.

## Testing

The primitive's unit tests (`memory.rs`) cover the admission semantics — chunk rounding, ready/pending grants, drop-returns-chunks, exhaustion-then-release, strict arrival order, the idle-only forced grant, waiter cancellation (including a middle waiter), and soft shrink. Two loom models check the lock-ordering invariant: `release_races_cancel` and `release_races_take` race the release path (`budget → slot`) against `WaitTicket::drop` (`slot → budget`); a nested-lock regression would deadlock and loom would report it.

The wiring tests (`transfer.rs`, `body.rs`) cover the download-path integration: a read-ahead claim parking on a tight budget and resuming on release, both terminal paths (cancel, fail) dropping a budget-parked claim without leaking a chunk or a queued waiter, budget release happening in the terminal `execute` with no subsequent `poll_work` (pinning the deadlock fix), and the reservation riding a `ChunkOutput` — charged once across clones, released when the bytes leave memory on both the stream and disk surfaces.

Verified on this branch: 507 lib tests pass; `fmt` and `clippy` (default features) clean; both loom models pass under `--cfg s3_tm_loom`. On the standard s3fio benchmark suite the two originally-failing download workloads (the `256 KiB × 10k` concurrent hang and the 32 GiB disk transfer) now pass.

## Changes

```
aws-sdk-s3-transfer-manager/
  src/
    runtime/
      memory.rs         NEW — MemoryBudget primitive: reserve/try_reserve, FIFO drain,
                              WaitTicket, RAII Reservation; unit + loom tests
      platform.rs       NEW — physical-RAM detection (Linux cgroup / macOS / Windows), budget sizing
      mod.rs            module wiring
    operation/download/
      transfer.rs       two-gate issuance; reserve_claim/reserve_chunk; terminal completion
                              moved into execute (decrement_in_flight, finalize_completion)
      context.rs        PendingClaim; enter_terminal (sole Terminal transition)
      body.rs           reservation rides BodySlot → ChunkOutput; releases with the bytes
    types.rs            MemoryBudgetConfig { Auto, Fraction, Limit }, MemoryBudgetSnapshot
    config.rs           client-default memory_budget setter
    client.rs           Handle.memory_budget; Client::memory_budget() snapshot
    telemetry.rs        budget reserve/grant/release folded onto the scheduling target
    examples/cp.rs      surface the budget config in the example
```

## Notes for review

- Budget observability rides the existing `scheduling` telemetry target (`aws_sdk_s3_transfer_manager::scheduling`): reserve/grant/release at `trace`, the saturation edge where a reserve parks at `debug`. A budget that can block a transfer must be observable in the field. It does not get a standalone target yet — the budget is a global admission gate, same class as scheduler capacity, and a separate env var is one more thing to forget and silently lose telemetry on; a dedicated target can come when a buffer pool lands and earns richer memory telemetry.
- `set_limit` (runtime resize) and the connection-cap detection in `platform.rs` carry `allow(dead_code)`: the resize has no adaptive-sizing driver yet and the connection cap has no connection-pool consumer yet. Both are tracked; the driver work is separate from this change.
- The budget bounds admitted memory (a ceiling on what transfers may buffer), not RSS.
- The budget is a global primitive; download is its first consumer. Upload bounds its in-flight bytes only through scheduler concurrency today, so a fleet of `upload_objects` children has the same unbounded-aggregate exposure this closes for downloads. Wiring upload onto the same budget is the follow-up — the primitive already carries the piece upload needs (the idle-only forced grant handles an un-sliced or oversized part), so it is consumer wiring, not a primitive change.

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
+++ b/aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
@@ -62,7 +62,7 @@ macro_rules! bail_if_terminal {
 
 /// Download transfer that generates and executes download work.
 ///
-/// Cheap to clone - all state is behind `Arc`.
+/// Clone shares all state via `Arc`.
 #[derive(Debug, Clone)]
 pub(crate) struct DownloadTransfer {
     inner: Arc<DownloadTransferInner>,
@@ -738,7 +738,8 @@ impl DownloadTransfer {
             ..Default::default()
         });
 
-        if self.decrement_in_flight(freed) {
+        let reached_terminal = self.decrement_in_flight(freed);
+        if reached_terminal {
             return self.finalize_completion();
         }
 
@@ -863,7 +864,8 @@ impl DownloadTransfer {
             ..Default::default()
         });
 
-        if self.decrement_in_flight(freed) {
+        let reached_terminal = self.decrement_in_flight(freed);
+        if reached_terminal {
             return self.finalize_completion();
         }
 
--- a/aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
+++ b/aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
@@ -473,7 +473,7 @@ impl DownloadTransfer {
                 let terminal_wake = self.inner.discovery_notify.notified();
                 if let Some(reservation) = ticket.take() {
                     tracing::debug!(
-                        target: crate::telemetry::TARGET_MEMORY,
+                        target: crate::telemetry::TARGET_SCHEDULING,
                         tid = %self.inner.ctx.id,
                         len,
                         "discovery chunk reservation granted; resuming",
--- a/aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
+++ b/aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
@@ -304,7 +304,7 @@ impl DownloadTransfer {
                         return self.park();
                     }
 
-                    // Gate admitted (and counted) the slot. Claim it from the ring and
+                    // Gate admitted (and counted) the slot. Claim it from the buffer and
                     // reserve its backing memory against the budget. A grant issues now;
                     // a queued reservation stashes the claimed slot in `pending` and
                     // parks until the budget wakes us. The gate's `issued` bump stays —
--- a/aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
+++ b/aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
@@ -18,12 +18,13 @@ use crate::error::{self, ChunkRef, Error};
 use crate::io::AggregatedBytes;
 use crate::operation::download::body::{BodySlot, BodyWriter, ChunkOutput};
 use crate::operation::download::chunk_meta::ChunkMetadata;
-use crate::operation::download::context::DownloadState;
+use crate::operation::download::context::{DownloadState, PendingClaim};
 use crate::operation::download::discovery::{discover_obj, ObjectDiscovery};
 use crate::operation::download::object_meta::ObjectMetadata;
 use crate::operation::download::read_ahead::ReadAhead;
 use crate::operation::download::recv_buffer::FillOutcome;
 use crate::operation::download::DownloadInput;
+use crate::runtime::memory::{NotifyFn, Reservation, Reserve};
 use crate::transfer::{IoRequest, PollWork, Transfer, TransferContext, TransferId, WorkOutcome};
 use crate::types::BucketType;
 
@@ -38,12 +39,22 @@ pub(crate) enum DownloadWork {
     },
 }
 
+/// Bytes in the next chunk sliced from `remaining` at `part_size`: the reservation
+/// size for that chunk. Peek only — does not advance `remaining`.
+fn chunk_len(remaining: &std::ops::RangeInclusive<u64>, part_size: u64) -> usize {
+    let start = *remaining.start();
+    let end = *remaining.end();
+    let chunk_end = cmp::min(start + part_size - 1, end);
+    (chunk_end - start + 1) as usize
+}
+
 /// Early return if transfer is terminal (failed/cancelled by another work item).
 macro_rules! bail_if_terminal {
     ($self:expr) => {
         if !$self.inner.ctx.is_active() {
-            // Bailing before any drain: no occupancy freed on this path.
-            $self.decrement_in_flight(0);
+            // Bailing before any drain: no occupancy freed on this path. The transfer is
+            // already terminal so `decrement_in_flight` will find `Terminal` and return false.
+            let _ = $self.decrement_in_flight(0);
             return WorkOutcome::Cancelled;
         }
     };
@@ -246,16 +257,31 @@ impl DownloadTransfer {
                 etag,
                 part_size,
                 gate,
+                pending,
             } => {
-                if let Some(range) = remaining.as_ref() {
-                    // Apply the read-ahead gate before issuing. The gate bounds resident
-                    // occupancy at `issued - released < window`, where `released` counts
-                    // both delivery surfaces (stream pull and disk drain), so a disk
-                    // download paces to its drain rate rather than to the in-order
-                    // delivery cursor. `try_issue` reads and mutates the gate under this
-                    // state lock; the consumer's `release` does the same, so the two are
-                    // ordered and a release that reopens the gate cannot be lost against
-                    // the `set_pending` below (the mutator protocol).
+                // Resolve the slot to issue into this poll, if one is ready. Two gates
+                // compose in order — the per-transfer read-ahead window, then the global
+                // memory budget — so issuance takes their min. The window is gated first
+                // (via `gate.try_issue`) so a window-blocked transfer never holds a slice
+                // of the fungible budget it cannot use yet, starving other transfers.
+                let slot = if pending.is_some() {
+                    // A budget-parked claim takes priority: the gate already admitted and
+                    // counted its slot (held in `pending`, not re-gated), waiting only on
+                    // the reservation the budget queued.
+                    match self.resume_pending_claim(pending) {
+                        Some(slot) => slot,
+                        // Not granted yet; the queued ticket is the waker.
+                        None => return self.park(),
+                    }
+                } else if let Some(range) = remaining.as_ref() {
+                    // Read-ahead gate. The gate bounds resident occupancy at
+                    // `issued - released < window`, where `released` counts both delivery
+                    // surfaces (stream pull and disk drain), so a disk download paces to
+                    // its drain rate rather than to the in-order delivery cursor.
+                    // `try_issue` reads and mutates the gate under this state lock; the
+                    // consumer's `release` does the same, so the two are ordered and a
+                    // release that reopens the gate cannot be lost against the
+                    // `set_pending` below (the mutator protocol).
                     //
                     // The gate guards issuance only. Once every range is generated
                     // (`remaining` is None) the completion path below runs unconditionally,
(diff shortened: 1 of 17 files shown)


aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
```diff
@@ -275,59 +301,198 @@ impl DownloadTransfer {
...
+            claim.slot
+        })
+    }
+
+    /// Claim a slot and reserve its backing memory against the global budget. The
+    /// read-ahead gate has already admitted (and counted) this slot. Returns the slot
+    /// with its reservation attached on an immediate grant; on a queued reservation,
+    /// stashes the claimed slot in `pending` and returns `None` so the caller parks
+    /// until the budget wakes it.
+    fn reserve_claim(
```
landonxjames, line 431: Think there might be a deadlock here.

A disk chunk's reservation is released only when the drain copies its bytes out (`ChunkOutput.reservation` drop, body.rs:422) or at the terminal `finalize()`. But a nonterminal drain only claims a run at the drain batch or a full segment (`take_drain_run`, recv_buffer.rs:757), and the default batch is `min(SEG_SIZE=16, window=1024) = 16` (body.rs:241). So a multi-part object smaller than one 16 part segment (2..15 parts = 8..128 MiB at the 8 MiB default) never drains nonterminally, it holds every part until it reaches terminal.

When `reserve_claim` parks (the `Reserve::Pending` arm at line 429), `poll_work` stashes the claim in `pending` and returns `Pending`. A subsequent poll takes the `resume_pending_claim -> ticket.take() -> None -> park()` path (lines 270-274) and never reaches the completion/terminal branch, `remaining` stays `Some`, so `decrement_in_flight` never sees `remaining.is_none() && ranges_in_flight == 0` (line 913). So `finalize` and any release is unreachable for a budget-parked transfer.

Now spread the shared budget across enough concurrent disk transfers that none can reserve its full part count: none finalizes -> none drains -> none releases -> global stall

Pretty sure this is reachable with defaults. Budget floor is 512 MiB = 64 chunks. The wedge occurs once roughly `budget_chunks / parts_per_object` concurrent multi-part disk transfers are live, ex: ~5 concurrent 100 MiB (13 part) downloads. `download_objects` fans out to many children to local files, so it is in range. A second pressure source: `reserve_chunk` (transfer.rs:462) backpressures discovery by awaiting inside `execute`, holding a worker/concurrency slot while parked, so budget-parked discoveries also reduce the parallelism available to make progress.

The two cited passing benchmarks don't exercise this: `256 KiB × 10k` is single-part-per-object (terminal after one part), and `32 GiB` is a single large object that drains per full segment (floor 64 chunks > batch 16). 

Don't think stream downloads are impacted. The consumer drives release independent of the drain batch, so a stream transfer that consumes in order always frees budget. The bug is specific to the disk surface, where release is drain batch gated.

Repro test:

```sh
cargo test -p aws-sdk-s3-transfer-manager --lib \
  budget_deadlock_concurrent_multipart_disk_below_drain_batch
```

```
test ... budget_deadlock_concurrent_multipart_disk_below_drain_batch ... ok
```

```rust
/// REVIEW REPRO: fungible-budget deadlock for concurrent multi-part
/// disk transfers whose parts stay below the drain batch.
///
/// A disk chunk's reservation releases only when the drain copies it out, and a
/// non-terminal drain fires only at the drain batch (`min(SEG_SIZE=16, window)`)
/// or a full segment. A multi-part object smaller than a segment (2..15 parts)
/// therefore holds ALL its parts resident until the *terminal* finalize — which
/// requires issuing every part, which requires budget. When the shared budget is
/// spread across concurrent such transfers so none can reach its own part count,
/// none finalizes, none drains, and none releases: a global stall.
///
/// Scaled to two 2-part transfers on a 2-chunk budget. Each object alone needs 2
/// chunks and the budget is 2, so either would complete on its own (hold both
/// parts -> terminal -> finalize -> release); together they wedge. Same shape at
/// the default 64-chunk budget once >~ budget/parts_per_object concurrent
/// multi-part disk objects are live (the `download_objects` workload this PR
/// targets).
#[cfg_attr(miri, ignore)]
#[tokio::test]
async fn budget_deadlock_concurrent_multipart_disk_below_drain_batch() {
    use crate::runtime::memory::BUDGET_CHUNK_BYTES;

    let part_size = BUDGET_CHUNK_BYTES as u64; // 1 part == 1 chunk
    let object_size = 2 * part_size; // 2 parts, below the 16-part drain batch

    // One shared handle => one shared budget across both transfers.
    let mk_client = || {
        let chunk = vec![0u8; part_size as usize];
        let get_obj = mock!(aws_sdk_s3::Client::get_object).then_output(move || {
            GetObjectOutput::builder()
                .content_length(part_size as i64)
                .content_range(format!("bytes 0-{}/{}", part_size - 1, object_size))
                .e_tag("test-etag")
                .body(ByteStream::from(chunk.clone()))
                .build()
        });
        mock_client!(aws_sdk_s3, RuleMode::MatchAny, &[get_obj])
    };
    let config = crate::Config::builder()
        .client(mk_client())
        .part_size(crate::types::PartSize::Target(part_size))
        .build();
    let handle = crate::client::Handle::test_handle_tokio(config);
    let budget = handle.memory_budget.clone();
    // Tight shared budget: exactly two chunks (one per transfer's discovery chunk).
    budget.set_limit(2 * BUDGET_CHUNK_BYTES);

    // Build a disk-mode transfer on the shared handle. Keep the tempdir alive.
    let build_disk = |handle: Arc<crate::client::Handle>| {
        let input = DownloadInput::builder()
            .bucket("test-bucket")
            .key("test-key")
            .build()
            .unwrap();
        let dir = tempfile::tempdir().unwrap();
        let file = std::fs::File::create(dir.path().join("out")).unwrap();
        let (writer, consumer) =
            crate::operation::download::body::new_recv_body_with_sink(file, 0, false);
        let (ctx, rx) = TransferContext::new(handle);
        let t = DownloadTransfer::new(ctx, BucketType::Standard, input, writer);
        (t, consumer, dir, rx)
    };
    let (a, _ca, _da, _ra) = build_disk(handle.clone());
    let (b, _cb, _db, _rb) = build_disk(handle.clone());

    // Discovery: each reserves + fills its seq-0 chunk. seq-0 is 1 part, below the
    // 16-part drain batch, so it stays RESIDENT on disk (undrained). in_use == 2.
    skip_discovery(&a).await;
    skip_discovery(&b).await;
    assert_eq!(
        budget.in_use_chunks(),
        2,
        "both discovery chunks reserved and resident (undrained: below the drain batch)"
    );

    // Each transfer tries to issue its part-1. The budget is full, so both park,
    // each holding a PendingClaim (the read-ahead window is wide open — only the
    // budget binds here).
    assert_pending(a.poll_work());
    assert_pending(b.poll_work());
    assert_eq!(
        budget.stats().waiters,
        2,
        "both part-1 reservations are queued on the budget"
    );

    // Exercise EVERY production path that could release the budget:
    //  (1) Non-terminal disk drain — the only drain a running transfer performs.
    //      seq-0 is a single filled part: below the drain batch (16) and not a full
    //      segment, so `take_drain_run(false)` claims nothing and frees nothing.
    assert_eq!(
        a.writer().drain(false).unwrap(),
        0,
        "seq-0 is below the drain batch: no non-terminal drain, no release"
    );
    assert_eq!(b.writer().drain(false).unwrap(), 0);
    //  (2) Re-poll — still parked; the budget is still full.
    assert_pending(a.poll_work());
    assert_pending(b.poll_work());

    // DEADLOCK: the budget is permanently pinned by two resident-but-undrainable
    // seq-0 chunks. Neither transfer can issue part-1 (budget full), drain seq-0
    // (below batch), nor finalize (not terminal — `remaining` is still `Some`).
    // No wake will ever fire (a budget grant needs a release; the only release is
    // a terminal finalize neither transfer can reach). Nothing re-polls them.
    assert_eq!(
        budget.in_use_chunks(),
        2,
        "budget permanently held; no reachable path releases a chunk => global stall"
    );
    assert_eq!(budget.stats().waiters, 2, "both transfers wedged forever");

    // Proof the wedge is exactly the undrained resident chunks: a TERMINAL drain
    // (which production can only run at completion — which neither transfer can
    // reach) frees A's seq-0, and the freed chunk is re-granted FIFO to A's queued
    // part-1. This is the escape the running state machine has no path to.
    assert_eq!(
        a.writer().finalize().unwrap(),
        1,
        "only a terminal drain frees the resident chunk"
    );
    assert_eq!(
        budget.stats().waiters,
        1,
        "the freed chunk was re-granted to a queued waiter (proving the pin)"
    );
}
```

Possible fixes:
- Drain before park on the disk surface path: when a transfer is about to park on
  the budget with resident but undrained parts, flush its current partial run (a
  smaller positioned write) to release those chunks rather than parking while
  holding them.
- Or admission with a per-active-disk-transfer batch-sized floor, so a disk
  transfer is only granted its first chunk if it can eventually accumulate a
  full drain batch (trades some concurrency for liveness).
- Or couple the effective drain batch to budget headroom, so a budget pressured
  transfer drains sub-batch runs (accepting smaller writes when memory is scarce, 
  the same tradeoff `sync_drain_batch` already makes for small windows).

Fix should include a test with N concurrent medium multi-part
objects to disk under a budget `< N × parts_per_object`, asserting completion.

aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
```diff
@@ -692,35 +888,65 @@ impl DownloadTransfer {
...
+            match &mut *work {
+                DownloadState::Transferring {
+                    ranges_in_flight,
+                    gate,
+                    remaining,
+                    ..
+                } => {
+                    *ranges_in_flight = ranges_in_flight.saturating_sub(1);
+                    gate.release(freed);
+                    if remaining.is_none() && *ranges_in_flight == 0 {
```
landonxjames, line 973: A `PendingClaim` is only ever set in `poll_work`'s `remaining.as_ref()` arm, so `pending.is_some()` implies that `remaining.is_some()`. In this `remaining.is_none()` branch the extracted claim is always `None`, and the `drop(pending)` below will be a no-op. That all seems correct, but probably worth a `debug_assert!(pending.is_none())` and a comment to make the invariant explicit and catch a future regression that lets a claim survive into the terminal path (where its held slot/ticket would leak).

landonxjames (approved):
Looks good, one potential deadlock issue that should probably be fixed

aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
aajtodd: Right again. 🤦‍♂️

aws-sdk-s3-transfer-manager/src/runtime/memory.rs
```diff
@@ -0,0 +1,942 @@
...
+            WaitState::Granted(res) => {
+                // Granted but never taken: drop the reservation to return chunks.
+                // Safe: slot lock is released, so Reservation::drop can lock the budget.
+                drop(res);
+            }
+            WaitState::Waiting => {
+                // Cancel: remove this waiter from the queue.
+                let mut inner = self.budget.inner.lock();
+                inner.waiters.retain(|w| !Arc::ptr_eq(&w.slot, &self.slot));
+            }
```
vcjana, line 411: I might be missing a wake path here, but if the front waiter is cancelled and the next waiter now fits, does it stay parked until some unrelated reservation releases? 

Can we make the fix be to call drain after removing the cancelled waiter, then fire the collected notifies
after dropping the budget lock?

aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
aajtodd: https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/154/changes/eede5e2c21dbba589ab24133c5818a8cc4ff0087

aws-sdk-s3-transfer-manager/src/runtime/memory.rs
aajtodd: https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/154/changes/f947ba111e2719eae739499dedfbc09ee5265d18

PR #155, opened 2026-07-11
retry: recover from throttle storms instead of aborting

aajtodd:
## Summary

Reworks the transfer manager's retry and adaptive-latency subsystem so high-fan-out and large-object workloads survive transient failure instead of aborting. Five coordinated pieces:

1. **A per-reason retry loop over pure classifiers** — a fast backoff for transient transport failures, a hard backoff for throttling, sharing one attempt budget.
2. **A per-bucket, time-refilling retry token bucket** cached on the client handle, so a drained bucket recovers rather than aborting the transfer.
3. **The adaptive deadline reframed as a hedged request** (in the sense of *The Tail at Scale*): a slow request past its expected latency is cancelled and re-issued once (a hedge), bounded by a success-replenished budget rather than a timeout. This replaces the per-call `P99 × 2` histogram, and the earlier control-variable model's tier-widening and value-based escape.
4. **The download deadline bounded to time-to-first-byte**, with the body read untimed-but-retryable; the size-blind upload deadline removed.
5. **Transient-transport classification that walks the error source chain**, so a nested/re-wrapped connector error is still recognized and retried.

A load-driven mock throttle and a fault-injecting connector reproduce every failure deterministically, with no real S3.

## Why

**A throttle storm aborted the small-object benchmark workloads.** A `256KiB × 10k` concurrent transfer fans out far more requests than a prefix absorbs, and S3 sheds the excess with `503 SlowDown`. The SDK retries a throttle against a token bucket shared region-wide by default (`s3-{region}`) with no time-based refill: once the opening wave drains it, later throttles surface un-retried, the transfer manager classified a throttle as terminal, and the default `FailedTransferPolicy::Abort` failed the whole transfer on one child.

**The `P99 × 2` deadline was size-blind, on both ends.** One histogram per direction served every request size. For a 256 KiB object P99 is single-digit milliseconds, so `P99 × 2` landed below the healthy spread under load and cancelled merely-slow requests — burning the retry budget and tearing down connections. For a large object on a slow link the opposite held: a healthy-but-slow read exceeded `P99 × 2` and was cancelled as a straggler.

**A slow request is a hedge candidate, not a timeout.** Treating the deadline as a hard timeout means picking a give-up rule, and every value-based rule we tried (tier-widening the deadline, escaping past a threshold) either false-cancelled under concurrency or could not give up when a whole population was slow. Reframing it as a *speculative re-issue* — cancel a straggler, retry it once on a fresh connection, and stop speculating when the aggregate hedge rate says it is no longer paying off — gives the correct give-up behavior as an emergent property of a budget, not a tuned threshold.

**The download body path has no SDK retry, and the retry loop conflated policy with schedule.** A chunk's body is consumed *after* `GetObject` orchestration completes, so a mid-stream body failure is outside the SDK's retry scope — the transfer manager is its sole retrier. The prior loop fused deadline, classifier, and backoff, so a throttle could not take a different backoff from a transient IO error without duplicating the loop.

### The retry loop and per-reason backoff

`retry` runs each attempt by calling a caller-supplied `build()` for a fresh future, consults a pure classifier (`Fn(&GuardError<Error>) -> RetryDecision`), and owns the schedule itself — the classifier holds no loop and no RNG, the loop holds no domain knowledge. Up to `MAX_ATTEMPTS` (3). `RetryDecision` is `Retry` (fast transient backoff), `RetryThrottle` (hard throttle backoff), or `NoRetry`; the loop selects the `Backoff` by decision, so the two retry reasons share the attempt budget but not the schedule.

`Backoff` is a full-jitter truncated-exponential schedule, `t = b · min(initial · 2^i, max)`, `b` uniform in `[0, 1)`, `max = 5s`. It is pure and `Copy` — it owns no RNG; the caller supplies the draw (`fastrand::f64()` in production, a fixed value in tests), so the progression is directly assertable. The transient base is `100ms`; the throttle base is `1s` (10×, matching the SEP throttling backoff), because a transient body-IO failure is safe to re-issue fast whereas a service shedding load needs room to recover.

Three classifiers differ only in which errors they recognize. `classify_upload_part_retry` and `classify_discovery_retry` retry a transient-transport failure (`Retry`) and a throttle (`RetryThrottle`); everything else terminal. `classify_body_retry` adds the download body arms: a mid-stream `IOError` (reset, truncation, SSP stall) is `Retry`; a checksum mismatch (`IntegrityError`) is terminal — a corrupt body must never be re-fetched and masked.

### The adaptive deadline is a hedged request

`metrics/latency.rs` holds a per-client `DeadlineController`. The first `WARM_THRESHOLD` (10) successes are buffered, not timed; at the threshold the deadline is seeded at `max(p90, 1s)` over the buffered samples — a tail percentile, not the mean, so a high-variance link starts with a deadline that clears its slow tail instead of false-cancelling. If the arithmetic warmup mean already reaches the `5s` escape the link is uniformly too slow to time and no deadline is ever armed (`stop_timeout`, a permanent per-client latch). Thereafter each success ages the observed-latency mean (`α = 1/64`) and relaxes the deadline toward `mean + offset` (`α = 0.01`); a timed-out request never enters the mean, killing the censored-percentile ratchet.

Arming the deadline is a **hedge**: when a request exceeds it, the in-flight future is dropped (releasing its connection) and re-issued once on a fresh one. Two rules bound the speculation:

- **Hedge-once per operation.** The retry loop passes `allow_hedge = true` until the first deadline exceedance, then `false` for every later attempt, which run untimed to completion. So a single operation is speculatively cancelled at most once, and no operation can exhaust its attempt budget on the deadline. The one-time hedge does **not** consume a retry attempt — a genuine transport failure still gets the full `MAX_ATTEMPTS` whether or not the operation also hedged, so a speculative cancel never steals a transport retry.
- **A success-replenished hedge budget.** A token bucket on the controller (`capacity 10`): a success rewards `+0.05`, a fired hedge costs `−1`. The sustainable hedge fraction is `reward / (reward + cost) ≈ 4.8%`, matching *The Tail at Scale*'s ~5% speculative-load bound. Normal tail-hedging fits under it and the budget stays full; a broad slowdown drives hedges faster than successes refill, the budget drains to zero, and speculation halts. Give-up is emergent from the balance, not a value threshold — a uniformly-slow population self-limits.

Replenishing on success (not the clock) is what makes the budget dry up exactly when hedging stops paying off. The budget bounds the *sustained* hedge fraction, not an instantaneous count: every warm request arms (the budget is checked, not reserved), and a token is charged only when a hedge actually fires.

### A per-bucket, refilling token bucket cached on the handle

`bucket_retry_partition` keys the SDK's retry token bucket by bucket (`s3-tm-{bucket}`) instead of the region-wide default, so a storm on one bucket cannot drain the retry budget for every other bucket on the client — matching CRT, which partitions per S3 host. It gives the bucket a time-based refill (`refill_rate = 10`/sec) the SDK default lacks, so a drained bucket recovers a bounded retry budget over time; under a total outage the sustained retry rate is `refill_rate / throttling_retry_cost` ≈ 2/sec per bucket, independent of concurrency.

The partition is cached per bucket name on the `Handle` (`retry_partitions: Mutex<HashMap<String, RetryPartition>>`). This is load-bearing, not an optimization: the SDK clones a partition's embedded bucket rather than sharing across custom partitions, so a partition rebuilt per operation would carry a fresh, full bucket every time and the shared budget would never take effect.

### The download deadline bounds time-to-first-byte, not the body read

The chunk GET composes the deadline over only the send (`recv_latencies.guarded(allow_hedge, send)`), which returns after response headers arrive ≈ TTFB, then reads the body untimed. TTFB is near-zero everywhere and stalls are always mid-body dead connections, so timing the whole GET watched the wrong phase and cancelled large-but-healthy body reads. A slow TTFB still hedges and re-issues on a fresh connection; a dead mid-body stream is caught by stalled-stream protection, tightened to a `2s` grace (SSP fires only on zero throughput, so it cannot false-cancel a progressing stream).

Discovery is the entry point for every download; its three sends (get-first-part, head, ranged-get) are wrapped in the same retry loop (`classify_discovery_retry`), so a small single-part object — where discovery is the whole download — recovers from a throttle storm rather than aborting on the first un-retried 503. Discovery is not hedged (no deadline), only retried.

### The upload path carries no adaptive deadline

Timing the whole part-send (connect + body-push + response) is size-blind, the same defect the download TTFB split cured, so the upload adaptive deadline is removed. Upload relies on the SDK connect timeout, stalled-stream protection for a mid-body stall, and the outer loop's transient-transport and throttle retries. One gap remains, documented in the module docs and an ignored test: a response that never arrives *after* the request body is fully sent is not bounded (SSP watches request-body throughput, already complete). The fix is an upstream send-complete-relative response-first-byte timeout; the SDK's `read_timeout` is request-initiation-relative and the wrong primitive. The seam is operation-agnostic so upload joins the hedge path once that primitive lands.

### Transient-transport classification walks the source chain

The http client tags a connector-level failure (connection reset, incomplete response, TLS timeout) `ConnectorError::Other(TransientError)`. A nested dispatch can re-wrap a fully-classified inner connector error inside an outer `ConnectorError::Other(None)` that carries no marker, and both the SDK's own classifier and the transfer manager previously inspected only that outer frame — misclassifying a retryable failure as terminal. `is_sdk_transient_transport` now walks the error source chain and treats the error as transient if any `ConnectorError` frame is an IO error, a timeout, or carries the `TransientError` marker; a bare `Other(None)` with no marker anywhere stays non-transient.

## How to review

Start with `retry.rs` — it stands alone: the `Backoff` schedule, the three `RetryDecision` classifiers, the `retry` loop with the policy/schedule split and the free-hedge attempt accounting, and the per-bucket partition builder. Then `client.rs` for the partition cache and why it must be cached. Then `metrics/latency.rs`, self-contained: the `DeadlineController` (p90 seed, `stop_timeout`, success relaxation), the hedge budget (`can_hedge`/`charge_hedge`/`hedge_deadline`/`guarded`), and its unit tests. The wiring sites are `download/transfer.rs` (TTFB-guarded send, untimed body), `download/discovery.rs` (retry-wrapped sends), and `upload/transfer.rs` (no deadline, with the `# Failure handling` module doc as the ownership statement). `error.rs` holds the chain-walking transient classifier and the bulk-abort message. The mock side is `s3-mock-server/src/throttle.rs` (the `RateThrottle`) plus handler wiring.

## Testing

`retry.rs` unit tests cover the loop (retry-then-succeed, exhaustion at `MAX_ATTEMPTS`, terminal-inner-error), hedge-once (`allow_hedge` cleared after the first exceedance, non-deadline failures don't clear it, and a hedge does not consume a transport retry), the backoff schedule under seeded jitter, each classifier arm, and the throttle code set. `client.rs` asserts the partition is cached once per bucket and reused. `metrics/latency.rs` pins the control variable against its constants (p90 seed, the arithmetic-mean `stop_timeout` gate that a single cold sample must not trip, success relaxation, that a timeout never moves the mean) and the hedge budget (starts full and caps, sustained hedging drains it despite interleaved successes, a drained budget runs untimed). `error.rs` pins the chain-walking classifier, including the exact double-wrapped `Other(TransientError)`-under-`Other(None)` shape.

Integration tests reproduce every failure deterministically without real S3. `upload_retry.rs` drives a fault-injecting connector for a client-side IO dispatch failure the mock's server-side faults cannot model (isolated transient, a correlated 64-part burst that starves the SDK bucket and recovers via the outer retry, a persistent outage that fails without hanging). `upload_objects.rs` covers the throttle storm both ways for uploads *and* downloads: a persistent storm aborts with the failing object named; a load-driven storm under the mock's `RateThrottle` recovers because backoff paces re-issues under the rate limit — the download variant fails without the discovery retry wrap, so it guards that fix. `download_retry.rs` covers the body-read retry path and the TTFB split. Real-S3 e2e (`integrity.rs`) validates a 6 GiB multipart download on general-purpose and Express buckets; the two are serialized and free their source after upload to bound runner temp-disk usage.

## Changes

```
aws-sdk-s3-transfer-manager/
  src/
    retry.rs                       RetryDecision + retry loop (policy/schedule split,
                                   hedge-once, free-hedge attempt accounting);
                                   per-reason Backoff; classify_{upload_part,body,
                                   discovery}_retry; is_throttle; per-bucket partition
    client.rs                      Handle retry-partition cache + override accessors
    metrics/latency.rs             hedged-request DeadlineController: p90 seed,
                                   stop_timeout latch, success relaxation, and the
                                   success-replenished hedge budget (replaces P99 and
                                   the tier-widening/escape control model)
    error.rs                       chain-walking transient-transport classifier;
                                   ChildOperationFailed names the first failing object
    telemetry.rs                   drop the send-side latency tracker (upload untimed)
    io/stream.rs                   into_sdk_body doc: native body keeps inline checksum
    operation/upload/transfer.rs   execute_{upload_part,put_object}: outer retry, no
                                   deadline; # Failure handling ownership doc
    operation/upload/handle.rs     abort-MPU uses the cached per-bucket partition
    operation/download/transfer.rs TTFB-guarded send, untimed body; parsed content-range
                                   validation; hedge-aware guarded() call
    operation/download/discovery.rs   discovery sends wrapped in the retry loop
    operation/{upload,download}_objects/transfer.rs  abort-site source() note
s3-mock-server/
  src/
    throttle.rs                    NEW — load-driven RateThrottle token bucket
    s3s.rs, server.rs, lib.rs      check_throttle in every handler; set_rate_throttle;
                                   upload-path service-error + request-body-stall faults
    faults.rs                      StallRequestRead fault; PaceBody unifies slow/stall
integration-tests/
  src/
    upload_retry.rs, upload_retry/fault_connector.rs   NEW — connector IO-fault storm
    upload_objects.rs              throttle-storm recovery/abort (upload + download)
    download_retry.rs              body-read retry + TTFB-split coverage
    integrity.rs                   large-object e2e: serialized, source freed after upload
    harness.rs                     e2e failures debuggable from their own logs
.github/workflows/ci.yml           cargo-deny --config before the check subcommand
```

## Known limitations

- **Status-only throttles.** `is_throttle` matches on the parsed service code; a throttle that arrives as an HTTP 429/503 without a recognized code (e.g. some `HeadObject` 503s) is not classified here and falls through to the SDK's inner retry only.
- **Upload response-wait gap.** No deadline bounds a response that never arrives after the upload request body is fully sent (see above); it awaits an upstream send-complete-relative timeout primitive.
- **Nested transient re-wrap at source.** The classifier now recovers a re-wrapped transient connector error, but the SDK's *own* inner retry still inspects only the outer frame, so it does not retry this class on its own; eliminating the double-wrap at its source in the managed-runtime dispatch is a follow-up.


aws-sdk-s3-transfer-manager/src/operation/download/discovery.rs
landonxjames, line 159: Think this misses retries for small single part objects?

The upload part send and the download body read paths were wrapped in `retry(classify_*_retry, …)`, which retries a throttle with the hard 1s backoff. But the three discovery methods: `discover_obj_with_get_first_part`, `discover_obj_with_head`, and `discover_obj_with_get` each do a single `.send().await` with only `bucket_partition_override` and no `retry` loop:

```rust
        .customize()
        .config_override(transfer.ctx().handle.bucket_partition_override(input.bucket()))
        .send()   // <- no retry(classify_body_retry, …) wrapping this
        .await
        .map_err(error::Error::from)?;
```

For a small single part object, `get_first_part` is the entire download (no separate range GET follows), so a `503 SlowDown` on discovery that outlasts the SDK's bounded inner retry is terminal and the default `FailedTransferPolicy::Abort` fails the whole transfer. 

The download aborts with `ChildOperationFailed`, `failed_downloads` carries `ServiceError { code: "SlowDown" }` on `GetObject`, the un-recovered discovery throttle. The PR's download tests only pin the two ends (`discovery_single_service_error_recovers` = one 503 recovers via SDK retry; `discovery_service_error_surfaces_underlying_cause` = persistent throttle fails); there is no relenting-storm recovery test for download, so the gap is invisible in CI.

Potential Fix: wrap the discovery sends in the same throttle retry loop the body/upload paths use (the symmetric fix, discovery is the entry point for every download), or just add a doc comment that this is a limitation.

Repro test in `integration-tests/src/upload_objects.rs`

```sh
cargo test -p integration-tests transient_throttle_storm_recovers -- --test-threads=1
```

```
test upload_objects::download_objects_transient_throttle_storm_recovers ... FAILED  (3/3)
test upload_objects::upload_objects_transient_throttle_storm_recovers   ... ok      (3/3)
```

```rust
/// DOWNLOAD counterpart to `upload_objects_transient_throttle_storm_recovers`.
/// Discovery (the sole request for a small single-part object) is NOT wrapped in
/// the TM `retry` loop, so a small-object download recovers from a throttle storm
/// only via the SDK's inner retry — strictly weaker than the upload path. Drives
/// the SAME storm the upload recovery test survives.
#[tokio::test]
async fn download_objects_transient_throttle_storm_recovers() {
    timeout(TEST_TIMEOUT, async {
        let (server, server_handle, tm) = setup_with_concurrency(16).await;
        let count = 200usize;
        let size = 4 * ByteUnit::Kibibyte.as_bytes_usize();
        let bucket = "test-bucket";
        let prefix = "recover-dl/";

        // Seed the store (no throttle installed yet).
        let dataset = make_flat_dataset(count, size);
        tm.upload_objects().bucket(bucket).source(dataset.path())
            .walker(FsWalker::builder().recursive(true).build())
            .key_prefix(prefix).initiate().expect("initiate seed upload")
            .join().await.expect("seed upload must succeed");

        // Now install the load-driven storm: every GET is rate-limited server-wide.
        server.set_rate_throttle(50.0, 15.0);

        let dest = TempDir::new().expect("dest tempdir");
        let handle = tm.download_objects().bucket(bucket).key_prefix(prefix)
            .destination(dest.path()).initiate().expect("initiate download_objects");

        let output = handle.join().await
            .expect("a load-driven throttle storm on download must recover (parity with upload)");
        assert_eq!(output.objects_downloaded(), count as u64);
        assert!(output.failed_transfers().is_empty());
        server_handle.shutdown().await.expect("shutdown");
    })
    .await
    .expect("download_objects_transient_throttle_storm_recovers timed out");
}
```

landonxjames (approved):
LGTM, one potential issue with downloads of small objects, but I'm not really sure if this PR was meant to cover that or not

aws-sdk-s3-transfer-manager/src/operation/download/transfer.rs
```diff
@@ -845,36 +872,53 @@ impl DownloadTransfer {
...
+                .customize()
+                .config_override(ctx.handle.download_get_override(input.bucket()));
+            async move {
+                // Timed (TTFB): obtain response headers. Validate the range here
+                // so a mismatch is classified before we commit to the body read.
+                let rh_validate = rh.clone();
+                let resp = recv_latencies
+                    .guarded(async move {
+                        let resp = req.send().await.map_err(crate::error::Error::from)?;
+                        validate_content_range(&rh_validate, resp.content_range())?;
```
vcjana, line 906: Could we compare parsed byte ranges here instead of relying on the helper's substring match? `bytes 10-99/100` contains `0-9`, so a mismatched response could pass validation for `bytes=0-9`

aws-sdk-s3-transfer-manager/src/retry.rs
```diff
@@ -94,173 +178,469 @@ fn into_error(ge: GuardError<Error>) -> Error {
...
     }
 }
 
+/// Whether a transfer error is an S3 throttle (a service error whose code denotes
+/// throttling). Code-only: the flattened [`Error`] carries the service code but
+/// not the HTTP status, which is sufficient because S3 throttles carry a code
+/// (`SlowDown`). The set mirrors the scheduler's throttle classifier
+/// (`scheduler::concurrency`); a status-only throttle would not be caught here.
+fn is_throttle(e: &Error) -> bool {
+    matches!(
```
vcjana, line 254: Could `is_throttle` also check HTTP status 429/503? `HeadObject` 503s may not have a parsed `SlowDown` code, so code-only matching can miss them

vcjana (approved):

ysaito1001:
Similar to Landon's discovery gap, but the range-GET path (`execute_get_range`) is already wrapped in `retry(classify_body_retry, …)`? So the wrap alone might not cover it once the tracker is warm.
  
  `guarded` times the whole `req.send()`, and that send includes the SDK's inner `503` retry. Warm, the deadline is at its `EXPECTED_OFFSET_US` floor (~700ms which is under the SDK's 1s throttle backoff) so it fires *during* that backoff and drops the send before the `503` ever returns. The TM then sees a generic `DeadlineExceeded`, not the
  throttle, so `classify_body_retry` takes the fast `Retry` arm instead of `RetryThrottle`: the 1s pacing + refilling per-bucket bucket are bypassed, and the chunk ends as   an `IOError` with the `SlowDown` cause dropped (`service: None`).
  
Scope is large / multi-part downloads on a warm handle, and that seems complementary to Landon's single-part discovery path, not the same gap.

Running the following test
```
  cargo test -p integration-tests range_get_throttle_masked_by_armed_deadline -- --nocapture
```

```
// Range-GET throttle masked by the armed TTFB deadline (fails today; passes once                                                                                 
// the throttle surfaces as a `ServiceError`). Contract mirrors                                                                                                   
// `discovery_service_error_surfaces_underlying_cause`, on the range path.  
//
// `execute_get_range` times the whole `req.send()` in `guarded`, and that send                                                                                   
// includes the SDK's inner `503` retry. Once warm, the deadline is at its ~700ms                                                                                 
// `EXPECTED_OFFSET_US` floor — under the SDK's 1s throttle backoff — so it fires                                                                                 
// mid-backoff and drops the send before the `503` ever returns. The TM then sees a                                                                               
// generic `DeadlineExceeded`, not the throttle, so `classify_body_retry` takes the                                                                               
// fast `Retry` arm instead of `RetryThrottle`: the 1s pacing + refilling per-bucket                                                                              
// bucket are bypassed and the chunk ends as `IOError` with `SlowDown` dropped                                                                                    
// (`service: None`). 
//
// `max_attempts = 8` only removes SDK backoff jitter so the preemption is                                                                                        
// deterministic; the mechanism holds at the default (700ms floor < 1s base).   

  #[tokio::test]
  async fn range_get_throttle_masked_by_armed_deadline() {
      let t = Target::mock_gp()
          .connect_mock_configured(Some(PART_SIZE), |b| {
              b.retry_config(
                  aws_sdk_s3::config::retry::RetryConfig::standard()
                      .with_initial_backoff(std::time::Duration::from_secs(1))
                      .with_max_attempts(8),
              )
          })
          .await;
  
      // Warm the adaptive TTFB deadline: only range-chunk GETs record a sample, so
      // WARMUP_COUNT multipart objects (four range chunks each) push past
      // WARM_THRESHOLD (10). Fast in-memory chunks seed the deadline at its floor.
      let warmup_data = multipart_data();
      for i in 0..WARMUP_COUNT {
          let key = format!("warmup-{i}");
          t.put(
              &key,
              warmup_data.clone(),
              ChecksumStrategy::with_calculated_crc32(),
          )
          .await;
          t.download(&key, Some(ChecksumMode::Enabled))
              .await
              .expect("warm-up download must succeed");
      }
  
      // A single-range-chunk object; throttle only its RANGE GET (skip discovery),
      // so the outcome is attributable to the range path's deadline + classify.
      let data = single_range_chunk_data();
      t.put("throttled", data, ChecksumStrategy::with_calculated_crc32())
          .await;
  
      let mock = t.mock().expect("requires the mock backend");
      mock.insert_fault(
          t.bucket(),
          &t.key("throttled"),
          FaultType::ServiceError { status: 503 },
          SKIP_DISCOVERY,
          Occurrence::Always,
      );
  
      let err = tokio::time::timeout(std::time::Duration::from_secs(60), async {
          t.download("throttled", Some(ChecksumMode::Enabled)).await
      })
      .await
      .expect("timed out — the deadline-driven fast-retry loop should terminate quickly")
      .expect_err("a persistent 503 on the range GET must fail the download");
  
      assert_eq!(
          *err.kind(),
          aws_sdk_s3_transfer_manager::error::ErrorKind::ServiceError,
          "a range-GET throttle must surface as the underlying ServiceError (SlowDown), \
           but the armed TTFB deadline masked it — got: {err:?}",
      );
  
      t.shutdown().await;
  }
```
fails with
```
  assertion `left == right` failed: a range-GET throttle must surface as the underlying
  ServiceError (SlowDown), but the armed TTFB deadline masked it — got:
  Error { kind: IOError, source: "request exceeded the 1704ms latency deadline",
  extra: Some(ErrorExtra { service: None,
  location: Some(ChunkRef { seq: 1, range: Some(8388608..=12582911) }), ... }) }
    left: IOError
   right: ServiceError
```

Fix idea: keep the SDK's inner throttle retry out of the guarded send so a `503` surfaces as `ServiceError` → `RetryThrottle` before the deadline, otherwise just document it as the current behavior.

ysaito1001 (approved):

PR #157, opened 2026-07-14
concurrency: seed a fixed instance-aware target for Auto

aajtodd:
## Summary

`ConcurrencyMode::Auto` resolves to a fixed in-flight-request target derived from the machine's network bandwidth, instead of running the adaptive controller. The bandwidth estimate comes from the detected EC2 instance type (a per-family Gbps-per-vCPU table) times the local vCPU count; the target is `ceil(gbps / 0.4)`, clamped to `[10, 10000]`. Instance type is detected once — local DMI, falling back to IMDS — into a `MachineProfile` that also carries the vCPU count and RAM, so the memory budget sizes from the same detection. `TargetThroughput(t)` uses the same derivation from `t`; `Explicit(n)` is unchanged except `Explicit(0)` now clamps to 1 instead of tripping an assert. The adaptive controller stays in the tree, unwired.

## Why

**The adaptive controller's needs re-worked.** Adaptive controller is the right idea but the current implementation is not what I want to ship. It's too slow to ramp and was handling some issues better handled in the new connection pool. It also will benefit from better signals from the new connection pool work. So `Auto` ships a fixed seed for the next preview cut; the controller returns when the pool lands.

**Every option has to answer "what concurrency on an unknown machine," and without detection the answer is a blind constant.** CRT's client answers it from a throughput target the caller supplies (default 10 Gbps → 25 connections) and never looks at the instance; its instance-type detection feeds only query APIs the caller must invoke. Doing the detection in the engine seeds a large box correctly with no configuration.

### The estimate

`FAMILY_NIC_GBPS` records `(family, max_nic_gbps, vcpus_at_that_size)` for each instance family, read from published EC2 network specs — one row per family, checkable against the spec sheet. A family's bandwidth has no closed form (c6gn tops at 100 Gbps, c7gn at 200), so it's a lookup; within a family it scales linearly with vCPU, so `max_gbps / vcpus_at_that_size` is the per-vCPU rate for every size. `vcpus` is `std::thread::available_parallelism()`, which honors cgroup CPU quota and affinity, so a CPU-limited container seeds from its slice rather than the host. `GBPS_PER_CONN = 0.4` is CRT's per-connection figure (100 Gbps → 250). A family absent from the table, or a machine where detection failed, falls back to `4 × vcpus` — the ratio recognized network-optimized families already get, so an unknown box seeds near a known one rather than far below it.

The target is in-flight requests, not connections. In HTTP/1 an in-flight request holds a connection, so the connection bounds `[MIN_CONN, ABSOLUTE_MAX_CONN]` = `[10, 10000]` (CRT's) are the right bounds for the seed and are reused rather than duplicated.

### Detection

`MachineProfile { instance_type, vcpus, ram_bytes }` is resolved once, in `ConfigLoader::load`. Instance type comes from DMI (`/sys/devices/virtual/dmi/id/{sys_vendor,product_name}`); a readable non-EC2 vendor is definitive and skips IMDS, an unreadable one (a container, or non-Linux) falls back to a single IMDS lookup gated on `AWS_EC2_METADATA_DISABLED`. IMDS returns only the instance-type string, so the family table is needed either way. `Client::new` reads the profile and resolves both the concurrency seed and the memory budget from it. A `Config` built directly, without the loader, has no profile; `Client::new` then detects locally (DMI + vCPU + RAM, no IMDS), so construction stays non-blocking and a container without DMI still seeds by vCPU.

### Memory budget

`MemoryBudgetConfig::resolve` and the `machine_safe_mem` / `mem_for_fraction` sizers take RAM as an argument from the profile rather than reading it themselves, so RAM is detected once alongside the instance type. The sizing arithmetic is unchanged.

## How to review

`runtime/platform.rs` is the core: the `FAMILY_NIC_GBPS` table and `auto_concurrency_seed` / `seed_from_gbps` (the estimate and clamp), `classify_dmi` (the three-state DMI gate, pure and unit-tested), and `MachineProfile`. `config/loader.rs` composes DMI with the IMDS fallback; `client.rs` has `resolve_concurrency_target` and the single-profile resolution in `Client::new`.

## Testing

Unit tests recompute the estimate for recognized families, the vCPU fallback, container-sized vCPU counts, the clamp floor and ceiling, and NaN/zero/overflow inputs; `classify_dmi` is table-tested across its three states including the non-EC2 gate. `Client::new` is checked end-to-end (seed reaches `controller.target()`) and on the no-profile bypass path. The file- and network-reading detectors are validated on real EC2: the resolved seed was confirmed by log on `m6idn.large` (2 vCPU → 10, the floor), `c6in.16xlarge` (64 → 250), and `c7gn.16xlarge` (Graviton, 64 → 500). 536 lib tests, clippy clean, fmt clean.

## Changes

```
aws-sdk-s3-transfer-manager/
  src/
    runtime/platform.rs   FAMILY_NIC_GBPS; auto_concurrency_seed / seed_from_gbps;
                          MachineProfile + detect_local; classify_dmi +
                          detect_instance_type_dmi; machine_safe_mem /
                          mem_for_fraction take ram
    config/loader.rs      detect_machine_profile (DMI, IMDS fallback); populate profile
    config.rs             MachineProfile on Config + Builder
    client.rs             resolve_concurrency_target; single-profile resolution
    types.rs              MemoryBudgetConfig::resolve takes ram_bytes
    scheduler/concurrency.rs, scheduler/mod.rs   adaptive controller unwired, kept
```

## Notes for review

- `GBPS_PER_CONN = 0.4` and the `4 × vcpus` fallback are estimates, not measured for this client; the `FAMILY_NIC_GBPS` figures are verified against the current EC2 spec sheet.
- `Auto` becoming adaptive at GA is a behavior change under a stable enum, not an API change.
- On non-Linux there's no DMI path, so `Auto` always attempts one IMDS lookup at construction; deliberate, since EC2 runs Windows and macOS and IMDS is the only detection path there.
- The DMI reads use synchronous `std::fs`: the `/sys` attributes are kernel-resident (SMBIOS parsed at boot), read once at construction, so a `spawn_blocking` hop would cost more than the read.
```

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/config/loader.rs
+++ b/aws-sdk-s3-transfer-manager/src/config/loader.rs
@@ -121,6 +121,10 @@ impl ConfigLoader {
     pub async fn load(self) -> Config {
         let shared_config = aws_config::defaults(BehaviorVersion::latest()).load().await;
 
+        // Detect machine facts here, on the async load path, so client
+        // construction stays free of blocking DMI reads and network IMDS calls.
+        let profile = detect_machine_profile().await;
+
         let mut sdk_client_builder = aws_sdk_s3::config::Builder::from(&shared_config);
 
         let interceptor = S3TransferManagerInterceptor {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: aws-sdk-s3-transfer-manager/src/runtime/platform.rs [hunk 1]
```diff
@@ -402,6 +405,246 @@ fn walk_paths(mount: &str, cgroup_path: &str, filename: &str) -> Vec<PathBuf> {
...
+/// Resolve the `Auto` concurrency seed from detected machine facts.
+///
+/// Recognized family -> bandwidth estimate (`per_vcpu x vcpus`) -> connection
+/// derivation. Unknown/undetected -> `FALLBACK_INFLIGHT_PER_VCPU x vcpus`
+/// (a concurrency heuristic, not a bandwidth estimate). Both are clamped to
+/// `[MIN_CONN, ABSOLUTE_MAX_CONN]`. Pure function of its inputs.
+pub(crate) fn auto_concurrency_seed(instance_type: Option<&str>, vcpus: usize) -> usize {
+    match instance_type.and_then(family_gbps_per_vcpu) {
+        Some(per_vcpu) => seed_from_gbps(per_vcpu * vcpus as f64),
+        None => (FALLBACK_INFLIGHT_PER_VCPU * vcpus).clamp(MIN_CONN, ABSOLUTE_MAX_CONN),
```
vcjana: Could this fallback be too low for non-EC2 machines? For instance, 4 vCPUs gives a fixed concurrency of 16, lower than the old Auto start of 32.

aws-sdk-s3-transfer-manager/src/runtime/platform.rs
```diff
@@ -402,6 +405,246 @@ fn walk_paths(mount: &str, cgroup_path: &str, filename: &str) -> Vec<PathBuf> {
...
+    let family = instance_type.split('.').next()?;
+    FAMILY_NIC_GBPS
+        .iter()
+        .find(|(name, _, _)| name.eq_ignore_ascii_case(family))
+        .map(|(_, max_gbps, vcpus_at_max)| max_gbps / *vcpus_at_max as f64)
+}
+
+/// In-flight-request target from a throughput estimate: `ceil(gbps /
+/// GBPS_PER_CONN)`, clamped to `[MIN_CONN, ABSOLUTE_MAX_CONN]`.
+pub(crate) fn seed_from_gbps(gbps: f64) -> usize {
```
landonxjames, line 637: `resolve_concurrency_target` routes `TargetThroughput(t)` through this function, and if my math is right `GBPS_PER_CONN = 0.4` and `MIN_CONN = 10` limits the minimum throughput to ~4Gbps, which seems like it might be high if a machine is running outside of a datacenter.

aws-sdk-s3-transfer-manager/src/scheduler/concurrency.rs
```diff
@@ -3,10 +3,15 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+// The adaptive controller is retained but not wired: the dev preview resolves
```
landonxjames, line 6: I understand deferring this to until after the inital launch of v next, but I do worry about someone using this and causing a retry storm and taking something down. Possible to pull it forward before the next release or would it push it back too much?

landonxjames (approved):
LGTM, couple of questions no blockers

aws-sdk-s3-transfer-manager/src/scheduler/concurrency.rs
aajtodd: What specifically about a retry storm do you see different between this and adaptive? CRT uses roughly same logic for determining concurrency targets.

I would love to pull it forward but I think we're going to have to fast follow. I have another PR that reworks retries a bit.

aws-sdk-s3-transfer-manager/src/runtime/platform.rs
aajtodd: Right, the ~4 Gbps floor is `MIN_CONN (10)` × `GBPS_PER_CONN (0.4)`. For `Auto` that floor is deliberate: we're guessing at an unknown box, so seeding below ~10 is pointless. For an explicit `TargetThroughput` it's _maybe_ the wrong policy.

If you ask for 1 Gbps to cap concurrency, the floor ignores you. Fix is to let `TargetThroughput` honor low targets down to 1 while `Auto` keeps its floor. I'm going to leave this for now because it's mostly benign and likely to change with adaptive scheduler.

aws-sdk-s3-transfer-manager/src/runtime/platform.rs
aajtodd: Agreed it was a bit low. I reworked it in https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/157/changes/02269609c8a4f2d3e59f194a8c62327d476ec74e to use 32 as the floor and bumped to `5 * vcpu` when falling back.

aws-sdk-s3-transfer-manager/src/config/loader.rs
```diff
@@ -129,11 +133,104 @@ impl ConfigLoader {
...
+/// in microseconds, so one short attempt is ample.
+async fn imds_instance_type() -> Option<String> {
+    if imds_disabled() {
+        tracing::debug!(
+            target: crate::telemetry::TARGET_CONCURRENCY,
+            "IMDS disabled via AWS_EC2_METADATA_DISABLED; using vCPU-scaled fallback",
+        );
+        return None;
+    }
+    let client = aws_config::imds::Client::builder()
```
ysaito1001, line 213: LGTM. It's unfortunate that TM pulls in `aws-config` (increasing size & build time) to reach out IMDS client. There is a plan to modularize participants in default credentials provider chain, and should IMDS logic be moved to its own crate, TM could benefit from it, too.

PR #158, opened 2026-07-16
runtime: select the execution runtime via config

aajtodd:
## Summary

`ConcurrencyMode`-style config now selects the execution runtime: a `RuntimeMode` enum with `Managed` (default) and `CurrentTokio`, resolved once in `Client::new`. `Managed` spawns and owns worker threads as before; `CurrentTokio` runs transfer workers on the caller's existing tokio runtime via `tokio::spawn`, for embedding where the transfer manager must not spawn its own OS threads. The runtime trait, components, topology, and builders stay `pub(crate)` — the caller picks a mode and nothing else, so the runtime internals remain free to change. Alongside the selector this adds the integration coverage needed to validate a second runtime: `download_objects` (previously untested) gains a full suite, and the mock-based test files move onto a shared harness helper with dual-runtime coverage on the concurrency-exercising tests.

## Why

The transfer manager already owned two runtimes behind a private trait but only ever ran on `ManagedThreadRuntime` — `Client::new` hard-wired it, and `TokioMultiThreadRuntime` was reachable only from `#[cfg(test)]` constructors. A caller embedding the transfer manager inside an existing tokio application had no way to run transfers on that runtime instead of having the manager spawn its own threads. This exposes that choice as a narrow, opinionated config knob without exposing the runtime abstraction itself.

`CurrentTokio` is an escape hatch, not a performance tuning knob: `Managed`'s per-thread locality is faster by design, and on a network-bound download-discard workload (m6idn.16xlarge) the two are close — `Managed` leads by roughly 10-15%, concentrated on the single-stream many-parts path where a shared work queue contends. Prefer `Managed` unless you specifically need to run on your own runtime.

### The selector

`RuntimeMode` is a `#[non_exhaustive]` public enum in `types.rs`, mirroring `ConcurrencyMode`: a `runtime_mode()` setter on the config builder (and the `from_env` loader), defaulting to `Managed`. `Client::new` matches on it at the single construction site, building `ManagedThreadRuntime` or `TokioMultiThreadRuntime`. Everything the runtime is made of — the `ExecutionRuntime` trait, `RuntimeComponents`, `Topology`, the builders — stays `pub(crate)`, so the connection-pool, NUMA, and pinning work still landing in that layer does not become public surface.

`CurrentTokio` uses the SDK's default HTTP client (the managed path's per-thread clients do not apply when the manager does not own the threads); this is intentional for the preview.

### The current-thread guard

`CurrentTokio` runs workers via `tokio::spawn`, so on a current-thread runtime they serialize onto the one thread and throughput collapses with no other symptom — a measured 6 Gbps where a multi-threaded runtime reached 66. `Client::new` checks the ambient runtime flavor when `CurrentTokio` is selected and logs a warning if it is current-thread, turning that silent cliff into a visible diagnostic. The check is best-effort and does not change the constructor's signature.

### Test harness

`harness.rs` gains a lightweight mock path — `mock_tm(runtime)` and `mock_tm_with(runtime, configure)` returning `MockTm { server, handle, client }` — beside the existing `Target` machinery, which stays for the real-S3 / bucket-kind matrix. The mock-only test files (`download`, `upload`, `upload_objects`, `metrics`, and the new `download_objects`) previously hand-rolled the same in-memory-server-to-client setup; they now share `mock_tm`, and the runtime is a parameter so a test can run on either runtime from one shared body.

Dual-runtime tests follow the existing per-variant idiom (`_mock_gp` / `_real_gp` / `_real_express`): a shared `async fn body(RuntimeMode)` plus thin `#[tokio::test]` wrappers — `_mock_gp` on the default runtime and `_tokio_mt` on a multi-threaded runtime. These are explicit named functions, not macro-generated, so each variant runs individually from an IDE. Only the tests that exercise genuine concurrent dispatch get a `_tokio_mt` sibling; runtime-agnostic tests stay single.

### download_objects coverage

The composite directory-download path had no integration tests. This adds a suite mirroring `upload_objects`, inverted to seed the mock bucket and verify files on disk: many-small-files, content roundtrip, nested prefixes, multipart children, empty prefix, deep/wide tree, and the `Abort` failure policy. It also covers paths upload cannot exercise — ListObjectsV2 pagination, the `Continue` failure policy (one faulted key fails while the rest succeed and the failure names that key), and key path-traversal safety (keys with `../` components must fail rather than write outside the destination, confirming the `path_clean` + validation guard holds) — plus an upload-then-download roundtrip asserting the two directory trees are byte-identical.

## How to review

`types.rs` and `config.rs` are the public surface (the `RuntimeMode` enum and the `runtime_mode` setter); `client.rs` is the resolution and the current-thread guard. `harness.rs` holds the `mock_tm` helper and the two-path module doc. `download_objects.rs` is the new suite; the diffs in `download.rs` / `upload.rs` / `upload_objects.rs` / `metrics.rs` are a mechanical swap of their local setup onto `mock_tm` plus the selective `_tokio_mt` splits — no test logic changed.

## Testing

Beyond the integration suite (which now exercises `CurrentTokio` on both the download_objects anchors and the per-file concurrency anchors), `CurrentTokio` was validated on real EC2 (m6idn.16xlarge) across the standard benchmark suite: it completes every workload `Managed` does and holds line rate on the concurrent max-throughput cases. This is what backs shipping a runtime that until now ran only under `#[cfg(test)]`.

## Changes

```
aws-sdk-s3-transfer-manager/
  src/
    types.rs            RuntimeMode enum (public, non_exhaustive)
    config.rs           runtime_mode on Config + Builder
    config/loader.rs    runtime_mode on the from_env loader
    client.rs           resolve runtime by mode; current-thread guard
    runtime/mod.rs, runtime/tokio_mt.rs, runtime/tokio_mt/worker_pool.rs
                        TokioMultiThreadRuntime reachable from Client::new (dead_code gates dropped)
  examples/cp.rs        --runtime flag; runtime flavor matches the selected mode
integration-tests/src/
    harness.rs          mock_tm / mock_tm_with / MockTm (lightweight mock path)
    download_objects.rs new suite (parity + pagination, partial-failure, path-safety, roundtrip)
    download.rs, upload.rs, upload_objects.rs, metrics.rs
                        moved onto mock_tm; _tokio_mt siblings on concurrency anchors
    lib.rs              register download_objects
```

## Notes for review

- `Managed` remains the default and is unchanged; `Auto` behavior and every existing test path are untouched.
- `CurrentTokio` requires a multi-threaded runtime — the current-thread guard warns, it does not error, so a caller who knows what they are doing is not blocked.
- Running the mock-only files across both runtimes exhaustively (rather than the selected concurrency anchors) is deliberately left for a follow-up; the harness now supports it uniformly.
- The `TokioMultiThreadRuntime` work queue is a single shared queue; the ~10-15% gap against `Managed` on the single-stream path is queue contention, a known and separable optimization, not a correctness issue.

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/client.rs
+++ b/aws-sdk-s3-transfer-manager/src/client.rs
@@ -277,9 +277,26 @@ impl Client {
                     }
                     Arc::new(builder.build())
                 }
-                RuntimeMode::CurrentTokio => Arc::new(
-                    crate::runtime::TokioMultiThreadRuntime::new(weak_handle.clone()),
-                ),
+                RuntimeMode::CurrentTokio => {
+                    // CurrentTokio runs transfer workers on the caller's runtime via
+                    // `tokio::spawn`. On a current-thread runtime they all serialize onto
+                    // the one thread, collapsing throughput with no other symptom. Warn
+                    // loudly rather than let that happen silently.
+                    if matches!(
+                        tokio::runtime::Handle::try_current().map(|h| h.runtime_flavor()),
+                        Ok(tokio::runtime::RuntimeFlavor::CurrentThread)
+                    ) {
+                        tracing::warn!(
+                            target: crate::telemetry::TARGET_SCHEDULING,
+                            "RuntimeMode::CurrentTokio selected on a current-thread runtime; \
+                             transfer workers will serialize onto one thread. Use a \
+                             multi-threaded runtime or RuntimeMode::Managed."
+                        );
+                    }
+                    Arc::new(crate::runtime::TokioMultiThreadRuntime::new(
+                        weak_handle.clone(),
+                    ))
+                }
             };
 
             let s3_client = match config.take_s3_client_source() {
--- a/aws-sdk-s3-transfer-manager/src/client.rs
+++ b/aws-sdk-s3-transfer-manager/src/client.rs
@@ -138,30 +138,17 @@ impl Handle {
         self.transfer_override(bucket)
     }
 
-    /// Create a Handle for testing with a custom scheduler factory.
+    /// Create a Handle for testing on the ambient tokio runtime with a fixed
+    /// concurrency target. The ergonomic common case; tests that need a custom
+    /// runtime or a controller they drive directly use
+    /// [`new_for_test_with_runtime`](Self::new_for_test_with_runtime).
     #[cfg(test)]
-    pub(crate) fn new_for_test(mut config: crate::Config, concurrency: usize) -> Arc<Self> {
-        Arc::new_cyclic(|weak| {
-            let scheduler = Scheduler::new(weak.clone());
-            let runtime: Arc<dyn ExecutionRuntime> =
-                Arc::new(crate::runtime::TokioMultiThreadRuntime::new(weak.clone()));
-            let s3_client = match config.take_s3_client_source() {
-                crate::config::S3ClientSource::Provided(client) => client,
-                crate::config::S3ClientSource::FromConfig(s3_config) => {
-                    aws_sdk_s3::Client::from_conf(s3_config.builder.build())
-                }
-            };
-            Self {
-                config,
-                s3_client,
-                scheduler,
-                runtime,
-                controller: Arc::new(crate::scheduler::FixedConcurrency::new(concurrency)),
-                telemetry: Arc::new(Telemetry::new(std::time::Duration::from_millis(500))),
-                memory_budget: MemoryBudget::new(TEST_MEMORY_BUDGET_BYTES, BUDGET_CHUNK_BYTES),
-                retry_partitions: std::sync::Mutex::new(std::collections::HashMap::new()),
-            }
-        })
+    pub(crate) fn new_for_test(config: crate::Config, concurrency: usize) -> Arc<Self> {
+        Self::new_for_test_with_runtime(
+            config,
+            Arc::new(crate::scheduler::FixedConcurrency::new(concurrency)),
+            |weak| Arc::new(crate::runtime::TokioMultiThreadRuntime::new(weak)),
+        )
     }
 
     /// Test handle using the ambient tokio runtime (no OS threads spawned).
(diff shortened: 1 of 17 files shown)


aws-sdk-s3-transfer-manager/src/runtime/tokio_mt.rs
landonxjames: `ensure_worker_capacity` above already advances `worker_count` to `target` via `compare_exchange(current, target, ...)`, and then calls `spawn_workers(target - current)`, which **also** `fetch_add`s the delta here. So on a growth pass the counter lands at `target + (target - current)` instead of `target`, and the next growth check sees an inflated `current` and stops spawning too early and the pool under-provisions workers.

```rust
fn test_handle(concurrency: usize) -> Arc<crate::client::Handle> {
    let s3_client = aws_smithy_mocks::mock_client!(aws_sdk_s3, []);
    let config = crate::Config::builder().client(s3_client).build();
    crate::client::Handle::new_for_test(config, concurrency)
}

#[tokio::test(flavor = "multi_thread")]
async fn ensure_worker_capacity_double_counts_on_growth() {
    let handle = test_handle(8); // FixedConcurrency(8)
    let weak = Arc::downgrade(&handle);
    let rt = TokioMultiThreadRuntime::new(weak);

    // Pretend 4 workers already exist (seeded at an earlier, lower target).
    rt.worker_count.store(4, Ordering::Relaxed);

    // target (8) > current (4): growth branch spawns (8 - 4) = 4 workers, so
    // the live worker count becomes 8.
    rt.ensure_worker_capacity();

    let counted = rt.worker_count.load(Ordering::Relaxed);
    assert_eq!(
        counted, 8,
        "worker_count should equal live workers (8) after growing 4 -> 8; got \
         {counted}: compare_exchange set it to target=8, then spawn_workers \
         fetch_add'd the (target-current)=4 delta again"
    );
}
```

Result:

```
test runtime::tokio_mt::tests::ensure_worker_capacity_double_counts_on_growth ... FAILED
assertion `left == right` failed: worker_count should equal live workers (8) after growing 4 -> 8; got 12 ...
  left: 12
 right: 8
```

aws-sdk-s3-transfer-manager/src/types.rs
```diff
@@ -81,6 +81,29 @@ impl MemoryBudgetConfig {
...
+    ///
+    /// An escape hatch for embedding the transfer manager where it must not
+    /// spawn its own OS threads. This is not a performance tuning knob —
+    /// `Managed` is faster by design; prefer it unless you specifically need
+    /// the transfer manager to run on your existing runtime.
+    ///
+    /// Requires a multi-threaded runtime: transfer workers run via
+    /// `tokio::spawn`, so on a current-thread runtime they serialize onto one
+    /// thread and throughput collapses.
+    CurrentTokio,
```
landonxjames: Could we rename this `MultiThreadTokio` or something? `CurrentTokio` makes me think of tokio's `current_thread` runtime which might be confusing since this is specifically the multi threaded runtime.

aws-sdk-s3-transfer-manager/src/client.rs
```diff
@@ -201,14 +201,36 @@ impl Client {
...
+                    }
+                    Arc::new(builder.build())
+                }
+                RuntimeMode::CurrentTokio => {
+                    // CurrentTokio runs transfer workers on the caller's runtime via
+                    // `tokio::spawn`. On a current-thread runtime they all serialize onto
+                    // the one thread, collapsing throughput with no other symptom. Warn
+                    // loudly rather than let that happen silently.
+                    if matches!(
+                        tokio::runtime::Handle::try_current().map(|h| h.runtime_flavor()),
```
landonxjames: Checking on `Client::new` feels like it might be an issue for users that construct their client synchronously before creating their runtime. Should the check be moved back to the first `spawn` site?

landonxjames (approved):
LGTM, one pre-existing double counting bug and a couple of questions, no blockers

aws-sdk-s3-transfer-manager/src/client.rs
vcjana: Can we check this live where the workers are spawned instead? A client can be built outside Tokio and used later from a current-thread runtime, so checking only in `Client::new` can miss this warning.

vcjana: this is the case I was thinking of too. Checking at the first spawn site seems better since it observes the runtime actually being used

vcjana (approved):

PR #159, opened 2026-07-16
vnext composite flow control

aajtodd:
## Summary

Composite directory transfers (`download_objects`, `upload_objects`) materialized child transfers as a side effect of `poll_work`, up to a batch of 32 per poll and a per-transfer cap that defaulted to 4096. Nothing tied that cap to the scheduler's concurrency target, so a directory transfer built thousands of child state machines to feed a pipeline that sustained a few hundred in-flight requests.

This change bounds child materialization to the concurrency the pipeline actually drains, and along the way decouples directory listing from the child budget and closes a lost-wake hang on the spawn-failure path. It introduces two `PollWork` outcomes (`Spawned`, and a `spawned` field on `Ready`) that let a composite spawn one child per poll and be re-polled only while capacity remains. There is no behavior change for non-composite (single-object) transfers.

## Why

The ready set gates dispatch on in-flight work items — a transfer is polled for new work only while `dispatched < target`. Child *materialization*, however, was governed by an independent per-transfer cap (`pipeline_depth`, default 4096) and a fixed spawn batch of 32. The two were never connected.

The consequences compounded:

- **Over-materialization.** A `download_objects` over a large prefix spawned children until it hit the 4096 cap, regardless of a network target that settled near 250. Telemetry showed `children_active` pinned at the cap while the target crawled — 10–30× more child state machines alive than the pipeline could ever have in flight. Every one of those carries scheduler-tree presence and per-poll bookkeeping for no throughput gain.
- **Inflated latency, then aborts.** Splitting a fixed body of work across thousands of transfers inflated per-request latency enough to trip the adaptive latency deadline, which is not retryable — so small-file directory downloads *failed iterations outright* rather than merely running slowly.
- **Listing starved by the same cap.** The walk that enumerates keys was gated on the child budget, not on how many keys were buffered, so discovery throttled exactly when spawning was active. Under a fixed 250-way target the key buffer emptied on ~40% of sampled polls, stalling on `ListObjectsV2` latency.
- **A lost-wake hang.** The pre-fusion poll ladder routed a spawn that *claimed* a queue entry but *failed to orchestrate* (e.g. a bad key under the `Continue` policy) to a `Pending` arm with no wake. Once the walker was exhausted, the remaining queued entries stranded forever. A single unreadable file could hang an entire directory upload.

The through-line: child materialization, network concurrency, and listing supply are three separate resources that were being governed by one number.

### Single-ticket spawn — `PollWork::Spawned`

A composite's `poll_work` now spawns at most one child per call and reports it with a new outcome:

```rust
pub(crate) enum PollWork {
    Ready { io: IoRequest, spawned: bool },
    Spawned,   // enqueued one child; re-poll me while capacity remains
    Pending,
    Done,
}
```

`Spawned` tells the scheduler that a child was enqueued and the transfer should be re-polled — but dispatches no work item and charges no concurrency ticket. The scheduler charges the parent's vruntime and reinserts it under the held claim; the spawned child accounts for its own ticket when it is later polled to `Ready`.

Because a transfer is only polled while `dispatched < target`, spawning is now self-pacing: the parent is re-polled to spawn the next child only while the pipeline has room, so child materialization tracks the concurrency target implicitly, with no arithmetic coupling to it. The 4096 operating point disappears.

### Directory listing decoupled from the child budget

The walk is regated on the key buffer alone — it advances while `pending_entries` is below a low-water mark, independent of how many children are alive. Listing (which fills a key buffer) and spawning (which consumes network concurrency) are now governed by their own resources.

On the download path, discovered keys are published to the buffer in small chunks *as the walker yields them*, rather than in one batch at the end of a page, so spawning can consume keys while the remainder of the page is still being fetched. Combined with single-flight dispatch this holds the buffer to roughly one page above the low-water mark. Empty-buffer polls drop from ~40% to ~1%.

### Fused reap + spawn, weighted under CFS

The poll ladder produced one action per poll under a strict priority (reap a full batch of ≥64 terminal children, else spawn one, else reap the remainder). That accumulated terminal children behind the reap threshold and, as above, stranded queued entries on a failed spawn. Both composites now run a single fused ladder:

```text
poll_work:
  1. walk        — refill listing if the key buffer is below low-water (early return)
  2. spawn one   — claim one entry; attempt to orchestrate a child
  3. reap        — collect whatever terminal children exist this poll (drain cap, not a threshold)
  4. combine     — reap present         → Ready { io: JoinChildren, spawned }
                   else entry claimed    → Spawned
                   else terminal/idle    → Done / Pending
```

`Ready` carries a `spawned` bool (it was a tuple) so a single poll can both dispatch a `JoinChildren` reap *and* record that it spawned, charging spawn vruntime without a second dispatch ticket. `PollWork::ready(io)` is the helper for the common non-spawning case.

The spawn signal is split into two conditions that key different effects:

| signal | means | drives |
|---|---|---|
| `attempted` | an entry was claimed and has left the queue | **re-poll** (`Spawned`) — keep the parent scheduled so it drains the rest of the queue, even after a failed orchestration |
| `materialized` | a live child was inserted | the **spawn vruntime charge** — a failed orchestration is not charged |

Routing on `attempted` rather than `materialized` is what closes the lost-wake hang: a claimed-but-failed spawn still returns `Spawned`, so the parent is re-polled and drains the remaining entries instead of stranding them.

A spawn is charged `SPAWN_WORK_COST` (half `IO_WORK_COST`) rather than the full IO cost, so a composite parent stays scheduling-competitive with its own children and refills continuously — integrating spawn cadence into CFS rather than governing it with ad-hoc batch-size constants. The charge lands on both individual and group vruntime. For composites with equal spawn-to-IO ratios this leaves cross-group ordering unchanged; composites with *different* ratios accrue group vruntime at different rates per dispatched IO, producing a bounded, self-limiting skew that mildly penalizes the spawn-heavier group — a group cannot spawn its way to a larger dispatch share, only a smaller one. The 1:2 ratio is a starting point, not yet tuned against benchmark data.

### Backstop, not an operating point

The per-transfer child cap default drops from 4096 to 512 — a conservative ceiling above the highest useful concurrency observed in practice, held only so a very large directory cannot materialize without bound. It is no longer the operating point; single-ticket spawn is. The public `max_concurrent_downloads` / `max_concurrent_uploads` knobs still raise it for hosts that can drive more.

### Retry-decision logging

Retries were previously visible only through SDK-internal logging. Two `debug`-level events on the transfer target now record each retry decision (attempt, max attempts, backoff, cause) and budget exhaustion. They fire only on the retry path (rare), are off by default, and render only the error's own context — kind, operation name, S3 error code, request IDs — never object keys, bucket names, URLs, or credentials.

## What this does and does not change

This is a **pipeline-structure and correctness** change. It makes directory transfers materialize the right number of children, keeps the key buffer fed, and removes a hang. On mixed and large-object directory shapes — where over-materialization was inflating latency — it modestly improves throughput. It is **not** a small-file throughput win on its own: the tiny-file ceiling is set by the raw GET path (per-request latency × sustainable in-flight concurrency), which this change does not touch. Bounding materialization is a prerequisite for that work, not a substitute for it.

## Testing

- **Scheduler model.** New tests cover the `Spawned` contract (reinsert-and-repoll, no dispatch ticket), materialization staying bounded at the target under a blocking child mock, the spawn spin-guard (a continuously-spawning composite cannot starve a peer), and cross-group fair share under single-ticket spawn.
- **Lost-wake regression.** A managed-runtime test drives two claimed-but-failed spawns under the `Continue` policy and asserts the parent's completion resolves. It must run on the managed runtime — the immediate poll harness busy-re-polls on `Pending` and cannot observe a lost wake. Reintroducing the `attempted → materialized` regression hangs it; the fix passes.
- **Listing decouple.** A test pins the walk gate to the key buffer (children at the child budget, buffer low → the walk still dispatches), and a lossless/ordered test drives the chunked incremental delivery across a multi-page boundary.
- **vruntime charge.** Unit tests assert `work_generated_spawn` charges both individual and group vruntime by the spawn delta, that the spawn charge is strictly less than the IO charge, and a scheduler test asserts a fused `Ready { spawned: true }` poll charges both the IO and the spawn cost.
- Full library suite green; `clippy` and `fmt` clean.

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/operation/upload_objects/transfer.rs
+++ b/aws-sdk-s3-transfer-manager/src/operation/upload_objects/transfer.rs
@@ -17,7 +17,7 @@ use crate::error;
 use crate::io::walk::{DirEntry, FsWalk};
 use crate::io::InputStream;
 use crate::operation::upload::{Upload, UploadHandle, UploadInput};
-use crate::operation::{DEFAULT_DELIMITER, SPAWN_BATCH_SIZE};
+use crate::operation::DEFAULT_DELIMITER;
 use crate::runtime::sync::Mutex;
 use crate::transfer::{IoRequest, PollWork, Transfer, TransferContext, TransferId, WorkOutcome};
 use crate::types::{FailedTransferPolicy, FailedUpload};
@@ -35,11 +35,22 @@ const MAX_PARALLEL_WALKS: usize = 16;
 /// entries added to `State::pending_entries` in one pass.
 const WALK_BATCH_SIZE: usize = 64;
 
-/// Terminal children reaped per `JoinChildren` work item. A reap injects no
-/// schedulable entity (it joins already-terminal handles), so unlike spawning
-/// it can batch large; the bound only caps worker-thread hold time on the
-/// serial join.
-const REAP_BATCH_SIZE: usize = 256;
+/// Maximum terminal children drained into a single `JoinChildren` work item per
+/// poll. NOT an accumulation threshold: `poll_work` reaps whatever terminals
+/// exist every poll (fused with spawning), so this only caps how many
+/// already-terminal handles one `execute_join_children` awaits back-to-back,
+/// bounding worker-thread hold time on the serial join. Kept modest so a poll's
+/// reap work stays small and terminals are retired continuously rather than in
+/// large lumps.
+const MAX_REAP_PER_POLL: usize = 64;
+
+/// Low-water mark for the entry buffer: `dispatch_walk` advances a walker only
+/// while `pending_entries` is below this, keeping the buffer fed ahead of spawn
+/// draining it. Sized to the walkers' aggregate burst
+/// (`MAX_PARALLEL_WALKS * WALK_BATCH_SIZE`) so every walker can be productively
+/// in flight before the buffer is considered full. Independent of the child
+/// in-flight budget.
+const WALK_LOW_WATER: usize = MAX_PARALLEL_WALKS * WALK_BATCH_SIZE;
 
 /// Work data variants for the UploadObjectsTransfer state machine.
 #[derive(Debug)]
@@ -82,7 +93,7 @@ struct ClaimedEntry {
 /// consumed by [`UploadObjectsTransfer::merge_spawned`] (paired
 /// decrement under the state lock) or released on `Drop` if
 /// `consume` is not reached. Constructed only by
-/// [`UploadObjectsTransfer::claim_spawn_batch`].
+/// [`UploadObjectsTransfer::claim_one`].
 ///
 /// The type makes the counter pairing structural: `merge_spawned`'s
 /// signature requires the token, so a caller cannot forget to release
@@ -97,7 +108,7 @@ struct Reservation {
 impl Reservation {
     /// Caller holds `state.lock`. Decrements `children_reserved` by the
     /// owned count and marks the reservation consumed. Always paired
-    /// 1:1 with the `claim_spawn_batch` that produced this token.
+    /// 1:1 with the `claim_one` that produced this token.
     fn consume(mut self, state: &mut State) {
         debug_assert!(
             !self.consumed,
@@ -116,7 +127,7 @@ impl Reservation {
 
 impl Drop for Reservation {
     fn drop(&mut self) {
-        if !self.consumed {
+        if !self.consumed && self.count > 0 {
             let mut state = self.transfer.state.lock();
             state.children_reserved = state.children_reserved.saturating_sub(self.count);
             tracing::warn!(
@@ -129,7 +140,7 @@ impl Drop for Reservation {
     }
 }
 
-/// Outcome of phase 1 (`claim_spawn_batch`).
+/// Outcome of phase 1 (`claim_one`).
 enum SpawnDecision {
     /// Abort policy tripped on a pre-orchestration failure. Caller exits
     /// `poll_work` immediately with `PollWork::Done`.
@@ -318,7 +329,7 @@ impl fmt::Debug for ChildTransfer {
 /// Mutable state of an `upload_objects` transfer.
 ///
 /// Walker enumeration produces `DirEntry`s into `pending_entries`.
-/// `poll_work` consumes them through `claim_spawn_batch` into child
+/// `poll_work` consumes them through `claim_one` into child
 /// `UploadHandle`s in `children`, reaps terminal children via
 /// `JoinChildren` work items, and accumulates outcomes into
 /// `successful_uploads` and `failed`.
@@ -328,7 +339,7 @@ impl fmt::Debug for ChildTransfer {
 /// the counter if `consume` is not reached:
 ///
 ///   counter             increment site          paired token (consumed at)
-///   children_reserved   claim_spawn_batch       Reservation (merge_spawned)
+///   children_reserved   claim_one               Reservation (merge_spawned)
 ///   reaping_in_flight   drain_terminal_children ReapingBatch (execute_join_children)
 ///   in_flight_walks     dispatch_walk           WalkSlot (execute_advance_walker)
 ///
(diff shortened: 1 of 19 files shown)


aws-sdk-s3-transfer-manager/src/operation/upload_objects/transfer.rs
```diff
@@ -453,115 +464,126 @@ impl UploadObjectsTransfer {
...
+            };
+        }
+        if attempted {
+            return PollWork::Spawned;
         }
 
-        // Inactive and fully drained: signal terminal. Children are cancelled
-        // via `scheduler.cancel_transfer(parent_id)` from the handle's
-        // abort/drop/join paths, which cascade outside the state lock.
+        // 5: Terminal/idle -- neither reaped nor claimed an entry this poll.
```
landonxjames, line 578: Think theres an asymmetry here when compared to `download_objects` and I think the download path has it right.

This fast path returns `Done` as soon as `is_active()` is false, regardless of whether the transfer still has in-flight walks (`in_flight_walks > 0`), in-flight reaping batches (`reaping_in_flight > 0`), reserved but unmerged claims (`children_reserved > 0`), or in the harmful case, still active child transfers in `children`. The full drain condition already exists in `check_terminal`, but this fast path is above it, so it's never consulted on the inactive path. `claim_one`'s `SpawnDecision::Abort` and `merge_spawned`'s `abort()`/`Some(Done)` return `Done` the same way, while previously-spawned siblings are still live.

`download_objects::check_terminal` does the opposite (`download_objects/transfer.rs` 627–646), on `!is_active()` it returns `None` (stays `Pending`) until `children.is_empty() && children_reserved == 0 && reaping_in_flight == 0 && !walk_in_flight`, and only then signals `Done`.

**Possible Fix:** Route the inactive path through a terminal check mirroring download and signal `Done` only once `in_flight_walks == 0 && reaping_in_flight == 0 && children_reserved == 0 && children.is_empty()`; otherwise return `Pending` and let the in-flight work wake the parent as it drains (the execute callbacks already `try_wake`).

Pair of tests showing the diverging behavior (both meant for transfer.rs in their respective operation):

```rust
// upload_objects/transfer.rs — reproduces the bug: Done without draining.
#[cfg_attr(miri, ignore)]
#[tokio::test]
async fn upload_objects_returns_done_with_outstanding_work_when_inactive() {
    let dir = tempdir().unwrap();
    fs::write(dir.path().join("a.txt"), "x").unwrap();
    let (transfer, _rx) = setup(
        dir.path(),
        FailedTransferPolicy::Continue,
        mock_s3_success(),
        false,
    );

    // First poll dispatches a walker advance; the walk is now in flight.
    let w1 = transfer.poll_work();
    assert!(matches!(w1, PollWork::Ready { .. }), "first poll should dispatch a walker advance, got {w1:?}");
    assert_eq!(transfer.inner.state.lock().in_flight_walks, 1, "a walker advance should be in flight");

    // Transfer goes inactive while that walk work item is still outstanding
    // (we hold `w1` un-executed, so its WalkSlot is not dropped).
    transfer.ctx().set_cancelled();

    let w2 = transfer.poll_work();
    assert!(matches!(w2, PollWork::Done),
        "BUG(F1): upload returns Done without draining the in-flight walk, got {w2:?}");
    assert_eq!(transfer.inner.state.lock().in_flight_walks, 1,
        "the walk was still outstanding at the instant Done was returned");

    drop(w1);
}
```

```rust
// download_objects/transfer.rs — control: download drains (returns Pending).
#[cfg_attr(miri, ignore)]
#[tokio::test]
async fn download_objects_drains_outstanding_work_when_inactive() {
    let dir = tempdir().unwrap();
    let (transfer, _rx) = setup(dir.path(), FailedTransferPolicy::Continue, mock_s3_success());

    let w1 = transfer.poll_work();
    assert!(matches!(w1, PollWork::Ready { .. }), "first poll should dispatch a walker advance, got {w1:?}");
    assert!(transfer.inner.state.lock().walk_in_flight, "a walker advance should be in flight");

    transfer.ctx().set_cancelled();

    let w2 = transfer.poll_work();
    assert!(matches!(w2, PollWork::Pending),
        "download drains: it must NOT return Done while a walk is in flight, got {w2:?}");

    drop(w1);
}
```

landonxjames (approved):
LGTM. One potential issue with `upload_objects` finishing without draining remaining work.

PR #162, opened 2026-08-13
upload: set MpuObjectSize on CompleteMultipartUpload

vcjana:
## Summary

`CompleteMultipartUpload` was sent without `MpuObjectSize`, so S3 had no client-supplied size to cross-check the assembled object against — a dropped or duplicated part completed as a wrong-sized object. 

Threads the already-known `content_length` through `Transferring → Completing` and sets it on the request, rather than summing parts at completion time (which would recompute the size from the same state that produced the
bug).

## Test plan

- [x] fmt / clippy clean (default + \`--cfg e2e_test\`)
- [x] lib + upload + upload_objects + upload_checksum tests green (630 + 25 + 28 + 2)
- [x] Both mock tests adversarially verified: stripped the fix, each test failed, restored, green
- [x] e2e (\`test_multi_part_upload_sends_mpu_object_size\`) runs in CI's \`test-hll\` step against a real bucket

```
By submitting this pull request, I confirm that you can use, modify, copy,
and redistribute this contribution, under the terms of your choice.
```

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/operation/upload/context.rs
+++ b/aws-sdk-s3-transfer-manager/src/operation/upload/context.rs
@@ -30,6 +30,8 @@ pub(crate) enum UploadState {
         parts_in_flight: usize,
         completed_parts: Vec<CompletedPart>,
         response_builder: UploadOutputBuilder,
+        /// Full object content length, carried through to CompleteMPU as `MpuObjectSize`.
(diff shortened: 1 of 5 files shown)


aajtodd (approved):
Looks good. Probably will have to be massaged a bit when we handle unknown content length/stream.

PR #163, opened 2026-08-14
upload: conditional-write support (IfMatch/IfNoneMatch)

### LINKED ISSUE #117: Support `if_match` and `if_none_match` in `UploadInput`
Fields [if_match](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/operation/complete_multipart_upload/builders/struct.CompleteMultipartUploadInputBuilder.html#method.if_match) and [if_none_match](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/operation/complete_multipart_upload/builders/struct.CompleteMultipartUploadInputBuilder.html#method.if_none_match), among others, can be configured on `CompleteMultiUpload`. The `UploadInput` type should allow customers to configure these fields, so they can be threaded through the input for `CompleteMultiUpload`.

vcjana:
## Summary

Adds conditional-write support to upload: callers set `if_match` / `if_none_match` on `UploadInput` ("overwrite only if unchanged" / "create only if absent"). The headers are forwarded on the two requests S3 accepts them — `PutObject` (single-PUT) and `CompleteMultipartUpload` (multipart).

A failed precondition is **not** promoted to its own error kind. It arrives as `ErrorKind::ServiceError` with `Error::code() == Some("PreconditionFailed")`, which is documented on `ServiceError` so it is discoverable. S3 reports that code in the response body, so it is reachable whatever HTTP status carries it — including the 200-with-error-XML case a status check would miss. Such a failure is terminal in the retry classifier via the existing non-throttle service-error path; no dedicated arm is needed, since it is neither transient transport nor a throttle code.

S3 doesn't accept these headers on `CreateMultipartUpload` (no builder setter), so on the multipart path they go on
[`copy_fields_to_complete_mpu_request`](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/c4e27dedf65cf98882d613234236a9b253da3637/aws-sdk-s3-transfer-manager/src/operation/upload/input/convert.rs) — which also re-enables the staged assertion. Scope is single-object upload.

New public surface: the two `UploadInput` fields.

Closes #117

## Test plan

- [x] `fmt` / `clippy` clean (default + `--cfg e2e_test`); lib + integration + doc tests green
- [x] Mocks pin both paths: header forwarded; a 412 fails the transfer with the `PreconditionFailed` service code reachable, issued exactly once
- [x] A unit test pins the classification contract end to end — kind, code, operation name, and the underlying `SdkError` preserved as the source
- [x] `test_upload_if_none_match_star` (`#[cfg(e2e_test)]`) runs against a real bucket in the e2e job, confirming real S3 supplies the code the mocks assert

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/error.rs
+++ b/aws-sdk-s3-transfer-manager/src/error.rs
@@ -82,14 +82,13 @@ pub enum ErrorKind {
     /// A request to S3 failed. The originating operation, service error code,
     /// message, and request ids are available via the accessors on [`Error`]
     /// ([`Error::operation_name`], [`Error::code`], [`Error::request_id`]).
+    ///
+    /// Service-specific outcomes are identified by their code rather than by a
+    /// kind of their own. A conditional write whose precondition did not hold
+    /// (`if_match` / `if_none_match` on an upload) arrives here with
+    /// [`Error::code`] of `"PreconditionFailed"`; S3 reports that code in the
+    /// response body, so it is available whatever HTTP status carries it.
     ServiceError,
-
-    /// A conditional-write precondition set on an [`upload`](crate::operation::upload)
-    /// (`If-Match` / `If-None-Match`) did not match; S3 responded with
-    /// `412 Precondition Failed`. The originating operation and service metadata
-    /// are still available via the accessors on [`Error`], the same as for a
-    /// generic [`ServiceError`](Self::ServiceError).
-    PreconditionFailed,
 
     /// Object integrity validation failed: the received bytes did not match the
     /// expected checksum. See the wrapped [`IntegrityError`] for detail.
@@ -270,21 +269,6 @@ impl Error {
         )
     }
 
-    /// Whether this error is a conditional-write precondition failure: a
-    /// caller-supplied `If-Match` / `If-None-Match` on an
-    /// [`upload`](crate::operation::upload) that S3 rejected with
-    /// `412 Precondition Failed`.
-    ///
-    /// Kept `pub(crate)` for now: the only consumer is the upload retry
-    /// classifier, and the public precondition signal callers branch on is the
-    /// [`ErrorKind::PreconditionFailed`] variant reached via [`Error::kind`].
-    /// Promoting this to `pub` later is non-breaking if a caller convenience is
-    /// wanted; adding it to the public surface now — during the 1.0 API freeze
-    /// (RUST-1202) — is the harder-to-reverse choice.
-    pub(crate) fn is_precondition_failed(&self) -> bool {
-        matches!(self.kind, ErrorKind::PreconditionFailed)
-    }
-
     /// Whether this error was a transient transport failure (connection IO error
     /// or client-side timeout) as opposed to a service response. Such failures
     /// are safe to re-issue and may not have been recovered by the SDK's own
@@ -381,12 +365,6 @@ impl fmt::Display for Error {
                     if let Some(c) = &m.code {
                         write!(f, " ({c})")?;
                     }
-                }
-            }
-            ErrorKind::PreconditionFailed => {
-                write!(f, "precondition failed")?;
-                if let Some(m) = self.service() {
-                    write!(f, " calling {}", m.operation)?;
                 }
             }
             ErrorKind::IntegrityError(ie) => {
@@ -548,19 +526,6 @@ where
     }
 }
 
-/// Whether an `SdkError` is a service response with HTTP status 412 (Precondition
-/// Failed). Used to distinguish a user-set `If-Match` / `If-None-Match`
-/// precondition failure from a generic service error, so callers can branch on
-/// [`ErrorKind::PreconditionFailed`] without string-matching a service code.
-fn sdk_err_is_precondition_failed<E>(
-    e: &SdkError<E, aws_smithy_runtime_api::client::orchestrator::HttpResponse>,
-) -> bool {
-    matches!(
-        e,
-        SdkError::ServiceError(ctx) if ctx.raw().status().as_u16() == 412
-    )
-}
-
 /// Whether a single [`ConnectorError`] frame denotes a transient transport
 /// failure: an IO error, a client-side timeout, or one the http client tagged
 /// [`RetryErrorKind::TransientError`] (an incomplete/truncated response, TLS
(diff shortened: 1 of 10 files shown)


aws-sdk-s3-transfer-manager/src/retry.rs
```diff
@@ -273,6 +280,7 @@ fn retry_cause(ge: &GuardError<Error>) -> String {
 pub(crate) fn classify_upload_part_retry(ge: &GuardError<Error>) -> RetryDecision {
     match ge {
         GuardError::DeadlineExceeded(_) => RetryDecision::NoRetry,
+        GuardError::Inner(e) if e.is_precondition_failed() => RetryDecision::NoRetry,
```
aajtodd: This isn't wrong but isn't this the default to fall through to NoRetry?

aws-sdk-s3-transfer-manager/src/operation/upload/input.rs
```diff
@@ -57,6 +57,16 @@ pub struct UploadInput {
     pub content_type: Option<String>,
     #[doc = std::include_str!("checksum_strategy.md")]
     pub checksum_strategy: Option<ChecksumStrategy>,
+    /// <p>Uploads the object only if the ETag (entity tag) value provided during the WRITE operation matches the ETag of the object in S3. If the ETag values do not match, the operation returns a <code>412 Precondition Failed</code> error.</p>
+    /// <p>If a conflicting operation occurs during the upload S3 returns a <code>409 ConditionalRequestConflict</code> response. On a 409 failure you should fetch the object's ETag and retry the upload.</p>
+    /// <p>Expects the ETag value as a string.</p>
+    /// <p>For more information about conditional requests, see <a href="https://tools.ietf.org/html/rfc7232">RFC 7232</a>, or <a href="https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-requests.html">Conditional requests</a> in the <i>Amazon S3 User Guide</i>.</p>
+    pub if_match: Option<String>,
```
aajtodd, line 64: Interesting these were missing off the input.

aws-sdk-s3-transfer-manager/src/operation/upload/handle.rs
```diff
@@ -167,6 +167,9 @@ impl UploadHandle {
             match abort_policy {
                 FailedMultipartUploadPolicy::Retain => Ok(AbortedUpload::default()),
                 FailedMultipartUploadPolicy::AbortUpload => {
+                    // AbortMultipartUpload carries no `If-Match`/`If-None-Match`
```
aajtodd: This comment seems out of place or like it appears somewhat randomly.

aws-sdk-s3-transfer-manager/src/operation/upload/input/convert.rs
```diff
@@ -483,17 +491,8 @@ mod tests {
             upload_req.expected_bucket_owner(),
             complete_mpu_req.expected_bucket_owner()
         );
-        // TODO(https://github.com/awslabs/aws-s3-transfer-manager-rs/issues/117): Enable these assertions
```
aajtodd, line 486: Reminder to link PR to issue and close out once merged (or released)

aws-sdk-s3-transfer-manager/src/error.rs
```diff
@@ -84,6 +84,13 @@ pub enum ErrorKind {
     /// ([`Error::operation_name`], [`Error::code`], [`Error::request_id`]).
     ServiceError,
 
+    /// A conditional-write precondition set on an [`upload`](crate::operation::upload)
+    /// (`If-Match` / `If-None-Match`) did not match; S3 responded with
+    /// `412 Precondition Failed`. The originating operation and service metadata
+    /// are still available via the accessors on [`Error`], the same as for a
+    /// generic [`ServiceError`](Self::ServiceError).
+    PreconditionFailed,
```
aajtodd: I'm wondering whether we want this as a distinct kind or to leave it as `ServiceError` and have users access it through `Error::service()` function. I tend to think we should not promote this to a distinct error kind.

@ysaito1001 any opinion on errors and hierarchy here?

aws-sdk-s3-transfer-manager/src/error.rs
```diff
@@ -520,6 +548,19 @@ where
...
+/// Whether an `SdkError` is a service response with HTTP status 412 (Precondition
+/// Failed). Used to distinguish a user-set `If-Match` / `If-None-Match`
+/// precondition failure from a generic service error, so callers can branch on
+/// [`ErrorKind::PreconditionFailed`] without string-matching a service code.
+fn sdk_err_is_precondition_failed<E>(
+    e: &SdkError<E, aws_smithy_runtime_api::client::orchestrator::HttpResponse>,
+) -> bool {
+    matches!(
+        e,
+        SdkError::ServiceError(ctx) if ctx.raw().status().as_u16() == 412
```
aajtodd: S3 has historically had an issue where they _can_ send a 200 response with an error XML payload. We may want to check or extend this to something like ` == 412 ||   ctx.as_service_error().is_some_and(|e| e.code() == Some("PreconditionFailed"))` as wider catch.

aajtodd (commented):
Overall looks good.

aws-sdk-s3-transfer-manager/src/error.rs
ysaito1001: Yeah, point taken. Feels the "granularity" of `PreconditionFailed` is finer than the other variants; `PreconditionFailed` belongs to bare s3, whereas other variants live in "transfer-manager" layer.

aws-sdk-s3-transfer-manager/src/retry.rs
vcjana: Removed the arm, folded its test into the existing non-throttle service error one.

aws-sdk-s3-transfer-manager/src/operation/upload/input.rs
vcjana: Yeah, never got wired through even though the SDK has them. Kept the field docs mirrored from the SDK so they stay in sync.

aws-sdk-s3-transfer-manager/src/operation/upload/handle.rs
vcjana: Removed. It was explaining the 412 classification, which is removed now

aws-sdk-s3-transfer-manager/src/operation/upload/input/convert.rs
vcjana: Updated the description

aws-sdk-s3-transfer-manager/src/error.rs
vcjana: dropped it. A 412 comes back as ServiceError with` code() == Some("PreconditionFailed")` now, and I noted that on `ErrorKind::ServiceError` so it's still discoverable without the variant.

aws-sdk-s3-transfer-manager/src/error.rs
vcjana: `code()` covers the 200-with-error-XML case since it's read from the body, and that's what detection uses now.

Error doesn't expose the status at all though, so if you want both signals available I can add a `status()` accessor felt like that belonged in its own change rather than here.

aws-sdk-s3-transfer-manager/src/error.rs
```diff
@@ -82,6 +82,12 @@ pub enum ErrorKind {
     /// A request to S3 failed. The originating operation, service error code,
     /// message, and request ids are available via the accessors on [`Error`]
     /// ([`Error::operation_name`], [`Error::code`], [`Error::request_id`]).
+    ///
+    /// Service-specific outcomes are identified by their code rather than by a
+    /// kind of their own. A conditional write whose precondition did not hold
+    /// (`if_match` / `if_none_match` on an upload) arrives here with
+    /// [`Error::code`] of `"PreconditionFailed"`; S3 reports that code in the
+    /// response body, so it is available whatever HTTP status carries it.
     ServiceError,
```
aajtodd, line 91: Nothing to do in this PR since this follows the current error hierarchy and mostly matches the SDK. It's probably worth pulling at before GA whether that's the right choice or if we can/want to make errors like `PreconditionFailed` or others easier to access in some way.

aws-sdk-s3-transfer-manager/src/scheduler/scheduler.rs
```diff
@@ -2203,6 +2208,7 @@ mod tests {
     }
 
     #[cfg_attr(miri, ignore)]
+    #[cfg_attr(s3_tm_asan, ignore)]
```
aajtodd, line 2211: Were these failing CI?

aws-sdk-s3-transfer-manager/tests/e2e_transfer_test.rs
```diff
@@ -550,3 +550,68 @@ async fn test_multi_part_upload_sends_mpu_object_size() {
...
+///      mock might not share with the service.
+///
+/// Runs against the general-purpose bucket only. Per the S3 conditional-writes
+/// docs, `If-None-Match` / `If-Match` are defined in terms of general-purpose
+/// (versioned / version-suspended) bucket behavior; support on S3 Express One
+/// Zone directory buckets is not affirmed there, so we don't assert it here.
+// TODO: extend to the S3 Express bucket once conditional-write support on
+// directory buckets is confirmed on a real run.
+#[tokio::test]
+async fn test_upload_if_none_match_star() {
```
aajtodd, line 572: Maybe worth an if-match test equivalent?

aajtodd (approved):
Looks good. Nothing blocking.

ysaito1001 (approved):

aws-sdk-s3-transfer-manager/tests/e2e_transfer_test.rs
vcjana: added `test_upload_if_match_etag` in 85b6b02

aws-sdk-s3-transfer-manager/src/scheduler/scheduler.rs
vcjana: yeah, asan was red. those three leave work in flight so the `Handle` Arc never hits zero and the S3 client leaks at exit - same cause as your `test_handle_managed` [note](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/846ed6098bfeffcdc8c5a3e6875ecda89ed96dba/aws-sdk-s3-transfer-manager/src/scheduler/scheduler.rs#L880-L889), that gate just didn't cover `test_handle`

PR #164, opened 2026-08-15
upload: support unknown content length on streaming uploads

vcjana:
## Summary

An upload from a `PartStream` with no size-hint upper bound panicked on `content_length required` instead of transferring. Such a source is always `is_mpu_only`, so it already routes through multipart; the only assumption that did not hold was that `Transferring` knows a part count up front.

Replaces `total_parts` with a `PartPlan` enum: `Known` carries the part count, `Unknown` dispatches parts speculatively and ends when the reader reports end-of-stream. An empty stream still has to produce a part, since S3 rejects a `CompleteMultipartUpload` that lists none, so one dispatch synthesizes a zero-length part 1 — claimed by a check-and-set under the state lock, because several part-work items can be dispatched before any of them reads and more than one can observe end-of-stream with nothing emitted.

Two follow-on bits: `MpuObjectSize` (#162) has no declared length to send here, so it is summed from the parts actually uploaded — the known path keeps its declared value, which is an independent witness a sum cannot replace. And the 10,000-part ceiling now fails with an actionable error instead of letting S3 reject part 10,001.

## Test plan

- [x] fmt / clippy clean (default + `--cfg e2e_test`)
- [x] lib + upload + upload_objects + upload_checksum tests green (631 + 11 + 28 + 25)
- [x] Adversarially verified: reinstating the original flag ordering, neutering the ceiling guard, and dropping the end-of-stream `total_bytes` set each fail their test, then restored green
- [x] e2e (`test_upload_unknown_content_length`, `test_upload_unknown_content_length_empty`) run against a real bucket in the e2e job — not yet exercised locally

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/io/part_reader.rs
+++ b/aws-sdk-s3-transfer-manager/src/io/part_reader.rs
@@ -5,6 +5,7 @@
 use std::cmp;
 use std::fs::File;
 use std::ops::DerefMut;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 
 use bytes::{Buf, Bytes};
@@ -89,6 +90,14 @@ impl Builder {
 pub(crate) struct PartReader {
     inner: Inner,
     stream_cx: StreamContext,
+    /// Count of parts this reader has yielded (returned as `Ok(Some(..))`).
+    ///
+    /// Incremented in [`next_part`](Self::next_part) as each part is produced, so it is ordered with
+    /// the reads: end-of-stream (`Ok(None)`) is only observed after every yielded part has already
+    /// bumped this. That lets a caller driving speculative, concurrent reads decide two things
+    /// race-free — whether the stream was empty (`0` at end-of-stream) and whether it has exceeded a
+    /// part limit — without a second lock of its own that could interleave with the read.
+    parts_yielded: AtomicU64,
 }
 
 impl PartReader {
@@ -106,12 +115,21 @@ impl PartReader {
         };
 
         let stream_cx = StreamContext::new(part_size, direct_io, metrics, telemetry);
-        Ok(Self { inner, stream_cx })
+        Ok(Self {
+            inner,
+            stream_cx,
+            parts_yielded: AtomicU64::new(0),
+        })
     }
 
     #[allow(dead_code)] // TODO: re-wire upload part validation
     pub(crate) fn part_size(&self) -> usize {
         self.stream_cx.part_size()
+    }
+
+    /// Number of parts yielded so far. See the field docs on the ordering guarantee.
+    pub(crate) fn parts_yielded(&self) -> u64 {
+        self.parts_yielded.load(Ordering::Acquire)
     }
 }
 
@@ -124,10 +142,22 @@ enum Inner {
 
 impl PartReader {
     pub(crate) async fn next_part(&self) -> Result<Option<PartData>, Error> {
+        // `parts_yielded` is passed down so each reader increments it *inside the same lock that
+        // produces the part*. Incrementing here in the outer call — after the inner released its
+        // lock — would leave a window in which a concurrent reader observes end-of-stream with a
+        // stale count, which is the race this counter exists to close.
         match &self.inner {
-            Inner::Bytes(bytes) => bytes.next_part(&self.stream_cx).await,
-            Inner::Fs(path_body) => path_body.next_part(&self.stream_cx).await,
-            Inner::Dyn(part_stream) => part_stream.next_part(&self.stream_cx).await,
+            Inner::Bytes(bytes) => bytes.next_part(&self.stream_cx, &self.parts_yielded).await,
+            Inner::Fs(path_body) => {
+                path_body
+                    .next_part(&self.stream_cx, &self.parts_yielded)
+                    .await
+            }
+            Inner::Dyn(part_stream) => {
+                part_stream
+                    .next_part(&self.stream_cx, &self.parts_yielded)
+                    .await
+            }
         }
     }
 
@@ -187,7 +217,11 @@ impl BytesPartReader {
 }
 
 impl BytesPartReader {
-    async fn next_part(&self, stream_cx: &StreamContext) -> Result<Option<PartData>, Error> {
+    async fn next_part(
+        &self,
+        stream_cx: &StreamContext,
+        parts_yielded: &AtomicU64,
+    ) -> Result<Option<PartData>, Error> {
         let mut state = self.state.lock().expect("lock valid");
         if state.is_end() {
             return Ok(None);
@@ -209,6 +243,8 @@ impl BytesPartReader {
         state.offset += data.len() as u64;
         state.remaining -= data.len() as u64;
         let part = PartData::new(part_number, data).mark_last(state.is_end());
+        // Under `state` — ordered with the yield.
+        parts_yielded.fetch_add(1, Ordering::Release);
         Ok(Some(part))
     }
 }
@@ -245,16 +281,21 @@ impl PathBodyPartReader {
 }
 
 impl PathBodyPartReader {
-    async fn next_part(&self, stream_cx: &StreamContext) -> Result<Option<PartData>, Error> {
-        let (offset, part_number, part_size, is_last) = match self.advance(stream_cx)? {
-            Some(PathBodyReadCursor {
-                offset,
-                part_number,
-                part_size,
-                is_last,
-            }) => (offset, part_number, part_size, is_last),
-            None => return Ok(None),
-        };
+    async fn next_part(
+        &self,
+        stream_cx: &StreamContext,
+        parts_yielded: &AtomicU64,
+    ) -> Result<Option<PartData>, Error> {
+        let (offset, part_number, part_size, is_last) =
+            match self.advance(stream_cx, parts_yielded)? {
+                Some(PathBodyReadCursor {
+                    offset,
+                    part_number,
+                    part_size,
+                    is_last,
+                }) => (offset, part_number, part_size, is_last),
+                None => return Ok(None),
+            };
         // grab a buffer to fill from the context
         let mut dst = stream_cx.new_buffer(part_size as usize);
         // SAFETY: We set the length to capacity so the read has a full slice to fill.
@@ -286,7 +327,11 @@ impl PathBodyPartReader {
     // Advances the `PartReaderState` to the next state and returns `PathBodyReadCursor`
     // (offset, part_number, part_size, is_last), which will be used in the upcoming
     // `read_file_chunk` execution.
-    fn advance(&self, stream_cx: &StreamContext) -> Result<Option<PathBodyReadCursor>, Error> {
+    fn advance(
+        &self,
+        stream_cx: &StreamContext,
+        parts_yielded: &AtomicU64,
+    ) -> Result<Option<PathBodyReadCursor>, Error> {
         let mut state = self.state.lock().expect("lock valid");
         if state.is_end() {
             return Ok(None);
@@ -305,6 +350,10 @@ impl PathBodyPartReader {
         state.offset += part_size;
         state.part_number += 1;
         state.remaining -= part_size;
+
+        // Under `state` — ordered with the yield. A part is committed here even though the file read
+        // happens after the lock is released; `advance` returning `Some` is the yield.
+        parts_yielded.fetch_add(1, Ordering::Release);
 
         Ok(Some(PathBodyReadCursor {
             offset,
@@ -382,11 +431,23 @@ impl DynPartReader {
             inner: tokio::sync::Mutex::new(inner),
         }
     }
-    async fn next_part(&self, stream_cx: &StreamContext) -> Result<Option<PartData>, Error> {
+    async fn next_part(
+        &self,
+        stream_cx: &StreamContext,
+        parts_yielded: &AtomicU64,
+    ) -> Result<Option<PartData>, Error> {
         // TODO - can we do better than a mutex here? should we spawn a dedicated task and use channels instead
         let mut stream = self.inner.lock().await;
         match stream.next(stream_cx).await {
-            Some(result) => result.map(Some).map_err(|err| err.into()),
+            Some(Ok(part)) => {
+                // Under the stream mutex — ordered with the yield. This is the whole reason the
+                // count lives on the reader: reads here are serialized, so a caller that observes
+                // end-of-stream (the `None` arm) is guaranteed to see every prior part counted, with
+                // no separate lock of its own that could interleave between a read and the count.
+                parts_yielded.fetch_add(1, Ordering::Release);
+                Ok(Some(part))
+            }
+            Some(Err(err)) => Err(err.into()),
             None => Ok(None),
         }
     }
@@ -401,6 +462,7 @@ impl DynPartReader {
 #[cfg(test)]
 mod test {
     use std::io::Write;
+    use std::sync::atomic::AtomicU64;
     use std::task::Poll;
 
     use bytes::{Buf, Bytes};
@@ -577,9 +639,10 @@ mod test {
         let data = Bytes::from("test data for alignment error");
         let reader = BytesPartReader::new(data);
         let stream_cx = test_stream_cx(5);
+        let yielded = AtomicU64::new(0);
 
         // First call should succeed
-        let result = reader.next_part(&stream_cx).await;
+        let result = reader.next_part(&stream_cx, &yielded).await;
         assert!(result.is_ok());
 
         // Manually corrupt the offset to create misalignment
@@ -589,7 +652,7 @@ mod test {
         }
 
         // Second call should fail with offset_not_aligned_with_part_number error
-        let result = reader.next_part(&stream_cx).await;
+        let result = reader.next_part(&stream_cx, &yielded).await;
         assert!(result.is_err());
     }
 
@@ -599,8 +662,13 @@ mod test {
         let data = Bytes::from("test");
         let reader = BytesPartReader::new(data);
         let stream_cx = test_stream_cx(10);
+        let yielded = AtomicU64::new(0);
 
-        let result = reader.next_part(&stream_cx).await.unwrap().unwrap();
+        let result = reader
+            .next_part(&stream_cx, &yielded)
+            .await
+            .unwrap()
+            .unwrap();
         assert!(result.is_last.unwrap());
     }
 
@@ -616,16 +684,17 @@ mod test {
         };
         let reader = PathBodyPartReader::new(path_body).unwrap();
         let stream_cx = test_stream_cx(5);
+        let yielded = AtomicU64::new(0);
 
         // First advance should succeed
-        let result = reader.advance(&stream_cx).unwrap().unwrap();
+        let result = reader.advance(&stream_cx, &yielded).unwrap().unwrap();
         assert_eq!(result.offset, 10);
         assert_eq!(result.part_number, 1);
         assert_eq!(result.part_size, 5);
         assert!(!result.is_last);
 
         // Second advance should succeed
-        let result = reader.advance(&stream_cx).unwrap().unwrap();
+        let result = reader.advance(&stream_cx, &yielded).unwrap().unwrap();
         assert_eq!(result.offset, 15);
         assert_eq!(result.part_number, 2);
         assert_eq!(result.part_size, 5);
@@ -638,7 +707,7 @@ mod test {
         }
 
         // Third advance should fail due to misaligned offset
-        let result = reader.advance(&stream_cx);
+        let result = reader.advance(&stream_cx, &yielded);
         assert!(result.is_err());
     }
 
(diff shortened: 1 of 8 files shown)


aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -207,38 +217,47 @@ impl UploadTransfer {
...
                     self.inner.ctx.try_wake();
                     return PollWork::Pending;
                 }
+
+                // An unknown-length upload has no part count to bound it, so guard the S3 limit
+                // here: without this the next dispatch would send `PartNumber = MAX_PARTS + 1` and
+                // take S3's `InvalidArgument`, which says nothing about the remedy. Failing in
+                // `poll_work` is safe — the transfer is terminal on return, so the `is_active()`
+                // check at the top of the next poll reports `Done`.
+                if matches!(plan, PartPlan::Unknown { .. }) && *parts_dispatched >= MAX_PARTS {
```
aajtodd: Sort of pre-existing and double check but `parts_dispatched` counts dispatches and end-of-stream on a custom part stream is only observed on the K+1 dispatch. An object using 10k parts hits EOS `Ok(None)` on the 10k+1 dispatch.

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -388,12 +428,54 @@ impl UploadTransfer {
...
                 return WorkOutcome::Success { data: None };
             }
             Err(e) => return self.fail(e.into()),
         };
 
+        // Claim part 1 now, at read time — NOT after the send completes. A part read is what
+        // establishes that the stream is non-empty, and the send that follows can take a long time.
+        // Recording it later would let a concurrently-dispatched read that hits end-of-stream see
+        // "nothing emitted" while this part is still in flight, and synthesize a second part 1.
+        self.mark_part_emitted();
```
aajtodd: sets `first_part_emitted` under a different state lock acquisition after `next_part()` which could interleave with `set_eof_and_claim_empty_first_part` to produce:

```
Task A (data): next_part -> Some(part1)   [releases reader mutex]
                             ...(preempted before locking state)...
Task B (EOF):  next_part -> None          [acquired reader mutex after A]
Task B:        set_eof_and_claim -> first_part_emitted==false -> CLAIMS empty part 1
Task A:        mark_part_emitted -> sets first_part_emitted=true  (too late)
```

----

I think maybe this and the previous comment on K+1 dispatches could be solved together by moving the state to where the reader is already serialized. 

Add `parts_yielded() -> u64` to the reader, incremented under the read lock as each `Some` is returned (`DynPartReader` lacks a count today; `Bytes`/`Fs` already carry a `part_number` cursor` so we'd unify it). That gives a bases for:
* **Empty detection:** on end-of-stream `parts_yielded() == 0` and TM can synthesize one empty part. 
* **Ceiling**: after each successful read, before `send_part`, fail if `parts_yielded() > MAX_PARTS`. A 10k part stream then isn't gated by `parts_dispatched`.

aws-sdk-s3-transfer-manager/src/operation/upload/context.rs
```diff
@@ -26,12 +51,18 @@ pub(crate) enum UploadState {
...
         response_builder: UploadOutputBuilder,
-        /// Full object content length, carried through to CompleteMPU as `MpuObjectSize`.
-        content_length: u64,
+        /// Full object size, carried through to CompleteMPU as `MpuObjectSize`.
+        ///
+        /// For [`PartPlan::Known`] this is the length declared by the source, set once at
+        /// `CreateMultipartUpload` time. For [`PartPlan::Unknown`] there is no declared length, so
+        /// it accumulates the byte length of each part actually uploaded and is exact only once
+        /// end-of-stream is reached — which is the point at which it is read.
+        object_size: u64,
```
aajtodd: This is semantically overloaded.

It is the declared length for `PartPlan::Known` and an accumulator for `PartPlan::Unknown`. Instead maybe we make it a consistent field as `bytes_uploaded` on both paths. Then move the declared length into the plan as `PartPlan::Known { total_parts, declared_object_size }`. 

* `Known` -> assert `bytes_uploaded == declared_object_size`, **send** `declared_object_size`
* `Unknown` -> send `bytes_uploaded`, no assertion

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
vcjana: Good catch. Moved the ceiling to read time: after a successful read, fail if `parts_yielded() > MAX_PARTS`. A legit 10k-part stream would no longer trip it.

aws-sdk-s3-transfer-manager/src/operation/upload/context.rs
vcjana: made it `<=` instead of `==`, `==` panics `test_large_upload_part_size_bump` and as it is a `debug_ assert`, S3's MpuObjectSize check is the real enforcement anyways <= catches an over count locally

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -428,53 +419,89 @@ impl UploadTransfer {
...
-                        "empty unknown-length stream; uploading a single empty part",
-                    );
-                    return self.send_part(PartData::new(1, Bytes::new())).await;
-                }
-                self.maybe_transition_to_completing();
-                return WorkOutcome::Success { data: None };
+                // Unknown-length end-of-stream. The reader's yielded-count is now final — it was
+                // incremented under the reader's own lock as each part was produced, so a `None`
+                // here is ordered after every part it will ever yield. That lets us decide, without
+                // a second racing lock, whether the stream was empty.
```
aajtodd: nit: No need to add project state/past history context here, comments like this can go stale and don't really explain more than the code does itself.

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -428,53 +419,89 @@ impl UploadTransfer {
...
 
-        // Claim part 1 now, at read time — NOT after the send completes. A part read is what
-        // establishes that the stream is non-empty, and the send that follows can take a long time.
-        // Recording it later would let a concurrently-dispatched read that hits end-of-stream see
-        // "nothing emitted" while this part is still in flight, and synthesize a second part 1.
-        self.mark_part_emitted();
+        // Guard the S3 part-count ceiling by the number of parts the reader has actually yielded,
+        // not by speculative dispatches. Fail before sending the (MAX_PARTS + 1)-th so the caller
+        // gets an error naming the remedy instead of S3's opaque `InvalidArgument` on part 10001.
+        if matches!(self.plan_kind(), Some(PartPlanKind::Unknown))
```
aajtodd: `PartPlanKind` seems unnecessary, you could just capture this under the existing lock taken at top of `execute_upload_part` without introducing a secondary enum

```rs

let (part_reader, is_unknown) = {
    let state = self.inner.state.lock().expect("lock poisoned");
    match &*state {
        UploadState::Transferring { part_reader, plan, .. } =>
            (part_reader.clone(), matches!(plan, PartPlan::Unknown)),
        _ => panic!("unexpected state for read_part"),
    }
};
// ...after the read:
if is_unknown && part_reader.parts_yielded() > MAX_PARTS { /* fail */ }
```

aws-sdk-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -89,6 +90,14 @@ impl Builder {
...
     inner: Inner,
     stream_cx: StreamContext,
+    /// Count of parts this reader has yielded (returned as `Ok(Some(..))`).
+    ///
+    /// Incremented in [`next_part`](Self::next_part) as each part is produced, so it is ordered with
+    /// the reads: end-of-stream (`Ok(None)`) is only observed after every yielded part has already
+    /// bumped this. That lets a caller driving speculative, concurrent reads decide two things
+    /// race-free — whether the stream was empty (`0` at end-of-stream) and whether it has exceeded a
+    /// part limit — without a second lock of its own that could interleave with the read.
+    parts_yielded: AtomicU64,
```
aajtodd: You could move this to `DynPartReader` state on the inner variant. It's only ever read currently on the unknown content length path. Threading the atomic through next_part + advance is a lot of plumbing and makes it feel tacked on. The buffer/fil variants already carry a `part_number` so you can still implement `parts_yielded() -> u64`

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -207,38 +225,33 @@ impl UploadTransfer {
...
-                            part_reader: Some(part_reader),
-                            completed_parts: Some(completed_parts),
-                            response_builder: Some(response_builder),
-                            complete_in_flight: false,
-                            content_length,
-                        };
-                    }
+                    transition_to_completing(&mut state);
                     drop(state);
                     self.inner.ctx.try_wake();
```
aajtodd: You didn't introduce this quirk but upon review it's a bit odd. 

We transition to `Completing`, then `try_wake()` and return `Pending`. But `try_wake()` only fires if a prior `set_pending()` armed the flag (which this branch doesn't do/observe) and wake (if it were to even fire) would only end up back in` poll_work()` which we are already in. 

We now have both `transition_to_completing` and `maybe_transition_to_completing` functions. 

Wondering if we want to maybe fold it into one mechanism:

```rust

/// All parts dispatched and none in flight → move Transferring → Completing. Returns whether it did.
fn try_begin_completing(state: &mut UploadState) -> bool { /* the current check + reshape */ }

// the part-completion event (was maybe_transition_to_completing):
fn on_part_completed(&self) {
    let mut state = lock;
    if let Transferring { parts_in_flight, .. } = &mut *state { *parts_in_flight -= 1; }
    if try_begin_completing(&mut state) { drop(state); self.inner.ctx.try_wake(); } // mutator: correct pairing
}
```


And make `poll_work` re-match after a transition instead of waking itself:

```rust
loop {
    match &mut *state {
        Transferring { .. } => {
            if try_begin_completing(&mut state) { continue; } // now Completing → falls through
            // ...dispatch next part...
        }
        Completing { complete_in_flight, .. } => {
            if *complete_in_flight { set_pending(); return Pending; }
            *complete_in_flight = true;
            return PollWork::ready(CompleteMPU);
        }
        // ...
    }
}

```

That gives one transition mechanism (`try_begin_completing`) and the two callers have distinct roles: `on_part_completed` is the mutator (decrement, maybe complete, wake), and `poll_work` loops/continues into `Completing` arm on the same poll as the transition and can emit `CompleteMPU` as `Ready` with the `complete_in_flight` guard in one place.

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -388,12 +419,90 @@ impl UploadTransfer {
...
+    /// Handle an unknown-length reader reaching end-of-stream: mark `eof`, and if the stream turned
+    /// out empty, synthesize the one zero-length part S3 requires (a multipart upload must list at
+    /// least one part).
+    ///
+    /// Emptiness is `part_reader.parts_yielded() == 0`, which is race-free: the count is bumped
+    /// under the reader's lock as each part is produced, so a `None` read is ordered after all of
+    /// them. Exactly one caller synthesizes, because only the caller that flips `eof` false→true —
+    /// a single-winner check under the state lock — takes the synthesize branch.
+    async fn on_end_of_stream(&self, part_reader: &PartReader) -> WorkOutcome {
+        let empty_and_owned = {
```
aajtodd, line 444: The old `set_eof_and_claim_empty_first_part` explicitly gated on `PartPlan::Known`. We now synthesize whenever `parts_yielded() == 0` which (technically) could allow a known part stream to yield a spurious empty part 1. Might be worth explicit gate here too?

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -1213,62 +1224,4 @@ mod tests {
...
-    /// The 10,000-part ceiling is enforced before dispatch, with an error that names the
-    /// remedy.
-    ///
-    /// An unknown-length upload has no part count to bound it, so without this guard the next
-    /// dispatch would send `PartNumber = 10001` and take S3's `InvalidArgument` — which says
-    /// nothing about part size. Reaching the ceiling through the public API would need 50 GiB
-    /// (part size is clamped to >= 5 MiB), so the plan state is driven directly here.
-    #[cfg_attr(miri, ignore)]
-    #[tokio::test]
-    async fn test_unknown_length_part_ceiling_is_enforced() {
```
aajtodd: Could consider replacing with a new test for the read-time guard. A custom `PartStream` yielding > `MAX_PARTS` tiny parts drives `parts_yielded` past the limit through the public API and would allow asserting the failure.

aajtodd (commented):
Couple more suggestions/discussion.

---

A general nit/comment on doc comments throughout. Several of the new comments document current implementation and iteration/history rather than behavior/invariants. I recommend shooting for "kernel-style" prose as a rule of thumb - say what the code guarantees and the non-obvious why, not how it currently does it or what it replaced. 

Smells:
* History / diff narration: "the racy emitted-flag it replaces", "the race this counter exists to close", "this is the whole reason the count lives on the reader", "now final". These describe the change, not the code; they go stale the moment someone refactors and read as commit-message text in the source.

* Restating the mechanism: if the comment tracks the code line-for-line ("increment here, under the lock"), it adds nothing the code doesn't. Keep the comment only where it states an invariant the code can't state itself.


**Examples** (not exhaustive)

**`context.rs`, `eof` field**
- Before: "…so it is a plain flag, not the racy emitted-flag it replaces. Emptiness is judged from the reader's `parts_yielded()`…"
- After: "Set once a read reports end-of-stream; gates further dispatch. Its timing is not correctness-critical."

**`part_reader.rs`, `next_part`**
- Before: "…would leave a window in which a concurrent reader observes end-of-stream with a stale count, which is the race this counter exists to close."
- After: "Increment inside the lock that produces the part, not here after it's released — otherwise a concurrent end-of-stream read could observe a stale count." *(keeps the why, drops the backstory)*

**`transfer.rs`, `on_end_of_stream` `Ok(None)` arm**
- The whole ordering paragraph duplicates the `parts_yielded` doc — drop it to a pointer: "Reader is at end-of-stream; `parts_yielded()` is final (see its docs for the ordering guarantee)."

aws-sdk-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -124,10 +142,22 @@ enum Inner {
...
+            Inner::Fs(path_body) => {
+                path_body
+                    .next_part(&self.stream_cx, &self.parts_yielded)
+                    .await
+            }
+            Inner::Dyn(part_stream) => {
+                part_stream
+                    .next_part(&self.stream_cx, &self.parts_yielded)
+                    .await
+            }
```
ysaito1001: I could be wrong, but by similarity to `ByteStream` in the smithy runtime, only `Dyn` can yield a stream of unknown length?

ysaito1001: Ah, aaron already [asked the same thing](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/164/changes#r3804270604) from a different angle.

aws-sdk-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -382,11 +431,23 @@ impl DynPartReader {
             inner: tokio::sync::Mutex::new(inner),
         }
     }
-    async fn next_part(&self, stream_cx: &StreamContext) -> Result<Option<PartData>, Error> {
+    async fn next_part(
+        &self,
+        stream_cx: &StreamContext,
+        parts_yielded: &AtomicU64,
+    ) -> Result<Option<PartData>, Error> {
         // TODO - can we do better than a mutex here? should we spawn a dedicated task and use channels instead
```
ysaito1001, line 423: Not blocking, but just an idea for a possible follow-up based on this `TODO`.
  
Most of the new bookkeeping (the `parts_yielded` atomic, the `eof` check-and-set in `on_end_of_stream`, `PartPlanKind`) is to keep concurrent speculative reads race-free. But the unknown path doesn't seem read concurrently; `DynPartReader::next_part` holds its mutex across `stream.next().await`, so the reads are already
  serialized for `Dyn`.
  
If the unknown path had its own reader task, looping `next_part()` into a bounded channel and closing on EOF, the reader becomes the only place that sees end-of-stream. "Was it empty" is then just `yielded == 0` in the task, the empty-part and 10k-part checks are local, `eof` becomes "channel closed", and `poll_work` pulls with `try_recv`. I think that lets us drop the atomic (plus the plumbing threading it through all three readers), `on_end_of_stream`, `PartPlanKind`, and the check-and-set.

I haven't built it, so may not work as described; the `Empty` → `set_pending` → reader-wake handoff is the bit I feel worth proving out. Curious whether you already considered that option.

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
vcjana: Dropped it

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
vcjana: now captured under the existing lock

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
vcjana: `try_begin_completing` is the only transition,` poll_work` loops and continues into Completing on the same poll so the `self-try_wake` is gone, and` on_part_completed` is the mutator side. Also pulled the duplicated `all_dispatched `check onto `PartPlan`.

Only cost is `try_begin_completing` checking readiness with `matches!` then re-destructuring in the `mem::replace`.

aws-sdk-s3-transfer-manager/src/io/part_reader.rs
vcjana: Moved to `DynPartReader`plumbing is gone, `next_part` and advance back to their old signatures.

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
vcjana: Good catch, gated on `Unknown` now - the old gate came free because `eof` lived inside `PartPlan::Unknown`, so matching the variant was the check, and moving `eof` to transferring dropped it silently. Added a test: source declares 5 MiB, delivers nothing, uploads a spurious empty part without the gate.

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
vcjana: Done, custom `PartStream` yielding 1-byte parts drives `parts_yielded` past the limit through the public API. 10,001 fails with the part-size remedy in the message, and added the exactly-10,000 case too so the off-by-one is pinned

aws-sdk-s3-transfer-manager/src/io/part_reader.rs
vcjana: `Buf` and `Fs` both build `SizeHint::exact `in `RawInputStream::size_hint`, so `Dyn` is the only variant that can be unknown-length. Counter's on `DynPartReader` now and the threading is gone

aws-sdk-s3-transfer-manager/src/io/part_reader.rs
aajtodd: This comment is quite old at this point. I don't know that I love the idea of owning another internal queue essentially and sizing it. We also will have to revisit dyn part reader and how it's integrated into upload, right now it can deadlock: https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/v0.2.0/aws-sdk-s3-transfer-manager/tests/upload_test.rs#L130

Agreed not this PR

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -177,91 +187,86 @@ impl UploadTransfer {
...
-                stream,
-            } => {
-                if *init_in_flight {
-                    self.inner.ctx.set_pending();
-                    return PollWork::Pending;
-                }
+        loop {
+            // Serve the `CompleteMPU` a transition unlocks on this same poll, rather than parking
+            // and waiting to be re-polled for work this call already knows about.
+            if try_begin_completing(&mut state) {
```
aajtodd: Might consider folding this into the `Transferring` arm since it only applies there. It drops the per-poll redundant check on the other four states and localizes the completion logic next to the dispatch logic it's a part of.

Something like:

```rust
Transferring { parts_dispatched, plan, eof, parts_in_flight, .. } => {
    if plan.all_dispatched(*parts_dispatched, *eof) {
        // Last one out: complete now and serve CompleteMPU on this poll.
        if try_begin_completing(&mut state) { continue; }
        // Otherwise parts are still in flight; on_part_completed will wake us.
        self.inner.ctx.set_pending();
        return PollWork::Pending;
    }
    *parts_dispatched += 1; *parts_in_flight += 1;
    return PollWork::ready(/* UploadPart */);
}

```

aajtodd (approved):
Nothing blocking. One minor suggestion if you take it. 

Are you merging this to `main` or `ga-ramp`? I'd prefer `main` unless there's a reason not to?

PR #165, opened 2026-08-17
memory pool design

aajtodd:
## Summary

This PR replaces `docs/design/memory-budget.md` with the design for one buffer pool shared by transfer admission, upload and download payloads, retry bodies, and pool-enabled HTTP receive.

The design covers admission, physical storage, byte ownership, reclamation, failure handling, integration boundaries, and verification. This PR does not add the production implementation.

[Rendered](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/bp-design/docs/design/memory.md)

## TL;DR

One `BufferPool` defines one accounting and physical-storage domain. Admission happens in `poll_work()` before dispatch; physical acquisition happens later in `execute(work)`.

```mermaid
flowchart TD
    PW["poll_work()"] --> RF["poll ReserveFuture from BufferPool::reserve(envelope)"]
    RF --> RES["Reservation"]
    RES --> READY["PollWork::Ready { io }"]
    READY --> DISPATCH["dispatch"]
    DISPATCH --> EXEC["execute(work)"]
    EXEC --> ACQ["BufferPool::acquire(&amp;reservation, min_bytes)"]
    HTTP["Zero-copy HTTP receive without a Reservation"] --> UNRES["BufferPool::acquire_unreserved(min_bytes)"]
    ACQ --> B["PooledBufMut"]
    UNRES --> B
    B --> P["Publish initialized ranges"]
    P --> O["Bytes / SegmentedBytes"]
    O --> F["Final owner drops"]
    F --> RC["Return physical carrier, then release charge"]
```

`poll_work()` admits the work before dispatch. `PollWork::Ready { io }` carries the `Reservation` into the dispatched work, and `execute(work)` acquires writable memory from it.

The unreserved branch is only for future pool-backed zero-copy HTTP receive where a `Reservation` cannot reach the transport. The baseline HTTP integration stays on the reserved branch: `execute(work)` copies foreign transport bytes into pooled staging.

The pool uses one fixed carrier size. Each carrier is page-aligned, occupies a whole number of pages, and shares no page with another carrier. Larger logical payloads use carrier chains rather than requiring contiguous memory.

```text
BlockSlot: stable virtual range

+----------------------------------------------------------------+
|                            block 0                             |
+----------------+----------------+-----+-------------------------+
|   carrier 0    |   carrier 1    | ... |       carrier N        |
+----------------+----------------+-----+-------------------------+

current -> ArcSwapOption<BlockIncarnation>
              +-- state: Active | Draining | Dead
              `-- occupancy bitmap

trim:   current A -> None; discard physical backing
revive: None -> current B with a fresh bitmap
```

A block groups carriers for mapping and reclamation. Its virtual range remains reserved for the pool lifetime. Trim removes only an all-free block; revival makes the same range accessible with a fresh incarnation so stale claim rollback cannot mutate the new bitmap.

Configured capacity limits normal admission. It is not a hard RSS, mapped-address, or allocation limit. Unreserved acquisition can exceed it, but that ownership remains accounted and delays later managed work until the bytes are released.

Most callers need no memory configuration:

```rust
let automatic = TransferManager::builder().build()?; // Same as MemoryConfig::Auto.

// Advanced: share one pool with another bounded component.
let pool = BufferPool::builder().build()?;
let shared = TransferManager::builder()
    .memory(MemoryConfig::Explicit(pool.clone()))
    .build()?;

// The other component enters the same admission and storage domain.
let reservation = pool.reserve(bytes).await?;
let buffer = pool.acquire(&reservation, min_bytes)?;
```

Explicit configuration gives the caller a retained pool handle. The transfer manager and other component then share configured capacity, FIFO admission, prepared storage, and ownership accounting.

## Why

The existing `MemoryBudget` admits fixed-size logical chunks for downloads but does not own or reuse their storage. Upload staging, retry bodies, retained download output, and HTTP receive do not share one ownership and accounting model. Independent limits either leave memory unaccounted or multiply beyond the intended process budget under concurrency.

The replacement needs to:

- admit planned memory before work occupies execution capacity;
- allocate for transport code that cannot carry a reservation through the HTTP stack;
- keep memory charged while cloned or sliced byte views remain alive;
- reuse page-granular storage without retaining peak physical backing;
- preserve FIFO admission and progress for requests larger than configured capacity;
- reclaim free blocks without racing acquisition or final byte-owner return.

## What to take away

- A reservation admits a planned envelope. It does not own carriers or predict whether reserved or unreserved acquisition will consume the envelope.
- A carrier is the unit of physical ownership and accounting. A block is the larger unit of mapping, growth, protection, and reclamation.
- Reserved and unreserved acquisition consume the same fungible aggregate coverage. Ownership beyond that coverage becomes an uncovered charge. A later grant does not erase it.
- Acquisition never waits for another carrier owner to return memory. It reuses compatible capacity, exhaustively checks existing capacity on the serialized path, grows, or returns an `AcquireError`.
- One `CarrierGuard` owns one carrier return and one accounting return. `Bytes` clones and slices share the guard. The final guard owner returns the carrier before releasing its charge.
- `PooledBufMut` represents one growable mutable allocation stream. Growth reuses its writable tail before acquiring more carriers. Independent buffers remain independently rounded.
- `SegmentedBytes` is the public immutable delivery type. Converting a multi-segment value to one `Bytes` may copy.
- Reclamation is block-local and fail-closed. A failed mapping or discard operation cannot make uncertain storage claimable and does not stop unrelated blocks from growing or reclaiming.

## Design shape

Admission and physical allocation are separate transactions. `BufferPool::reserve()` returns a `ReserveFuture`; polling it grants a `Reservation`, waits in one FIFO slot, or returns an error. A grant prepares enough fungible capacity for the admitted total before publishing the reservation. Closing the reservation removes unused planned demand but does not release carriers retained by mutable or immutable owners.

`BufferPool::acquire()` spends direct-acquisition authority. `BufferPool::acquire_unreserved()` has no reservation argument. Both install aggregate charges before exposing writable memory. The common path performs bounded optimistic reuse. A miss enters serialized fallback, checks all compatible prepared storage, and only then grows. Fresh growth claims the requesting batch before publishing the remaining free capacity.

The baseline HTTP integration copies foreign transport `Bytes` into reserved pooled staging. An optional writable-storage provider can instead place decoded payload directly into pooled storage through unreserved acquisition.

Acquired carriers become `PooledBufMut`. `PooledBufMut::reserve()` extends that buffer by consuming its existing tail before acquiring the rounded shortfall. Closing the public reservation revokes growth that needs new carriers; existing tails and initialized bytes remain valid. Publishing initialized prefixes creates owner-backed `Bytes` views or `SegmentedBytes` without changing the charge. Carrier return clears physical ownership before accounting release so admission cannot credit storage that remains unavailable.

Blocks retain stable virtual addresses. Trim publishes `Draining` before checking the ownership bitmap; claim sets its bit before checking block state. This gate prevents claim and trim from both succeeding. Revival publishes a fresh bitmap instead of resetting one that a stale claimant may still access.

## Changes

- Adds `docs/design/memory.md`.
- Removes `docs/design/memory-budget.md`, whose logical download budget is superseded by the combined admission, storage, and ownership design.
- Updates the scheduler and paged receive-buffer designs for reservation dispatch and the new admission-progress model.

## Scope

This PR does not:

- implement or benchmark the production pool;
- select final carrier size, block size, scan budget, or reclaim timing;
- define the upstream Hyper/H2 writable-buffer API;
- require private protocol scratch to use pooled storage.

Implementation and benchmarking may change private types and measured constants. A change to accounting, ownership, progress, failure, or safety guarantees requires a corresponding design update.

## Future

Implementation is scoped as seven PRs targeting `buffer-pool-main`: physical-storage foundation, arena acquisition, admission and pool composition, byte ownership and delivery, maintenance and operations, download integration, and upload integration. Benchmark selection and production hardening are series exit work rather than a standalone PR.

Optional zero-copy Hyper/H2 receive ownership remains a separate upstream workstream. The baseline copy path does not depend on it. The transfer-manager adapter follows after the pool acquisition and owner contracts stabilize and the upstream lifecycle covers both H1 and H2 terminal paths.


docs/design/scheduler.md
```diff
@@ -285,8 +285,12 @@ the gap is exhausted, the download returns `Pending`. When the consumer reads da
 the read head and wakes the transfer. The gap scales with concurrency to avoid becoming a
 throughput bottleneck at high concurrency while still bounding memory from out-of-order completion.
 
-**Buffer pool.** Transfers acquire buffer tickets before generating work. When the memory budget is
-exhausted, transfers return `Pending`. When buffers are released, blocked transfers are woken.
+**Buffer pool.** A transfer polls a `ReserveFuture` for the work item's planned memory envelope before
```
ysaito1001, line 288: Occurred to me while trying to build a mental model. The scheduler is CFS with vruntime-based priority but pool admission is FIFO by arrival, so can a high-priority transfer end up waiting for memory behind an earlier lower-priority one (priority inversion at pool admission)?

aajtodd: Yes correct. Which I believe is the desired behavior. CFS scheduling respects priority of a transfer's ability to generate additional work (i.e. consume additional execution time). An earlier lower priority transfer that is polled and blocked on memory acquisition should be granted execution time before a higher priority transfer polled later that also blocks on memory.

ysaito1001 (approved):

PR #166, opened 2026-08-19
buffer pool: physical foundation

aajtodd:
## Summary

This is the first of several implementation PRs for the [buffer-pool design](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/165). It adds the physical-storage foundation: checked geometry, stable virtual ranges, fixed-size carrier ownership, and the claim/trim protocol.

The new module remains private and disconnected from transfer-manager operation. This PR is stacked on #165 so the implementation and design corrections can be reviewed together. After #165 merges, this branch will be rebased onto `buffer-pool-main` and retargeted before merge.

## Why

The transfer manager needs reusable storage that can serve variable-size I/O without retaining its peak resident set. The pool reserves stable virtual ranges, prepares their backing storage on demand, subdivides each range into one carrier size, and later removes backing storage while retaining the virtual address range.

That lifecycle creates a claim/trim race. A claimant must not publish ownership from an incarnation that trim has made inaccessible, and recovery from a failed deactivation must preserve ownership already recorded in that incarnation. This PR implements those physical invariants before adding arena growth, admission accounting, or public buffer types.

### Physical layout

A block slot owns one stable virtual range. Its current block incarnation owns the prepared mapping state and a bitmap with one ownership bit per carrier.

```text
BlockSlot
+-- VirtualRange: fixed base and length, reserved until drop
|   +-------------+-------------+-----+-------------+
|   |  carrier 0  |  carrier 1  | ... |  carrier N  |
|   +-------------+-------------+-----+-------------+
|
`-- current -> BlockIncarnation A
    +-- state: Active | Draining | Dead
    `-- in_use: one ownership bit per carrier
```

Checked geometry establishes the page, carrier, block, and bitmap relationships before virtual memory is reserved. The platform layer then provides the same reserve, prepare, deactivate, and release contract on Linux, macOS, and Windows.

See [Block geometry and lifetime](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/main/docs/design/memory.md#block-geometry-and-lifetime) and [Platform contract](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/main/docs/design/memory.md#platform-contract).

### Claim and return

A bitmap claim is provisional until its incarnation gate passes. Successful conversion produces a `CarrierAllocation`; dropping the final owner clears the carrier bit.

```text
free bitmap bit
      |
      v
BlockClaimAttempt
      |
      | Active gate passes
      v
ProvisionalBits
      |
      | ownership conversion
      v
[...]

PR #167, opened 2026-08-26
buffer pool: arena acquisition

aajtodd:
## Summary

This is the second of seven implementation PRs for the [buffer-pool design](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/165). It composes the physical block primitive from [#166](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/166) into a private arena with immutable registry publication, bounded optimistic batch acquisition, serialized fallback and growth, address classification, trim candidate selection, and internal diagnostics.

The design and implementation series are:

1. [Buffer-pool design](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/165)
2. [Physical-storage foundation](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/166)
3. **This PR: Arena acquisition and growth**
4. Admission and pool composition
5. Mutable buffers and segmented delivery
6. Maintenance, configuration, and operations
7. Scheduler and download integration
8. Upload staging and retry integration

The arena remains private and disconnected from transfer-manager operation. This PR is stacked on #166.

## Why

The block primitive provides safe ownership and reclamation within one fixed virtual range. The pool also needs to find carriers across many blocks, return complete multi-block batches, reuse fragmented capacity before growing, and publish new capacity without allowing another claimant to steal the carriers that caused the growth.

The common acquisition path must remain bounded as the pool grows. A bounded scan can miss reusable capacity, so a miss retains its provisional ownership and enters a serialized path that exhaustively rechecks the arena before preparing another block. New incarnations preclaim the missing carriers before publication. This prevents repeated growth under contention while preserving lock-free claims against already published blocks.

This PR establishes that physical allocator contract before admission accounting, reservations, byte containers, maintenance policy, or public APIs are added.

### Physical model

The arena coordinates blocks but does not expose byte storage. A `BlockSlot` retains one stable virtual range for its lifetime. Its replaceable incarnation contains the current ownership bitmap, where one set bit represents one owned carrier.

```text
Arena
`-- BlockSlot[]
    |-- stable VirtualRange
    |-- serialized MappingState
    `-- current BlockIncarnation
        `-- in_use[]: one atomic bit per carrier
```

Mapping state remains mutex-protected while the incarnation gate and bitmap remain atomic. This keeps carrier claims off the mapping lock. The legal lifecycle combinations are explicit:

```text
Reserved                     + no current incarnation
Prepared                     + Active or trim-owned Draining incarnation
ActivationRecoveryPending    + no current incarnation
DeactivationRecoveryPending  + Draining incarnation
```

### Registry

`ArenaState` owns the stable slot list and serializes topology changes. Each change rebuilds and atomically publishes one immutable `RegistryGeneration`.

```text
ArenaState
+-- slots: [BlockSlot 0, BlockSlot 1, ...]
|
| rebuild and publish
v
RegistryGeneration
+-- slots_in_claim_order: [slot 0, slot 1, ...]
`-- address_ranges:       [start..end -> slot, ...]  sorted by address
```

The generation contains topology, not a frozen copy of mapping or bitmap state. Those remain live behind each `Arc<BlockSlot>`. Acquisition and address classification load the registry without taking `ArenaState`. Claim order remains stable, while the sorted address index supports binary-search classification of one complete range within one slot. Both indexes publish together because every address entry indexes `slots_in_claim_order`. Classification compares integer addresses only; it does not construct a pointer or authorize memory access.

The corresponding sections in design PR #165 are **Pool structure** and **Block geometry and lifetime**.

### Acquisition

One arena acquisition requests a complete carrier count. `claim_optimistic` scans from a rotating origin and charges an exact bitmap-word budget. It retains every gate-passed bit in `ClaimBatch`.

```text
request N carriers
        |
        | load RegistryGeneration
        | topology is pinned; block state remains live
        v
bounded optimistic scan                         no arena mutex
        |
        v
BlockClaimAttempt
        |
        | fetch_or wins bitmap bits
        | finish() checks Active
        +---------------- gate fails ----------------> clear won bits
        |
        v
ProvisionalBits
        |
        v
ClaimBatch
        |
        +-- complete -------------------------------> CarrierAllocation[N]
        |
        `-- partial
              |
              | complete_claim_serialized()          ArenaState held
              |
              +-- exhaust active blocks
              +-- prepare inactive slots
              `-- reserve new slot
                       |
                       | create private incarnation
                       | preclaim missing carriers
                       | prepare complete mapping
                       | add prepared capacity
                       `-- publish Active
              |
              v
          complete ClaimBatch
              |
              | release ArenaState
              v
          CarrierAllocation[N]
```

| Stage                     | Owner of each set bit     | Protection carried by the owner                                             |
| ------------------------- | ------------------------- | --------------------------------------------------------------------------- |
| Before the state gate     | `BlockClaimAttempt`       | The protected incarnation keeps its bitmap valid for success or rollback    |
| After the state gate      | `ProvisionalBits`         | The gate-passed live bit forces trim to abandon                             |
| Across several blocks     | `ClaimBatch`              | Each provisional owner retains its slot, incarnation, and exact won mask    |
| After complete conversion | `CarrierAllocation`       | The allocation owns one live bit until drop                                 |
| Failed acquisition        | The remaining owner drops | Only bits won by that owner are cleared                                     |

A complete batch converts into one `CarrierAllocation` per won bit. An incomplete batch cannot be converted into `CarrierAllocation`s. Dropping it clears only the bits won by that batch.

> **Serialization boundary:** Optimistic scan and claim do not take `ArenaState`. Exhaustive fallback and growth hold `ArenaState`. This PR passes prepared-capacity state as `&mut CarrierCount`; the next PR replaces that parameter with `&mut AdmissionGuard<'_>`, making pool-wide admission serialization part of the type signature.

The bounded scan may miss free carriers outside its word budget or lose selected bits to a concurrent claimant. Serialized fallback uses an exhaustive block claim that folds each collision into its local occupancy view and continues until the request is complete or every valid bit has been observed occupied.

### Fallback and growth

Serialized fallback holds `ArenaState` while it rechecks every active slot, reuses inactive slots, and reserves new stable ranges. It prepares new capacity with the missing carriers already owned.

The slot may already be visible in a registry generation, but it is not claimable until the preclaimed incarnation is published. Optimistic claimants therefore cannot consume the new capacity before the fallback claimant receives its complete batch. Preparation failure records `ActivationRecoveryPending` without publishing an incarnation. A failed trim deactivation records `DeactivationRecoveryPending` while retaining the draining incarnation, so the recovery cause is explicit rather than inferred solely from incarnation presence.

Preparation failure preserves blocks prepared earlier in the same fallback. The caller retains the partial `ClaimBatch`, so rollback remains exact.

The corresponding section in design PR #165 is **Carrier claiming and fallback**.

### Trim selection and diagnostics

For reclamation, the arena scans for a prepared block whose carrier bitmap is currently empty. That lock-free observation identifies only a candidate. `BlockSlot::start_trim` remains the authority that rechecks lifecycle state, ownership bits, and the prepared-capacity floor under admission serialization.

Private saturating counters record optimistic scan work and misses, serialized fallback, block preparation, virtual-range growth, unpublished claim rollback, and trim-selection work. They do not participate in allocator decisions and are not a public metrics surface.

## How to review

Follow the acquisition diagram once through optimistic success using `one_batch_accumulates_carriers_across_blocks`, then through the partial branch using `serialized_fallback_exhausts_fragmented_capacity_before_growth`. Stop growth after one successful preparation with `later_preparation_failure_preserves_earlier_capacity` to see rollback preserve prepared capacity and return provisional ownership.

Finish with the concurrency models beside the transitions they exercise: `registry_generation_is_published_coherently` for publication, `concurrent_claims_use_distinct_rotated_windows` for scan distribution, `exhaustive_claim_retries_after_an_optimistic_collision` for collision recovery, `private_growth_cannot_be_stolen_by_an_optimistic_claim` for preclaim publication, and `claim_and_trim_cannot_both_pass_the_gate` for reclamation exclusion. `production_registry_guard_survives_concurrent_publication` separately exercises the production ArcSwap adapter because Loom substitutes its own registry cell.

## Future

The next PR joins the arena to admission accounting, reservations, FIFO waiters, and the private `BufferPool` core. It will replace caller-supplied prepared state with the admission guard and prove complete acquisition, rollback, close, return, and shutdown through the composed pool boundary.

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/runtime/buffer_pool/arena.rs
+++ b/aws-sdk-s3-transfer-manager/src/runtime/buffer_pool/arena.rs
@@ -1,0 +1,1695 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Stable block registry and physical-allocation coordination.
+
+use std::collections::TryReserveError;
+use std::fmt;
+use std::sync::atomic::{AtomicU64, Ordering as DiagnosticOrdering};
+
+use super::block::{BlockError, BlockSlot, CarrierAllocation, ProvisionalBits};
+use super::geometry::PoolGeometry;
+use super::CarrierCount;
+use crate::runtime::sync::sync::atomic::{AtomicUsize, Ordering};
+use crate::runtime::sync::sync::{Arc, Mutex};
+
+/// Failure to configure, claim, reserve, or publish arena storage.
+#[derive(Debug)]
+pub(super) enum ArenaError {
+    /// A block operation failed.
+    Block(BlockError),
+    /// Registry metadata could not be allocated.
+    Allocation(TryReserveError),
+    /// The optimistic bitmap-word budget is zero.
+    InvalidScanBudget,
+    /// A batch claim requested no carriers.
+    InvalidClaimCount,
+    /// The flattened registry scan space cannot be represented.
+    ScanSpaceOverflow {
+        /// Slots in the registry generation.
+        slots: usize,
+        /// Bitmap words in each slot.
+        words_per_slot: usize,
+    },
+    /// Conversion was requested before the complete batch was owned.
+    IncompleteClaim {
+        /// Carriers required by the batch.
+        required: usize,
+        /// Carriers provisionally owned by the batch.
+        claimed: usize,
+    },
+    /// No stable block identifier remains.
+    SlotIdExhausted,
+    /// The registry cannot represent another slot.
+    RegistryCapacityOverflow,
+    /// A stable virtual range overflows address representation.
+    AddressOverflow {
+        /// Slot containing the invalid range.
+        slot_id: u32,
+        /// First byte of the range.
+        start: usize,
+        /// Reserved range length.
+        len: usize,
+    },
+    /// Two stable virtual ranges overlap.
+    AddressOverlap {
+        /// First overlapping range start.
+        first_start: usize,
+        /// First overlapping range end.
+        first_end: usize,
+        /// Second overlapping range start.
+        second_start: usize,
+        /// Second overlapping range end.
+        second_end: usize,
+    },
+}
+
+impl fmt::Display for ArenaError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Block(error) => error.fmt(f),
+            Self::Allocation(error) => write!(f, "arena metadata allocation failed: {error}"),
+            Self::InvalidScanBudget => f.write_str("arena optimistic scan budget must be nonzero"),
+            Self::InvalidClaimCount => f.write_str("an arena claim must request carriers"),
+            Self::ScanSpaceOverflow {
+                slots,
+                words_per_slot,
+            } => write!(
+                f,
+                "arena scan space {slots} slots by {words_per_slot} words overflows"
+            ),
+            Self::IncompleteClaim { required, claimed } => write!(
+                f,
+                "cannot finish incomplete arena claim: required {required}, claimed {claimed}"
+            ),
+            Self::SlotIdExhausted => f.write_str("arena block identifier space is exhausted"),
+            Self::RegistryCapacityOverflow => f.write_str("arena registry capacity would overflow"),
+            Self::AddressOverflow {
+                slot_id,
+                start,
+                len,
+            } => write!(
+                f,
+                "block slot {slot_id} address range {start:#x}+{len:#x} overflows"
+            ),
+            Self::AddressOverlap {
+                first_start,
+                first_end,
+                second_start,
+                second_end,
+            } => write!(
+                f,
+                "block ranges {first_start:#x}..{first_end:#x} and \
+                 {second_start:#x}..{second_end:#x} overlap"
+            ),
+        }
+    }
+}
+
+impl std::error::Error for ArenaError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            Self::Block(error) => Some(error),
+            Self::Allocation(error) => Some(error),
+            Self::InvalidScanBudget
+            | Self::InvalidClaimCount
+            | Self::ScanSpaceOverflow { .. }
+            | Self::IncompleteClaim { .. }
+            | Self::SlotIdExhausted
+            | Self::RegistryCapacityOverflow
+            | Self::AddressOverflow { .. }
+            | Self::AddressOverlap { .. } => None,
+        }
+    }
+}
+
+impl From<BlockError> for ArenaError {
+    fn from(error: BlockError) -> Self {
+        Self::Block(error)
+    }
+}
+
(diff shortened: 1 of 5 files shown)


aws-sdk-s3-transfer-manager/src/runtime/buffer_pool/arena.rs
```diff
@@ -0,0 +1,1806 @@
...
+                    words_per_slot,
+                })?;
+        let mut batch = ClaimBatch::new(required, Arc::clone(&self.diagnostics));
+        if total_positions == 0 {
+            self.diagnostics.record_optimistic_scan(0, true);
+            return Ok(batch);
+        }
+
+        let budget = self.optimistic_scan_words.min(total_positions);
+        let origin = self.scan_origin.fetch_add(budget, Ordering::Relaxed) % total_positions;
```
ysaito1001, line 149: If I read it correctly, `budget` is clamped to `total_positions`, so once `optimistic_scan_words` reaches `total_positions` every call advances `scan_origin` by exactly one full lap of the scan space, and `%
  total_positions` gives 0 every time.

That seems to matter because the rotation is what keeps concurrent claimants off the same bitmap words, and with a frozen origin they all start on word 0 and `take_lowest` has them select the same bits?

<details>
<summary>Failing test</summary>

```
#[test]
fn scan_origin_rotates_regardless_of_budget() { 
    let distinct_words = |scan_words: usize| { 
        let arena = Arena::new(geometry_with_carriers(130), scan_words).unwrap();
        let slot = prepare_slots(&arena, 1).pop().unwrap();
        let words = (0..3)
            .map(|_| { 
                let claimed = arena
                    .claim_optimistic(CarrierCount::new(1))
                    .unwrap()
                    .finish()
                    .unwrap();
                let word = claimed[0].carrier_index() / 64;
                drop(claimed);
                word
            })
            .collect::<std::collections::BTreeSet<_>>();
        assert_fully_free(&slot);
        words.len()
    };

    // 130 carriers is 3 bitmap words, ceil(130 / 64) = 3, so one slot gives total_positions == 3.
    assert_eq!(distinct_words(2), 3, "budget below the scan space rotates");
    assert_eq!(distinct_words(3), 3, "budget covering the scan space rotates");
}
```

The test claims one carrier three times in a row and records which bitmap word each claim came from. Since the rotation is meant to move the starting point of the scan, the three claims should come from three different words.
  
With budget 3 the starting point never moves, so all three claims begin at word 0. Each claim is released before the next one runs, which frees the carrier it just took, so the next claim finds that same carrier free and takes it again. All three come back as carrier 0. That is one word, not three.
  
With budget 2 the starting point does move, through words 0, 2, then 1, and the claims come back as carriers 0, 128, and 64.

</details>

Advancing by `inspected_words + 1`, or by 1 when `budget == total_positions`, should do it.

ysaito1001 (approved):
LGTM, with one finding worth checking.

aws-sdk-s3-transfer-manager/src/runtime/buffer_pool/arena.rs
aajtodd: fde44b9

PR #168, opened 2026-08-26
user agent: extend instead of replacing

vcjana:
## Summary

Our interceptor stored a fresh `AwsUserAgent` over whatever was already in the config bag, so the customer's `app/<id>` was dropped on every transfer-manager request. It now extends what it finds, and copies the `AppName` itself — the SDK's own interceptor runs later and yields to any value already present.

Assembly moves from `config/loader.rs` to `config/user_agent.rs` and installs from `Client::new` on both client sources; previously only `from_env().load()` got an interceptor. A caller's own client is rebuilt to carry it, and left alone if it already has one. Adds `md/rust-tm#<version>`, since neither `aws-sdk-rust/…` nor `api/s3/…` says which transfer manager is running.

Not here: runtime mode or resolved config, and framework metadata stays `#[doc(hidden)]`.

## Test plan

- [x] fmt / clippy clean; 642 lib + 94 integration + 11 doc tests green
- [x] Mutation-checked each part — removing the extend, the `AppName` copy, the rebuild, or the dedup guard each reds a specific test

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/config/user_agent.rs
+++ b/aws-sdk-s3-transfer-manager/src/config/user_agent.rs
@@ -10,9 +10,10 @@ use aws_runtime::user_agent::{AdditionalMetadata, ApiMetadata, AwsUserAgent, Fra
 use aws_sdk_s3::config::{AppName, Intercept, IntoShared};
 use aws_types::os_shim_internal::Env;
 
-/// This crate's identity in the user agent. Neither `aws-sdk-rust/…` nor `api/s3/…` carries
-/// the transfer manager's version, and `rust-tm` distinguishes it from the CRT-backed one.
-const TM_METADATA: &str = concat!("rust-tm#", env!("CARGO_PKG_VERSION"));
+/// This crate's name and version, as the `md/` pair the user agent carries. Neither
+/// `aws-sdk-rust/…` nor `api/s3/…` carries the transfer manager's version, and `rust-tm`
+/// distinguishes it from the CRT-backed one.
+const RUST_TM_VERSION: &str = concat!("rust-tm#", env!("CARGO_PKG_VERSION"));
 
 /// Install the transfer manager's user agent attribution on an S3 client builder.
 ///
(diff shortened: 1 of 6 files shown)


aws-sdk-s3-transfer-manager/src/config/user_agent.rs
```diff
@@ -0,0 +1,339 @@
...
+//! User agent attribution for requests the transfer manager issues.
+
+use aws_runtime::sdk_feature::AwsSdkFeature;
+use aws_runtime::user_agent::{AdditionalMetadata, ApiMetadata, AwsUserAgent, FrameworkMetadata};
+use aws_sdk_s3::config::{AppName, Intercept, IntoShared};
+use aws_types::os_shim_internal::Env;
+
+/// This crate's identity in the user agent. Neither `aws-sdk-rust/…` nor `api/s3/…` carries
+/// the transfer manager's version, and `rust-tm` distinguishes it from the CRT-backed one.
+const TM_METADATA: &str = concat!("rust-tm#", env!("CARGO_PKG_VERSION"));
```
aajtodd: nit: `const RUST_TM_VERSION` or `TM_PACKAGE_VERSION` or something slightly more descriptive

aws-sdk-s3-transfer-manager/src/config/user_agent.rs
```diff
@@ -0,0 +1,339 @@
...
+        );
+    }
+
+    /// A client the caller hands over is used as built, and interceptors cannot be pushed onto
+    /// a finished client — so this path carries none of our attribution. Asserted rather than
+    /// left implicit, because the sections are absent by construction and a reader has no way
+    /// to tell that from an oversight.
+    #[cfg_attr(miri, ignore)]
+    #[tokio::test]
+    async fn provided_client_carries_no_attribution() {
```
aajtodd: Would be nice if we could get attribution on both paths but this may be ok for now.

aajtodd (approved):
Looks good overall, minor nits, nothing blocking.

aajtodd:
@vcjana One last thought. We need minor updates in aws-runtime to make AwsUserAgent possible to instantiate/re-use right? (IIRC you said some of the APIs are internal/private and we need to expose more to actually be able to re-use it)

vcjana:
actually it's less than i thought - the constructor and the `with_*` methods are already public only, so they can build one today itself. just the leading token is locked, `SdkMetadata` has no setter.

aws-sdk-s3-transfer-manager/src/config/user_agent.rs
vcjana: renamed to RUST_TM_VERSION in 6ddbcad

aws-sdk-s3-transfer-manager/src/config/user_agent.rs
vcjana: done in ec55a47. rebuilt the caller's client from `config().to_builder()` and skip one that already carries the interceptor, otherwise `download_objects` doubles the `md/rust-tm`.

PR #169, opened 2026-08-27
buffer pool: admission and pool composition

aajtodd:
## Summary

This is the third implementation PR in the eight-PR [buffer-pool design and implementation series](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/165). It composes the arena from [#167](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/167) into a `BufferPool` with planned-demand admission, cancellation-safe FIFO reservations, reserved and unreserved carrier acquisition, physical-first return, and coherent memory metrics.

The series is:

1. [Buffer-pool design](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/165)
2. [Physical-storage foundation](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/166)
3. [Arena acquisition and growth](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/167)
4. **This PR: Admission and pool composition**
5. Mutable buffers and segmented delivery
6. Maintenance, configuration, and operations
7. Scheduler and download integration
8. Upload staging and retry integration

This PR is stacked on #167 and does not change transfer-manager behavior.

## Why

The arena can return physical carriers, but it does not decide which work may consume memory or keep enough capacity prepared for admitted work. The transfer manager needs to admit planned demand against a normal memory ceiling, let an otherwise idle oversized request make progress, support acquisitions that are not backed by a reservation, and preserve accounting while reservations close independently of retained byte owners.

A reservation is granted once for an admitted work item. That work may acquire and return many carriers over its lifetime. Acquisitions covered by open reservation envelopes and returns that restore their coverage use one packed atomic state. Operations that can change reservation eligibility or prepared capacity use the admission mutex: grants, close, FIFO mutation, uncovered shortfalls and repayments, physical preparation, and serialized arena fallback.

The composed path also introduces failure boundaries after accounting has changed or bitmap bits have been claimed. Those paths must return physical ownership before releasing its charge, prevent a close race from reopening acquisition authority, and avoid waking reservation futures while pool locks are held.

### Pool state

`BufferPool` is a cloneable handle to one `Arc<PoolInner>`. Reservations and carrier owners retain the same shared state, so dropping one manager-local handle does not invalidate ownership that escaped it.

```text
BufferPool
`-- Arc<PoolInner>
    |-- admission: Mutex<AdmissionState>
    |   |-- configured capacity
    |   |-- prepared capacity
    |   |-- active planned demand
    |   `-- FIFO waiters
    |-- coverage: AtomicU64
    |   |-- available coverage
    |   `-- uncovered charges
    `-- arena: physical carrier storage
```

The accounting terms describe different parts of one memory state:

| Term                    | Meaning                                                                 |
| ----------------------- | ----------------------------------------------------------------------- |
| Active planned demand   | Sum of open reservation envelopes                                       |
| Available coverage      | Unused portion of active planned demand                                 |
| Covered charges         | Portion of active planned demand currently held by acquisitions         |
| Uncovered charges       | Carrier charges not represented by active planned demand                |
| Outstanding charges     | All covered and uncovered carrier ownership, including in-flight claims |
| Admission used          | Active planned demand plus uncovered charges                            |
| Prepared capacity       | Carrier capacity whose physical preparation completed                   |

```text
covered charges     = active planned demand - available coverage
outstanding charges = covered charges + uncovered charges
```

The normal grant rule keeps `admission_used + new_envelope` within configured capacity. When no reservation envelope remains active, one fresh request or the FIFO head may exceed that ceiling so work can make progress. Every grant prepares the arena through its complete post-grant admission floor before publishing the reservation. Block-sized preparation may leave prepared capacity above that floor.

The corresponding design sections are **Admission and accounting**, **Reservation admission**, and **Wait queue**.

### Reservation admission

`BufferPool::try_reserve` attempts an immediate grant without bypassing older waiters. `BufferPool::reserve` creates a lazy `ReserveFuture`; its first poll either receives an immediate prepared reservation or enters the pool-wide FIFO.

```text
try_reserve(bytes)                  reserve(bytes).poll()
        |                                  |
        `---------- carrier envelope ------'
                           |
                    lock admission
                           |
             queue empty and eligible?
                    /              \
                  yes               no
                   |                 |
       prepare post-grant floor      `--> FIFO WaitSlot --> Pending
                   |
       publish planned demand
       and available coverage
                   |
             Reservation
```

The FIFO is strict: an ineligible head blocks later requests that would fit. `WaitSlot` is the linearization point between grant, cancellation, and result consumption. Admission stores a terminal grant or failure before unlinking the head, then invokes its waker after releasing pool locks. Dropping a queued future removes its link and reconsiders the exposed head. Dropping a granted but unobserved future closes the transferred reservation.

`Reservation` is non-cloneable. Its private `ReservationState` carries the complete envelope and one packed atomic containing the closed bit plus direct owners and in-flight direct debits. `Reservation::close_acquisition` and `Drop` perform the same idempotent close: new direct debits stop, the envelope leaves active planned demand, unused coverage is removed, and surviving charges become uncovered. Close uses the reservation's direct count and remaining active demand so it cannot consume coverage needed by another open reservation. Existing carrier owners remain valid until their final drops.

### Acquisition

`BufferPool::acquire` uses a reservation and enforces its direct carrier envelope. `BufferPool::acquire_unreserved` skips reservation-local authority but uses the same aggregate accounting and physical transaction. Both round the byte request to complete carriers and return `AcquiredRuns`, which groups adjacent carrier owners for the mutable-buffer layer.

```text
acquire(reservation, bytes)          acquire_unreserved(bytes)
              |                                |
              | advisory direct precheck       |
              `------------- aggregate debit --'
                                  |
                   available coverage sufficient?
                         /                  \
                       yes                   no
                        |                     |
                 packed CAS fast path    lock admission
                                           publish shortfall charge
                                           prepare admission floor
                                           unlock
                        \                     /
                         `-- aggregate charge published
                                      |
                         reserved direct debit?
                           |             |
                         reject        authorized or unreserved
                           |             |
                    aggregate rollback  optimistic arena claim
                                      |
                              complete batch?
                              /             \
                            yes              no
                             |                |
                             |          lock admission
                             |          serialized fallback
                             |          unlock
                             `------- ClaimBatch::finish
                                      |
                                  CarrierGuard[]
                                      |
                                  AcquiredRuns
```

`acquire` can cheaply reject a reservation already known to be closed or out of capacity. Close may race that check, so the later atomic direct debit decides whether the acquisition is authorized. If it loses, the pool rolls back the aggregate charge already published for the attempt.

The shortfall path publishes its charge before releasing admission. That charge raises the prepared-capacity floor while the optimistic bitmap claim is in flight, so trim cannot reclaim the capacity being acquired. A partial optimistic claim later reacquires admission and passes the held `AdmissionGuard` into arena fallback; the arena cannot acquire admission recursively.

`PendingAcquisition` owns the accounting debit, provisional bitmap claim, and any converted guards until the complete transaction commits. Every failure path drops guards and provisional physical ownership before releasing aggregate and reservation-local charges. Collection allocation for guard and run metadata is fallible, and failure injection exercises rollback after each commit boundary.

The corresponding design sections are **Mutable buffer acquisition and growth**, **Carrier claiming and fallback**, and Appendix A's **Acquisition transitions**.

### Return and progress

Each `CarrierGuard` owns one physical `CarrierAllocation`, one aggregate charge, and optional reservation-local provenance. Its final drop clears the physical bitmap bit first, retires direct authority second, and releases the aggregate charge last.

```text
CarrierGuard::drop
        |
        +-- return CarrierAllocation to the bitmap
        +-- retire reservation-local ownership
        `-- release aggregate charge
                    |
           removed uncovered charge?
              /                \
            no                  yes
             |                   |
      lock-free return      lock admission
                            drain eligible FIFO heads
                            unlock
                            wake terminal futures
```

A carrier acquired within an open envelope is covered, not uncovered. Its return restores available coverage through the packed atomic without taking admission. Ownership becomes uncovered when it exceeds available coverage or remains live after its reservation closes. A return that removes such a charge takes admission because the reduced pressure may make the FIFO head eligible. Closing a reservation while many carriers remain live therefore makes their later returns use the serialized path. Close and return may occur in either order and converge on the same final accounting state.

### Metrics

`BufferPool::metrics` returns a private, getter-only `MemoryMetrics` value. It holds admission serialization and loads the packed coverage state once, producing one sample of configured capacity, admission use, active planned demand, charged capacity, overage, prepared capacity, queue depth, and cumulative parked reservations. The representation does not expose ledger fields, pool authority, bitmap population, or arena policy.

The corresponding design section is **Observability**.

## How to review

Start with one reservation and the carriers acquired through it. Let `C` be one carrier's byte capacity. This sequential lifecycle establishes the accounting model before concurrency and rollback add more paths:

| Operation on the same reservation | Active planned demand | Available coverage | Covered live carriers | Uncovered live carriers |
| --- | ---: | ---: | ---: | ---: |
| `reservation = pool.try_reserve(4 * C)` | 4 | 4 | 0 | 0 |
| `pool.acquire(&reservation, 3 * C)` | 4 | 1 | 3 | 0 |
| `drop(one CarrierGuard)` while the reservation is open | 4 | 2 | 2 | 0 |
| `reservation.close_acquisition()` with two guards still live | 0 | 0 | 0 | 2 |
| `drop(the remaining two CarrierGuards)` | 0 | 0 | 0 | 0 |

`Available coverage` is unused capacity inside open reservation envelopes, not free physical capacity. Close withdraws the complete envelope: its two unused units disappear, while its two live covered carriers become uncovered. Those later returns reduce uncovered ownership because no open envelope remains whose available coverage could be restored.

Trace that lifecycle through the implementation in the same order:

1. Begin with the grant at `BufferPool::try_reserve`. Follow the request through eligibility, arena preparation, and grant publication. The reservation must not become visible until physical capacity covers the complete post-grant admission floor. Then inspect `ReserveFuture` as the queued form of that same transition: a terminal grant or failure is stored in its `WaitSlot` before the FIFO link is removed, and the waker runs only after pool locks are released.
2. Follow `BufferPool::acquire` for the three-carrier request. The aggregate debit first converts three units of available coverage into covered ownership. The reservation-local debit independently proves that this reservation still authorizes those carriers. Only then does the transaction claim bitmap ownership and transfer the accounting and physical capabilities into `CarrierGuard`s. Compare the covered fast path with a shortfall path: shortfall publishes its charge and prepares its floor under admission, releases admission for the optimistic claim, and reacquires one guard only if serialized arena fallback is required.
3. Follow the first `CarrierGuard` return while the reservation remains open. Drop returns the physical bitmap bit, retires reservation-local ownership, and finally releases the aggregate charge. That packed return restores one unit of available coverage without taking admission because active planned demand has not changed.
4. Close the reservation while two guards remain live. `ReservationState::close_acquisition` first prevents new direct debits, then removes the four-unit envelope under admission. The two unused units leave available coverage and the two live covered charges become uncovered. Trace the multi-reservation regression alongside this step: closing one reservation must not consume unused coverage that belongs to another open reservation.
5. Return the final two guards after close. The physical-first drop order is unchanged, but each aggregate return now removes uncovered ownership. That reduction can make the FIFO head eligible, so the return enters admission, drains eligible heads, unlocks, and only then invokes their wakers.

After the sequential trace is clear, inspect the seams where another operation may observe an intermediate state: direct debit versus close, close versus return, repayment versus waiter enqueue, and unlocked shortfall claim versus trim. The Loom models cover those linearization points. Allocation-failure tests exercise rollback after accounting, provisional bits, guards, or run metadata have been created and verify that physical ownership returns before its charges.

The small-state tests in `coverage.rs` validate each transition equation. The multi-reservation composition test validates the outcome those equations must collectively preserve: live ownership remains within the normal configured ceiling, charged capacity equals live ownership, and active planned demand equals the sum of open envelopes after every operation.

## Future

The next PR consumes `AcquiredRuns` into `PooledBufMut`, adds retained reservation-backed growth, publishes owner-backed immutable byte views, and builds segmented delivery. Public configuration, maintenance, transfer-manager wiring, and transport integration remain outside this PR.

### DIFF
--- a/aws-sdk-s3-transfer-manager/src/runtime/buffer_pool/admission/coverage.rs
+++ b/aws-sdk-s3-transfer-manager/src/runtime/buffer_pool/admission/coverage.rs
@@ -1,0 +1,562 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Lock-free aggregate acquisition accounting.
+//!
+//! Active planned demand remains under admission serialization. Available
+//! coverage and uncovered charges share one atomic word so debit, return,
+//! grant, and close cannot observe or publish a torn pair.
+
+use crate::runtime::sync::sync::atomic::{AtomicU64, Ordering};
+
+use super::super::CarrierCount;
+
+/// Number of bits in one packed accounting lane.
+const LANE_BITS: u32 = u32::BITS;
+
+/// Largest count represented by one packed accounting lane.
+pub(in crate::runtime::buffer_pool) const MAX_PACKED_CARRIERS: CarrierCount =
+    CarrierCount::new(u32::MAX as usize);
+
+/// A packed transition cannot represent its result.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(super) struct CoverageOverflow;
+
+/// Aggregate state changed at carrier-acquisition frequency.
+pub(in crate::runtime::buffer_pool) struct CoverageState {
+    /// Low lane: available coverage. High lane: uncovered charges.
+    packed: AtomicU64,
+}
+
+/// One coherent sample of [`CoverageState`].
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(super) struct CoverageSnapshot {
+    /// Open-envelope capacity not occupied by an acquisition.
+    pub(super) available: CarrierCount,
+    /// Ownership pressure outside open-envelope coverage.
+    pub(super) uncovered: CarrierCount,
+}
+
+/// Result of installing aggregate acquisition charges.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(super) struct CoverageDebit {
+    /// Uncovered charges introduced by this transition.
+    pub(super) uncovered_added: CarrierCount,
+}
+
+/// Result of retiring aggregate acquisition charges.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(super) struct CoverageReturn {
+    /// Uncovered charges removed by this transition.
+    pub(super) uncovered_removed: CarrierCount,
+}
+
+impl CoverageState {
+    /// Creates empty aggregate accounting.
+    pub(in crate::runtime::buffer_pool) fn new() -> Self {
+        Self {
+            packed: AtomicU64::new(0),
+        }
+    }
+
+    /// Debits only when available coverage satisfies the complete request.
+    ///
+    /// `Ok(false)` leaves state unchanged. The caller must enter admission
+    /// serialization before installing an uncovered charge.
+    pub(super) fn try_debit_covered(&self, count: CarrierCount) -> Result<bool, CoverageOverflow> {
+        checked_lane(count)?;
+        let mut current = self.packed.load(Ordering::Acquire);
+
+        loop {
+            let snapshot = unpack(current);
+            let Some(available) = snapshot.available.checked_sub(count) else {
+                return Ok(false);
+            };
+            let next = pack(available, snapshot.uncovered)?;
+            match self.packed.compare_exchange_weak(
+                current,
+                next,
+                Ordering::AcqRel,
+                Ordering::Acquire,
+            ) {
+                Ok(_) => return Ok(true),
+                Err(observed) => current = observed,
+            }
+        }
+    }
+
+    /// Installs acquisition charges, consuming coverage before adding overage.
+    ///
+    /// Callers that may add an uncovered charge hold admission serialization.
+    /// `maximum_uncovered` enforces the packed global admission bound.
+    pub(super) fn debit(
+        &self,
+        count: CarrierCount,
+        maximum_uncovered: CarrierCount,
+    ) -> Result<CoverageDebit, CoverageOverflow> {
+        checked_lane(count)?;
+        checked_lane(maximum_uncovered)?;
+        self.update(|snapshot| {
+            let covered = CarrierCount::new(count.get().min(snapshot.available.get()));
+            let uncovered_added = count.checked_sub(covered).ok_or(CoverageOverflow)?;
+            let available = snapshot
+                .available
+                .checked_sub(covered)
+                .ok_or(CoverageOverflow)?;
+            let uncovered = snapshot
+                .uncovered
+                .checked_add(uncovered_added)
+                .filter(|next| *next <= maximum_uncovered)
+                .ok_or(CoverageOverflow)?;
+            Ok((
+                CoverageSnapshot {
+                    available,
+                    uncovered,
+                },
+                CoverageDebit { uncovered_added },
+            ))
+        })
+    }
+
+    /// Retires charges, removing uncovered charges before restoring coverage.
+    pub(super) fn release(&self, count: CarrierCount) -> CoverageReturn {
+        checked_lane(count).expect("released carrier count must fit packed accounting");
+        self.update(|snapshot| {
+            let uncovered_removed = CarrierCount::new(count.get().min(snapshot.uncovered.get()));
+            let restored = count
+                .checked_sub(uncovered_removed)
+                .ok_or(CoverageOverflow)?;
+            let available = snapshot
+                .available
+                .checked_add(restored)
+                .ok_or(CoverageOverflow)?;
+            let uncovered = snapshot
+                .uncovered
+                .checked_sub(uncovered_removed)
+                .ok_or(CoverageOverflow)?;
+            Ok((
+                CoverageSnapshot {
+                    available,
+                    uncovered,
+                },
+                CoverageReturn { uncovered_removed },
+            ))
+        })
+        .expect("aggregate return must fit packed accounting")
+    }
+
+    /// Publishes unused coverage from one newly active envelope.
+    ///
+    /// Existing uncovered charges remain unchanged. A later grant cannot
+    /// absorb ownership that predates it.
+    pub(super) fn add_coverage(&self, count: CarrierCount) -> Result<(), CoverageOverflow> {
+        checked_lane(count)?;
+        self.update(|snapshot| {
+            let available = snapshot
+                .available
+                .checked_add(count)
+                .ok_or(CoverageOverflow)?;
+            Ok((
+                CoverageSnapshot {
+                    available,
+                    uncovered: snapshot.uncovered,
+                },
+                (),
+            ))
+        })
+    }
+
+    /// Withdraws one closing envelope without consuming another envelope's coverage.
+    ///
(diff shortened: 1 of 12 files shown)


aws-sdk-s3-transfer-manager/src/runtime/buffer_pool/admission/coverage.rs
```diff
@@ -0,0 +1,575 @@
...
+                Err(observed) => current = observed,
+            }
+        }
+    }
+
+    /// Installs acquisition charges, consuming coverage before adding overage.
+    ///
+    /// Callers that may add an uncovered charge hold admission serialization.
+    /// `maximum_uncovered` enforces the packed global admission bound.
+    pub(in crate::runtime::buffer_pool) fn debit(
```
ysaito1001, line 97: There might be a missing operation.

The pool tracks memory in two buckets: reserved-but-unused (`available`) and used-without-a-reservation (`uncovered`, the overage). Spending prefers the polite bucket; ask for memory and it takes from available first, only creating overage if that runs dry. Refunding prefers the other bucket. Give memory back and it pays down overage first, only returning capacity to available once the overage is clear.
  
Individually both are right. Spending should avoid creating overage; returning should prioritize clearing it. The problem is that the code uses the refund path to cancel a spend that never happened.
  
Ex. We ask for 1 unit, it comes out of available. The request then fails. The cancel runs the refund logic, which sees outstanding overage and puts our unit there instead of back where it came from.

<details>
<summary>Failing test — paste into <code>mod tests</code> in this file</summary>

    #[test]
    fn debit_then_release_restores_the_prior_classification() { 
        let coverage = CoverageState::new();
        // An unreserved acquisition leaves overage behind.
        coverage
            .debit(CarrierCount::new(2), CarrierCount::new(64))
            .unwrap();
        // A later grant publishes its envelope; the overage persists (A5).
        coverage.add_coverage(CarrierCount::new(4)).unwrap();
        let before = coverage.snapshot();
        assert_eq!(
            (before.available.get(), before.uncovered.get()),
            (4, 2),
            "setup"
        );

        // A fully covered acquisition, then the rollback path used on failure.
        assert!(coverage.try_debit_covered(CarrierCount::new(1)).unwrap());
        coverage.release(CarrierCount::new(1));
        let after = coverage.snapshot();

        assert_eq!(
            (before.available.get(), before.uncovered.get()),
            (after.available.get(), after.uncovered.get()),
            "rollback of an unexposed debit must not reclassify pre-existing overage"
        );
    } 

```
assertion `left == right` failed: rollback of an unexposed debit must not reclassify pre-existing overage

 left: (4, 2)
right: (3, 1)
```
</details>

ysaito1001 (approved):
LGTM, with one potential bug in the accounting core.

aws-sdk-s3-transfer-manager/src/runtime/buffer_pool/admission/coverage.rs
aajtodd: 73a87af

PR #173, opened 2026-09-01
buffer pool: maintenance

aajtodd:
## Summary

This PR gives the buffer pool an owner for work that must happen after allocation: checked construction, idle block reclamation, recovery from failed mapping cleanup, and memory diagnostics. Pool construction remains lazy. It selects capacity and geometry but reserves no virtual ranges, prepares no backing, and starts no thread.

This is the maintenance layer in the buffer-pool series:

- **This PR:** configuration, reclamation, cleanup recovery, and diagnostics

The design authority is `docs/design/memory.md` as inherited by this branch, especially **Capacity configuration**, **Idle epochs and reclamation**, **Shutdown and destruction**, and **Metrics and observability**. The design PR is linked above because those sections are not yet present on `main`.

## Why

Returning the last owner clears a carrier's bitmap bit, but it does not make the containing block inaccessible or release its backing. Without maintenance, a pool that reached a high prepared capacity would retain that capacity for its complete lifetime. A failed protection or discard operation can also leave a block deliberately nonclaimable; ordinary allocation must not guess whether that mapping is safe to reuse.

The scheduler knows when the transfer manager becomes globally idle. This PR uses that low-frequency event to retain warm capacity across short gaps, then reclaim complete free blocks without adding work to every carrier return. Mapping failures use the same worker for bounded recovery.

### Capacity and geometry

`MemoryBudgetConfig` is the single `Auto`, `Fraction`, or `Limit` policy used to construct a pool. `PoolConfig` resolves that policy against detected effective memory and runtime page size, then produces checked accounting and storage geometry.

The provisional defaults are a 16 KiB carrier and a 128 MiB target block. Runtime page size may increase the carrier size, and pools configured below 128 MiB use their complete carrier-rounded capacity as one block. Block bytes are policy; carriers per block and bitmap words are derived. This keeps the reclamation extent independent of bitmap representation and allows common multipart payloads to fit within one stable provenance range.

`Auto` selects one quarter of detected effective memory with a 32 GiB ceiling and a 2 GiB fallback when detection is unavailable. `Fraction` requires detected memory and uses an exact binary64 floor so capacity never rounds above the requested share. `Limit` bypasses detection.

The optimistic scan target is also byte-denominated. It covers at least 32 MiB after rounding to complete bitmap words, so an 8 MiB acquisition uses one quarter of the default scan window instead of requiring every observed carrier to be free.

The pool and the existing transfer-manager path now share the `MemoryBudgetConfig` type but do not yet share resolution behavior. The existing public path retains its documented clamping rules; this staged pool uses the checked rules above. Production integration must converge them on one public contract.

### Idle epochs and reclamation

Global idle starts a reclaim epoch. New managed work invalidates that epoch before it can reclaim against stale demand.

```text
BufferPool::record_global_idle()
    |
    +-- arm one 30-second deadline for the current activity epoch
    |
    +-- BufferPool::record_managed_activity() before expiry
    |       `-- invalidate the deadline and any reclaim retry
    |
    `-- deadline expires
            `-- capture one whole-block retention target for this epoch
                    `-- execute bounded reclaim passes until complete or blocked
```

The target takes one quarter of the configured whole-block ceiling, rounds down to complete blocks, and remains fixed across retries. A pool spanning fewer than four blocks therefore retains no warm block after the deadline instead of making reclamation a no-op. Each pass recomputes the current admission floor, so a stale target cannot reclaim capacity promised to reservations or held by owners.

Reclaim separates authorization from platform work:

```text
under admission serialization:
    recompute current floor
    select and recheck one all-free block
    remove that block from prepared capacity

without admission or maintenance-control locks:
    make the range inaccessible
    discard backing
```

Selection is only a hint. `BlockSlot::start_trim` remains the authority that closes the claim-trim gate and verifies the floor. A concurrent claim therefore either keeps the block active or loses its provisional bits before trim proceeds.

### Cleanup recovery and worker lifetime

A failed mapping transition leaves the affected slot nonclaimable and publishes a cleanup request. Cleanup generations prevent completion of an older retry from erasing newer work, and cleanup runs before idle reclaim because a failed protection transition can strand otherwise reusable capacity.

The first idle deadline or cleanup request starts one named maintenance thread. The worker waits on serialized state plus a condition variable, which coalesces repeated activity, deadlines, cleanup generations, and shutdown into current state instead of queueing stale events. It retains a `Weak<PoolInner>` while waiting, so maintenance alone cannot keep the pool alive. Final destruction requests a stop, wakes the worker, and joins it unless the worker temporarily owns the final pool reference and is performing its own teardown.

Thread creation failure disables maintenance without disabling allocation. The failure is logged under the memory telemetry target and reported as `maintenance_disabled` in diagnostics.

### Metrics and diagnostics

`BufferPool::metrics()` reads configured, promised, owned, prepared, and queued capacity under admission serialization. These values form one accounting sample. Preparation rounds its current floor up to complete blocks, adding less than one block beyond that floor; idle-only admission overage can also place prepared capacity above the normal configured ceiling.

`BufferPool::diagnostics()` answers different operational questions: whether claims are missing optimistic scans, whether fallback or rollback is frequent, how often blocks are prepared or reclaimed, how expensive trim selection is, and whether mapping cleanup or the worker is degraded. It briefly locks maintenance control to read permanent-disable state; its lifecycle fields come from a best-effort arena scan while the pool is active. Test audits make the same reconstruction authoritative only after the caller establishes quiescence.

## How to review

1. Follow `BufferPool::from_capacity` into `PoolConfig::resolve` and `PoolInner::new`. This establishes the memory policy, independent carrier and block geometry, and the guarantee that construction allocates no ranges or worker.
2. Follow `BufferPool::record_global_idle` and `BufferPool::record_managed_activity` into `MaintenanceState`. The state machine turns explicit scheduler events and time into reclaim or cleanup actions while rejecting stale epochs and cleanup completions.
3. Continue through `MaintenanceCoordinator::ensure_worker`, `wait_for_action`, and `worker_loop`. This shows lazy creation, condition-variable wakeups, weak pool ownership, spawn failure, and final shutdown before pool operations enter the picture.
4. Read `execute_maintenance`, `reclaim_to`, `Arena::start_trim`, and `TrimCleanup::finish` as one path. The important boundary is where admission authorizes removal and releases its lock before protection and discard.
5. Finish with `BlockSlot::sample_lifecycle`, `MemoryMetrics`, `MemoryDiagnostics`, and `PoolAuditReport`. These distinguish coherent accounting from a concurrent operational sample and from an exact externally quiescent audit.

The colocated state-machine tests exercise stale epochs, fixed retry targets, cleanup-generation races, disable, and stop. The native worker tests exercise timed wakeup, concurrent start, spawn failure, weak retention, and final-owner teardown. Loom composes claim, return, trim, cleanup, and shutdown transitions. Failure injection demonstrates that failed mapping transitions remain unavailable until recovery, while deliberate audit mismatches prove that the global reconciliation rejects accounting drift.

## Future

Global idle is the first reclamation trigger. Sustained demand that falls from a peak without reaching zero may retain peak prepared capacity. A later utilization-decay policy can reuse the same worker, admission floor, trim gate, and retry machinery if workload measurements show that retention matters in practice.

Carrier, block, and scan defaults remain measurement choices. The hardening pass will compare geometry using optimistic-hit and serialized-fallback rates, segment count, opaque-`Bytes` copies, mapping cost, reclaim behavior, and retained RSS, and will expand native platform qualification beyond the current Linux, macOS, Windows, and FreeBSD coverage.

The next integration PR will replace the legacy `MemoryBudget` path with this pool, settle one resolution contract for `MemoryBudgetConfig`, carry reservations through scheduler dispatch, copy ordinary Hyper frames into pooled staging, and expose the resulting memory state through the production transfer-manager surface.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,7 @@ jobs:
       - dial9-build
       - miri
       - loom
+      - freebsd-test
     steps:
       - run: exit 0
 
@@ -173,6 +174,30 @@ jobs:
         run: cargo check --workspace
         env:
           RUSTFLAGS: "" # remove -Dwarnings
+
+  freebsd-test:
+    name: Test FreeBSD 14.4
+    runs-on: ubuntu-24.04
+    timeout-minutes: 30
+    needs: basics
+    steps:
+      - uses: actions/checkout@v4
+      - name: test transfer-manager on FreeBSD
+        uses: vmactions/freebsd-vm@83b151f58c6047089f4c80eb5ba2039d158ce093 # v1.5.3
+        with:
+          release: '14.4'
+          envs: RUST_BACKTRACE RUSTFLAGS
+          usesh: true
+          copyback: false
+          prepare: |
+            pkg install -y cmake curl ninja perl5 pkgconf
+            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs -o rustup.sh
+            sh rustup.sh -y --profile minimal --default-toolchain ${{ env.rust_stable }}
+          run: |
+            set -eux
+            . "$HOME/.cargo/env"
+            export CMAKE_GENERATOR=Ninja
+            cargo test -p aws-sdk-s3-transfer-manager --lib
 
   check-external-types:
     name: check-external-types (${{ matrix.os }})
--- a/aws-sdk-s3-transfer-manager/src/runtime/buffer_pool.rs
+++ b/aws-sdk-s3-transfer-manager/src/runtime/buffer_pool.rs
@@ -35,6 +35,7 @@ use block::BlockError;
 use config::ResolvedPoolConfig;
 pub(crate) use config::{BufferPoolBuildError, MemoryCapacity};
 use geometry::{GeometryError, PoolGeometry};
+use maintenance::MaintenanceCoordinator;
 use metrics::{MemoryMetricState, MemoryMetrics};
 use pooled_buf::GrowthAuthority;
 pub(crate) use pooled_buf::PooledBufMut;
@@ -195,6 +196,8 @@ struct PoolInner {
     coverage: CoverageState,
     /// Stable virtual ranges and physical carrier ownership.
     arena: Arena,
+    /// Lazy idle reclamation and mapping-cleanup worker.
+    maintenance: MaintenanceCoordinator,
     /// Per-pool failure injection and test-only observations.
     #[cfg(test)]
     test_hooks: TestHooks,
@@ -212,6 +215,7 @@ impl PoolInner {
             admission: Mutex::new(AdmissionState::new(configured_capacity)),
             coverage: CoverageState::new(),
             arena: Arena::new(geometry, optimistic_scan_words)?,
+            maintenance: MaintenanceCoordinator::new(),
             #[cfg(test)]
             test_hooks: TestHooks::new(),
         })
@@ -443,6 +447,12 @@ impl PoolInner {
             wakers
         };
         wake_all(wakers);
+    }
+}
+
+impl Drop for PoolInner {
+    fn drop(&mut self) {
+        self.maintenance.shutdown();
     }
 }
 
--- a/aws-sdk-s3-transfer-manager/src/runtime/buffer_pool.rs
+++ b/aws-sdk-s3-transfer-manager/src/runtime/buffer_pool.rs
@@ -36,7 +36,7 @@ use config::ResolvedPoolConfig;
 pub(crate) use config::{BufferPoolBuildError, MemoryCapacity};
 use geometry::{GeometryError, PoolGeometry};
 use maintenance::MaintenanceCoordinator;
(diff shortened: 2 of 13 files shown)


ysaito1001 (approved):
Code LGTM.
  
To clarify, if I'm reading this right, memory goes back to the OS only when the transfer manager is globally idle. So a single multi-hour transfer would prevent a pool that has already grown from ever shrinking (assuming that's the case the utilization-decay policy under `Future` section is meant to cover)?

aajtodd:
> To clarify, if I'm reading this right, memory goes back to the OS only when the transfer manager is globally idle. So a single multi-hour transfer would prevent a pool that has already grown from ever shrinking (assuming that's the case the utilization-decay policy under `Future` section is meant to cover)?

Yes. For now it was simpler to try reclaim when totally idle, nothing prevents us from something more sophisticated potentially but wanted to see if it was even an issue before committing to a more complicated reclaim protocol.

PR #175, opened 2026-09-08
Add the functional specification for `s3 sync`

ysaito1001:
## Summary
This PR adds a functional specification for `aws s3 sync`-equivalent functionality in the Rust transfer manager.
  
Subsequent PRs will build on it, each pairing a targeted design document covering a subset of these requirements with the code that implements them.

[Rendered](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/d777636ee85e6ded24856ec1e86271ba9fa25855/docs/design/sync/functional-spec.md)


docs/design/sync/functional-spec.md
```diff
@@ -0,0 +1,675 @@
...
+| **relative key** | An entry's name relative to its own root, as a **sequence of bytes** with `/` as the separator. The join key between the two sides. Bytes rather than text because a local filename need not be valid UTF-8, and deriving a text form from one that is not loses information. |
+| **pair state** | For a relative key: `SrcOnly`, `DestOnly`, or `Both`. |
+| **action** | The decision for a relative key: `Transfer`, `Delete`, or `Skip`. |
+| **direction** | `Upload` (local → S3), `Download` (S3 → local), `Copy` (S3 → S3). |
+| **plan** | The complete set of (relative key, action) decisions for a run. |
+| **run** | One invocation: enumerate both roots, decide an action per relative key, execute. |
+
+---
+
+## 2. Directions and roots (`FR-Root-*`)
```
aajtodd, line 78: (minor/nit): Might be good for each FR category (root, enum, etc) to add a small description of the category before leading into the individual requirements. Took me a second to realize `Enum` was "enumeration" for instance.

docs/design/sync/functional-spec.md
```diff
@@ -0,0 +1,675 @@
...
+## 1. Definitions
+
+| Term | Meaning |
+|---|---|
+| **root** | A source or destination location: a local directory, or a bucket plus key prefix. |
+| **entry** | One comparable unit under a root: a local file, or an S3 object. |
+| **relative key** | An entry's name relative to its own root, as a **sequence of bytes** with `/` as the separator. The join key between the two sides. Bytes rather than text because a local filename need not be valid UTF-8, and deriving a text form from one that is not loses information. |
+| **pair state** | For a relative key: `SrcOnly`, `DestOnly`, or `Both`. |
+| **action** | The decision for a relative key: `Transfer`, `Delete`, or `Skip`. |
+| **direction** | `Upload` (local → S3), `Download` (S3 → local), `Copy` (S3 → S3). |
```
aajtodd, line 72: Question on `Copy`. We have it planned to support copy but it doesn't exist yet, do you plan on building sync without copy to start and then backfilling it? Or implementing copy as a pre-requisite? (I don't think it matters that much so no preference, just curious what your plan is)

docs/design/sync/functional-spec.md
```diff
@@ -0,0 +1,675 @@
...
+One policy, so a caller reasons about failure once. Two things sit outside it:
+
+- An entry that could never be transferred whatever the settings — device, FIFO, socket, a symlink sync was
+  told not to follow — is a warning under either policy (FR-Enum-3).
+- A failure that leaves nothing to carry on with — a source root that is unreadable or is not a directory
+  (FR-Fail-6) — ends the run under either policy. Not because the policy is overridden, but because there
+  is nothing left to continue doing.
+*`[DERIVED]` — from FR-Fail-2 and FR-Fail-3: a policy only some failures consult is not a policy, and a per-site decision is how FR-Enum-12 and FR-Fail-7 come to disagree. `[DOC]` rsync draws the same line — a "non-regular file" warning changes neither exit status nor deletion, while an I/O error does both. The exclusion preserves FR-Enum-3, whose provenance is two reports of `aws s3 sync` failing an entire run over one such entry.*
+
+## 8. Dry run (`FR-Dry-*`)
```
aajtodd, line 662: I think there's two different ways we can view dry run, (1) build it in to TM, (2) expose enough of the underlying walk/enumeration primitives that dry run can be implemented outside TM. I don't have a strong preference but the current design of the walk APIs was sort of leaning towards the latter. What are your thoughts on this?

docs/design/sync/functional-spec.md
```diff
@@ -0,0 +1,675 @@
...
+source, but they are issued **against the destination bucket** and name the source in a `CopySource`
+parameter. So the destination credentials must be able to read the source object. Supplying separate source
+credentials does not lift that requirement: those credentials are used for listing and for reading source
+properties, never for the copy.
+*`[DOC]` `--source-region`, [Example 7](https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html#examples). `[CLI]` `subcommands.py` → `S3TransferCommand._get_source_and_transfer_clients` builds the pair; `CommandArchitecture.run` gives the forward (source) generator `self._source_client` and the reverse (destination) generator `self._client`. The copy going to the destination follows from `_get_transfer_manager(botocore_transfer_client=transfer_client)` — the transfer manager that issues the copy holds only the destination client. Source property reads follow from `subscribers.py`, where the tag and metadata subscribers read through `fileinfo.source_client` (`self._source_client.get_object_tagging`).*
+
+**FR-Root-7** Sync MUST reject a configuration it cannot work correctly with, before doing any work, and
```
aajtodd, line 124: This is what CLI does today on a directory bucket. I think it's fine as a baseline. 

Can we do better though? What's it take to support sync on directory buckets is a question I'd be asking and hoping we can resolve to unblock that (probably want to talk with CRT or CLI folks) at some point.

docs/design/sync/functional-spec.md
```diff
@@ -0,0 +1,675 @@
...
+opt-in, and a retained multipart upload MUST be reported, because no object listing reveals one.
+*`[DOC]` S3 user guide: "you must either complete or stop the multipart upload to stop incurring charges for storage of the uploaded parts", and "After you initiate a multipart upload, there is no expiry". `[DERIVED]` — reporting is required because the leak is invisible to every listing a caller would think to check.*
+
+**FR-Exec-23** A run MUST be cancellable. On cancellation sync MUST stop starting new work, MUST NOT leave a
+half-written destination entry looking like a finished one, and MUST report what did finish.
+*`[NEW]` — the CLI's sync path has no cancellation surface: `s3handler.py` → `S3TransferHandler.call` submits inside `with self._transfer_manager:` and blocks, and no sync module contains cancel or shutdown handling. Process-level interrupt behavior sits outside the sync code and is not asserted here. FR-Exec-13 supports the "no partial entry" half.*
+
+## 7. Skips, warnings, failures (`FR-Fail-*`)
+
+**FR-Fail-1** An object sitting in an archival storage class, not currently restored, MUST be skipped with
```
aajtodd, line 568: The requirement says storage class and restore state both come back in the listing. Storage class does but I don't think restore state does. `Restore` is a HeadObject header. 

ListObjectsV2 can give us this via `x-amz-optional-object-attributes: RestoreStatus`. It's a tri-state: absent means never restored, IsRestoreInProgress=true means underway, false plus RestoreExpiryDate means ready.

docs/design/sync/functional-spec.md
```diff
@@ -0,0 +1,675 @@
...
+**FR-Fail-5** The result MUST distinguish three overall outcomes: at least one failure; no failures but
+at least one warning; clean.
+*`[CLI]` `subcommands.py` → `CommandArchitecture.run`: `rc = 1` if `num_tasks_failed > 0`, `rc = 2` if `num_tasks_warned > 0`, else `0`.*
+
+**FR-Fail-6** Failing to list a root at all MUST fail the run, and MUST be reported differently from an
+individual entry failing. One entry failing leaves a usable run; a root that cannot be listed means sync
+never knew what was there.
+*`[CLI]` `filegenerator.py` → `list_objects` / `_list_single_object` let `ClientError` propagate out of the generator, failing the command rather than producing an empty side.*
+
+**FR-Fail-7** A key MUST NOT be deleted unless sync actually saw the source side and saw that the
```
aajtodd: If a source ListObjectsV2 page fails and we're in continue mode, is every key in that page's range unknown?

docs/design/sync/functional-spec.md
aajtodd, line 78: Do we need a bullet or FR on preventing download writing outside the destination root?

docs/design/sync/functional-spec.md
```diff
@@ -0,0 +1,675 @@
...
+
+**NFR-Cost-1** A no-op sync of N entries per side MUST cost O(N / page size) list requests per side and
+zero per-entry requests, unless the caller opted into one of the modes FR-Cmp-5 excepts.
+*`[DERIVED]` from FR-Cmp-5 plus `[DOC]` `--page-size` (default and maximum 1000) and `utils.py` → `BucketLister.list_objects` pagination.*
+
+**NFR-Lat-1** The time until the first transfer starts MUST NOT depend on how many entries there are in
+total. A run over a million keys MUST begin transferring as promptly as a run over ten.
+*`[CLI]` satisfied by the generator chain (FR-Enum-7).*
+
+**NFR-Mem-1** Peak memory MUST NOT grow with the number of entries a run covers. Measured over the same
```
aajtodd, line 722: Do we need to consider really large number of object (million+) run that has a lot of errors or telemetry? i.e. what if we had to retain a million error records to surface to a user?

docs/design/sync/functional-spec.md
```diff
@@ -0,0 +1,675 @@
...
+  attached (FR-Obs-1).
+*`[CLI]` `syncstrategy/base.py` → `BaseSync.determine_should_sync` / `register_strategy` / `use_sync_strategy` with the `choosing-s3-sync-strategy` event, consumed by `subcommands.py` → `CommandArchitecture.choose_sync_strategies`. The three slots are `file_at_src_and_dest`, `file_not_at_dest`, `file_not_at_src` (`VALID_SYNC_TYPES`).*
+
+**FR-Cmp-7** Timestamps MUST be compared to the whole second, and each mode MUST state what it does with
+the fraction of a second a local filesystem records and S3 does not. Once a file and an object agree, they
+MUST keep agreeing: repeated runs over an unchanged pair MUST NOT flip between transferring and skipping.
+*`[CLI]` `subscribers.py` → `ProvideLastModifiedTimeSubscriber` writes back `int(time.mktime(last_modified.timetuple()))` — whole seconds — via `utils.set_file_utime`; `syncstrategy/base.py` → `total_seconds` yields float seconds, so `ExactTimestampsSync` demands exact equality. `[DERIVED]` non-oscillation follows only because S3 `LastModified` is second-granular and the write-back matches it.*
+
+**FR-Cmp-8** The same two sides and the same options MUST always produce the same plan. It MUST NOT matter
```
aajtodd: "It MUST NOT matter what order the entries were listed in..." 

Doesn't this conflict with other requirements around keys in alphabetical order (FR-Root-7, FR-Enum-13).

aajtodd (approved):
Overall looks good.

docs/design/sync/functional-spec.md
ysaito1001: Thanks for calling that out, I meant to mention it. I'll backfill S3 → S3 `Copy` once the overall sync work lands. I did do a quick spike implementation of `Copy` to confirm it can be treated as just another sync-supported operation, so leaving it as a placeholder here shouldn't bite us back.

docs/design/sync/functional-spec.md
ysaito1001: Currently leaning toward (1), mainly because the plan is more than a pure function of the two listings; it also depends on what enumeration couldn't see. For instance, if a local directory is unreadable we have to hold back deletes for everything under it, since we don't know what's in there. Anyone rebuilding the plan from the walk primitives would have to rebuild that rule too, which can be subtle and easy to get wrong.

It also means the plan and the real run share one code path, so a fix lands in one place.

docs/design/sync/functional-spec.md
ysaito1001: Good point, this has recently come up. I take it the cross-SDK decision is still open, so I'll add an `FR-Root` requirement saying that there is this use case, and the design must not preclude whatever behavior we settle on.

Specified in bbd4584

docs/design/sync/functional-spec.md
aajtodd: If we go that route we don't want to be responsible for the formatting/output logging, we'd want to expose it as a stream of events or tied into transfer progress API somehow (most likely)

docs/design/sync/functional-spec.md
ysaito1001: Yes, and it can be worse than that; everything would be unknown from the last key we observed onward, since the next page's token only comes back inside the current page's successful response, and no token for an error response.

Chasing that, I think the spec is missing a rule. Each key has three states per side:
- **Present** — enumeration produced it
- **Absent** — enumeration passed its position in key order without producing it
- **Unknown** — enumeration failed somewhere covering it

Once a key is `Unknown` on either side, no action is licensed for it. Under abort mode the run stops anyway, so this matters for continue mode (default):
- `Unknown` source, `Present` destination → don't delete. That's `FR-Fail-7` today.
- `Present` source, `Unknown` destination → don't transfer, since that skips the comparison. Missing today, and it's the worse one: it can overwrite a newer destination object, and it breaks `--no-overwrite` outright.

Updated in 08076e7.

docs/design/sync/functional-spec.md
ysaito1001: Yeah, good catch. `FR-Fail-4` contradicts `NFR-Mem-1`. It asks the result to carry "the list of per-entry
errors" and "the outcome of each individual delete", both unbounded, while `NFR-Mem-1` says memory can't grow with entry count.
  
A million error records probably aren't a million pieces of information. If a million entries failed it's almost certainly the same cause repeated, e.g. `AccessDenied` on a prefix, credentials expiring mid-run. What's probably more useful is "247,000 failures, AccessDenied, all under logs/2019/".
  
So `FR-Fail-4` needs amending: the result carries counts and byte totals as it does now, plus failures grouped by cause and common prefix, which is bounded by the number of distinct causes. Per-entry detail goes out as events while the run is going, over the same bounded channel `FR-Obs-*` already needs.

A slow reader fills the bounded channel and misses per-entry events, but the aggregate in the result still tells them what happened.

Clarified in 6994697

docs/design/sync/functional-spec.md
ysaito1001: Ah, the wording in `FR-Cmp-8` is misleading. Ordering here means interleaving: the plan MUST NOT depend on how the two sides interleave, where the order in which earlier transfers finished. Key order on each side is mandatory, as you say. ~I'll reword it~ Updated in 1c452de.

docs/design/sync/functional-spec.md
ysaito1001: You're right, and this is TIL. `Restore` is a HeadObject header — the listing carries `RestoreStatus`, and only when the request sends `x-amz-optional-object-attributes: RestoreStatus`. ~Will update `FR-Fail-1` description accordingly.~ Updated in d64a00a.

docs/design/sync/functional-spec.md
ysaito1001: Agreed. It's also related to the million-entry case; we don't want to render a million decisions, instead we likely hand back a stream and let the caller decide how far to go.

Similar to the `FR-Fail-4` / `NFR-Mem-1` problem where one requirement promises completeness, the other promises memory that never grows. ~Will edit FR-Dry-1 too.~ Edited in 2992c1

docs/design/sync/functional-spec.md
ysaito1001: Updated in 9386edd.

PR #176, opened 2026-09-11
buffer pool: upload integration

aajtodd:
## Summary

This PR makes the shared buffer pool the production staging path for uploads. Multipart file reads, file-backed `PutObject` bodies, adapted asynchronous readers, and custom `PartStream` implementations can produce pool-owned data. Multipart request bodies remain retryable when that data spans several immutable segments; they do not need to gather the payload into another contiguous allocation first.

It also changes custom-stream scheduling so a source that returns `Poll::Pending` does not retain a scheduler execution slot. Sources that can produce parts immediately continue directly into concurrent `UploadPart` requests.

This PR also subsumes the unknown-content-length upload behavior and tests from [#164](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/164), integrating them into the new upload state machine rather than retaining the parallel implementation.

This is the upload-integration layer in the buffer-pool series:

- **This PR:** upload staging and retry integration

The detailed memory, ownership, and upload-path contracts are documented in `docs/design/memory.md`.

## Why

The preceding PRs made download admission and storage one lifecycle, but upload staging remained outside that boundary. Multipart file reads allocated separate heap buffers, file-backed `PutObject` bodies used independent chunks, and custom streams had no way to request transfer-manager-owned writable storage. The configured memory policy therefore did not cover the complete transfer-manager data path.

Download could reserve a range's known response length in `poll_work` before dispatching its network request. Upload could not use that boundary uniformly: an input might already own immutable bytes, a file reader might need one complete part-sized buffer, or a custom source might acquire several smaller envelopes as data became available. Memory admission therefore had to remain behind the part-reader abstraction instead of becoming a universal scheduler-side reservation.

Custom streams also awaited caller-provided data inside scheduled upload work. Enough externally blocked streams could consume the scheduler's execution capacity and prevent a later stream from running even when that stream had data available. This was the liveness failure represented by RUST-1178 and the Mountpoint many-upload regression.

Upload integration had to preserve two properties while closing those gaps. A source that could produce parts immediately still needed to feed concurrent network requests without another scheduler round trip, and an SDK retry still needed to restart the request body from byte zero without copying segmented pooled data into another allocation.

## Overview

An upload part now moves through one ownership path:

```text
file, adapted reader, or custom source
        |
        | memory admission and mutable filling
        v
PartData backed by SegmentedBytes
        |
        | retryable contiguous or segmented SDK body
        v
UploadPart request
        |
        | final immutable-owner release
        v
buffer-pool capacity returned
```

In-memory `Bytes` supplied by an application remain valid input and retain the SDK's native contiguous request-body path. Sources that need transfer-manager-owned writable storage use the shared pool before publishing immutable part data.

The admission point intentionally differs from download. Download reserves a known response range in `poll_work` before request dispatch. Upload `poll_work` schedules a part operation without assuming how that source stores its data. Multipart file readers acquire a complete part-sized envelope while executing the read, adapted readers acquire through `PartBuffer`, and custom streams decide whether and how much pooled storage to acquire through `StreamContext`. If memory admission or source readiness returns `Pending`, the unfinished part read is retained and its scheduler execution slot is released.

### Upload scheduling

The scheduler may dispatch several part operations for one transfer so source production and earlier network requests can overlap. If a source produces a part immediately, the execution proceeds directly into `UploadPart`; mutable access to the source does not serialize the resulting network requests.

If the source returns `Poll::Pending`, the unfinished `NextPartFuture` is retained and the execution returns `WorkOutcome::Yielded`. The scheduler releases that execution slot without reporting completed I/O or failure. The source's waker later makes the same read eligible again, preserving partial progress and notification state instead of reconstructing the source operation.

`PartStreamGate` protects exclusive access to one custom stream. While a poll is actively making progress, another execution can receive the stream through direct handoff and continue producing parts without another scheduling cycle. Once the source itself blocks, speculative executions retract their scheduled part accounting and return rather than accumulating independent futures behind the same blocked source. End-of-stream closes the gate so already-dispatched work cannot poll the source again.

`PartTransferState` keeps the state that must remain consistent across those transitions: speculative dispatch, retained source reads, active requests, completed parts, uploaded bytes, and end-of-stream.

`PartTransferState` also incorporates #164's unknown-content-length planning. `PartPlan::Unknown` keeps source dispatch open until end-of-stream, enforces S3's 10,000-part limit against parts actually produced, synthesizes the required empty part for an empty stream, and derives the completed object size from uploaded bytes.

### Pooled part production

Multipart file readers reserve the complete effective part range, acquire pooled writable storage, fill it with positioned reads, and publish only initialized bytes. Admission failure, short reads, I/O errors, and cancellation release unpublished storage.

Custom and adapted streams can create a `PartBuffer` through `StreamContext::part_buffer`. Creating the buffer does not reserve memory. `PartBuffer::poll_acquire` enters admission for a caller-selected logical envelope, and the producer retains the buffer in its own state across `Poll::Pending`.

The common path acquires `StreamContext::part_size()` once and fills one complete part. An externally paced source may use smaller envelopes to limit the memory retained by that source while input is incomplete. `PartBuffer` publishes completed envelopes internally and reuses writable carrier tails, so incremental producers do not assemble segments themselves or discard carrier-rounded capacity. `freeze` returns the complete immutable value as `SegmentedBytes`.

### Custom part streams

`PartStream::poll_part` is the advanced source interface. A source that already owns final immutable bytes can return `PartData::new`. A source that fills transfer-manager-owned storage retains a `PartBuffer` beside its own pending operation.

The following abbreviated example adapts an application-specific source that writes available data into a pooled part. `ExternalSource::poll_fill` represents the application's readiness interface: it may fill part of `dst` before returning `Pending`, but it must arrange for `cx.waker()` to be notified when it can make progress.

```rust
use std::io;
use std::pin::Pin;
use std::task::{ready, Context, Poll};

use aws_sdk_s3_transfer_manager::io::{
    PartBuffer, PartData, PartStream, SizeHint, StreamContext,
};

enum Fill {
    PartReady,
    EndOfStream,
}

// Example application-specific source.
struct ExternalSource {
    // Source-specific readiness and partial-fill state.
}

impl ExternalSource {
    fn poll_fill(
        self: Pin<&mut Self>,
        cx: &mut Context<'_>,
        dst: &mut PartBuffer,
    ) -> Poll<io::Result<Fill>> {
        // Poll the external source, write available bytes into `dst`, and
        // retain any partial operation needed to resume after `Pending`.
        ...
    }
}

struct ExternalParts {
    source: ExternalSource,
    buffer: Option<PartBuffer>,
    next_part: u64,
    size_hint: SizeHint,
}

impl ExternalParts {
    fn finish_part(&mut self) -> PartData {
        let data = self
            .buffer
            .take()
            .expect("completed part lost its buffer")
            .freeze();
        let part = PartData::from_segmented(self.next_part, data);
        self.next_part += 1;
        part
    }
}

impl PartStream for ExternalParts {
    fn poll_part(
        self: Pin<&mut Self>,
        cx: &mut Context<'_>,
        stream_cx: &StreamContext,
    ) -> Poll<Option<io::Result<PartData>>> {
        let this = self.get_mut();
        let buffer = this
            .buffer
            .get_or_insert_with(|| stream_cx.part_buffer());

        if let Err(error) = ready!(buffer.poll_acquire(cx, stream_cx.part_size())) {
            this.buffer = None;
            return Poll::Ready(Some(Err(error)));
        }

        match ready!(Pin::new(&mut this.source).poll_fill(cx, buffer)) {
            Ok(Fill::PartReady) => Poll::Ready(Some(Ok(this.finish_part()))),
            Ok(Fill::EndOfStream) if buffer.len() == 0 => {
                this.buffer = None;
                Poll::Ready(None)
            }
            Ok(Fill::EndOfStream) => Poll::Ready(Some(Ok(this.finish_part()))),
            Err(error) => {
                this.buffer = None;
                Poll::Ready(Some(Err(error)))
            }
        }
    }

    fn size_hint(&self) -> SizeHint {
        self.size_hint
    }
}
```

Applications with a Tokio `AsyncRead` source do not need to implement `PartStream`. The provided adapter performs memory admission, reads directly into pooled writable storage, and retains a partial part across `Poll::Pending`:

```rust
use aws_sdk_s3_transfer_manager::io::adapters::TokioIo;
use aws_sdk_s3_transfer_manager::io::{InputStream, SizeHint};
use tokio::io::AsyncRead;

fn input_stream<R>(reader: R, content_length: u64) -> InputStream
where
    R: AsyncRead + Send + Sync + 'static,
{
    InputStream::from_part_stream(TokioIo::new(
        reader,
        SizeHint::exact(content_length),
    ))
}
```

`PartData::with_checksum` continues to accept a caller-calculated part checksum, and `PartStream::full_object_checksum` continues to provide a final full-object checksum when requested by the selected strategy. A stream without an upper size bound remains multipart-only, reads until end-of-stream, enforces S3's 10,000-part limit, and synthesizes the empty part required for an empty multipart upload.

### Request bodies and retries

`PartData` stores `SegmentedBytes`. A value with one presentation segment continues through Smithy's native in-memory `SdkBody` path. A multi-segment value uses an exact-length retryable body that emits one owner-backed `Bytes` frame per segment.

Each SDK rebuild creates a fresh cursor over the same immutable owner set. A retry after partial body polling therefore starts again at the first segment, while the pool charges remain live until every request-body owner is released.

The current Smithy checksum path uses streaming aws-chunked trailer encoding for a multi-segment body. This preserves checksum behavior and retryability without gathering the payload, but it does not use the same checksum-header path as a contiguous body.

### File-backed uploads

Multipart file reads fill pooled writable runs through checked positional reads. Managed transfer-manager workers perform the read directly; shared Tokio runtimes offload the same operation through `spawn_blocking`.

File-backed `PutObject` remains a bounded streaming body rather than staging the complete object. It opens the source once before request dispatch, retains that file identity across retries, and reads at most 1 MiB of pooled data at a time. Managed workers use `DirectFileBody`; shared runtimes use `OffloadedFileBody`. Both emit owner-backed frames and reacquire pooled storage when the SDK reconstructs the body for a retry.

The old `io::Buffer` wrapper is removed after multipart file reads, adapted readers, and file request bodies move to the shared pooled-buffer path.

### Concurrent owner return

An upload part closes its reservation after publication, while its immutable request body retains carriers until the SDK releases every owner. Profiling found final carrier-return accounting beneath 33.7% of weighted stack samples.

Carriers were previously returned and accounted one at a time. This PR reduces that `Drop`-side work by skipping admission locking when the waiter queue is safely empty and batching uniquely owned whole-value returns when returns overlap or admission is queued. The queue's armed state and repayment epoch ensure that a racing enqueue either observes the repayment or is reconsidered by the return path.

Batched returns group carriers by concrete slot incarnation and bitmap word. Physical storage becomes reusable before reservation and aggregate accounting are released. Shared owners, mixed reservation provenance, progressive consumption, and ordinary uncontended returns retain the granular path.

### Diagnostics

The transfer manager resolves concurrency once when the client is constructed. A structured debug event reports the selected mode and final fixed target. Automatic mode also reports the detected instance type, usable vCPUs, estimate source, estimated throughput, and the sizing assumption used to derive the fixed target.

`AWS_S3_TM_DIAGNOSTICS=transfer.detail=1` enables one multipart upload summary containing source-read latency, pending-source duration, request timing, presentation-segment counts, body drain, and multipart create and complete timing. Level `2` additionally records scheduling and request transitions. The environment setting controls collection cost; tracing filters control whether collected `DEBUG` summaries and `TRACE` transitions are emitted.

## How to review

1. Begin with `UploadTransfer::poll_work` and `execute_upload_part` in `operation/upload/transfer.rs`. Read them alongside `PartTransferState` and `PartPlan` in `operation/upload/context.rs`. This is the central state machine: each scheduled part must either become an active `UploadPart`, retain one exact source future and yield, or retract or retire its accounting. The `Known` and `Unknown` plans should converge on the same completion path, with only end-of-stream and object-size discovery differing.

2. Follow source acquisition into `PartReader::start_part_read` in `io/part_reader.rs`. Its `Ready`, `Blocked`, and `Finished` results correspond directly to the branches in `execute_upload_part`. `NextPartFuture` preserves an unfinished read, while `PartStreamGate` serializes custom-source access without serializing the resulting network requests.

3. Follow storage production by source type. `PathBodyPartReader::next_part` covers multipart files. `PartBuffer::poll_acquire` defines custom and adapted-stream admission and tail reuse. `TokioIo::poll_part` is the concrete adapter showing how a producer retains pooled storage and partial progress across `Poll::Pending`.

4. Once `PartData` is produced, return to `UploadTransfer::send_part`, then follow `part_body::sdk_body`. The contiguous branch preserves the existing Smithy body path; the segmented branch must report the exact length, emit owner-backed frames in order, and rebuild from byte zero for retries.

5. Review `file_body.rs` separately for file-backed `PutObject`. It does not pass through multipart `PartData`, but must preserve the same pooled ownership and retry guarantees while retaining one open file identity.

6. Finally, review owner return independently from upload scheduling. `SegmentedBytes::drop` selects batching, and `CarrierReturnBatch::drop` must publish physical carrier availability before releasing reservation and aggregate accounting. Shared or incompatible ownership must remain on the granular path.

## Future Work

Multi-segment bodies currently use the SDK's streaming checksum and aws-chunked path. Hardening will compare a replayable segmented Smithy body with checksum-header traversal, transfer-manager checksum calculation while source data remains cache-hot, and the current trailer path without assuming that contiguity is the only useful outcome.

The portable buffered file path is the baseline for direct-I/O and storage work. `O_DIRECT`, `io_uring`, registered buffers, NUMA placement, checksum locality, source-file mutation behavior, and explicit durability guarantees require separate measurement and qualification.

Allocator hardening will evaluate search and placement independently. The objective is to reduce presentation segments and downstream vectored-I/O work without weakening admission progress, utilization, reclamation, or aggregate throughput.


aws-sdk-s3-transfer-manager/src/io/fs.rs
```diff
@@ -88,11 +169,32 @@ mod sys {
...
+    pub(super) fn read_exact_at(file: &File, dst: &mut [u8], mut offset: u64) -> io::Result<()> {
+        while !dst.is_empty() {
+            let count = file.seek_read(dst, offset)?;
+            if count == 0 {
+                return Err(io::Error::new(
+                    io::ErrorKind::UnexpectedEof,
+                    "unexpected end of file",
+                ));
+            }
+            dst = &mut dst[count..];
```
ysaito1001: `dst: &mut [u8]` is a mutable reference held by an immutable binding, this assignment to immutable `dst` might not compile on Windows (since gated that way). Should be a matter of updating `read_exact_at`'s signature to `mut dst: &mut [u8]`

aws-sdk-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -323,98 +492,245 @@ struct PathBodyReadCursor {
...
+                if result.is_none() {
+                    this.part_stream.gate.finish(stream);
+                } else {
+                    this.part_stream.gate.release(stream);
+                }
+                match result {
+                    Some(Ok(part)) => {
+                        this.part_stream
+                            .parts_yielded
+                            .fetch_add(1, Ordering::Release);
```
ysaito1001: `parts_yielded()` at line 194 documents: "Exact for custom streams because the count is advanced before the gate returns the stream." The code might break the invariant.

`release()` calls `notify_one()`, so a waiter can be scheduled on another worker immediately: task X releases the gate → waiter Y acquires the stream, polls, gets `Ready(None)`, and reaches `on_end_of_stream` → `parts.observe_end_of_stream(part_reader.parts_yielded())` (`upload/transfer.rs:526`) — all before X executes its `fetch_add`.

`PartTransferState::observe_end_of_stream` (`upload/context.rs:311`) returns true when `PartPlan::Unknown && parts_yielded == 0`, which makes the transfer synthesize and upload `PartData::new(1, Bytes::new())` (`upload/transfer.rs:536`) even though the source really produced part 1. Result: a duplicate part number in `CompleteMultipartUpload`, or part 1 silently replaced with zero bytes.

Fix: count the part before publishing the gate; hoist `if matches!(result, Some(Ok(_))) { parts_yielded.fetch_add(1, Ordering::Release); }` above the `finish/release` block and drop the increment inside the match.

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -654,37 +838,54 @@ impl UploadTransfer {
...
+                    bytes_uploaded <= declared_object_size,
+                    "uploaded {bytes_uploaded} bytes, more than declared upper bound \
+                     {declared_object_size}"
+                );
+                declared_object_size
+            }
+            PartPlan::Unknown => bytes_uploaded,
+        };
+        self.inner.ctx.set_total_bytes(object_size);
+
```
ysaito1001, line 902: For `PartPlan::Known`, `object_size` comes from `declared_object_size`, i.e. `stream.size_hint().upper()` — documented as an upper *bound*, not an exact size. The surrounding code accommodates a short source (`PartPlan::all_dispatched` short-circuits on `eof` "so a source
  that ends before its declared upper bound does not get polled repeatedly"; the `debug_assert!` only checks `bytes_uploaded <= declared_object_size`).
  
So when a known-length source ends early, every part uploads successfully and then CompleteMultipartUpload carries `x-amz-mp-object-size: <declared>` for a smaller uploaded byte count. S3 rejects that with HTTP 400 `InvalidRequest` — the transfer fails at the very last request, after all data has moved.
  
The base version did not set `MpuObjectSize` at all. It is reachable today; `tests/upload_test.rs::test_large_upload_part_size_bump` declares `SizeHint::exact(100 GiB)`, streams zero bytes, and passes only because the mocked client does not validate `mpu_object_size`.
  
**Fix:** derive `object_size` from `bytes_uploaded` (keeping `declared_object_size` for the `debug_assert!`), or only send `.mpu_object_size(...)` when `bytes_uploaded == declared_object_size`, and surface a typed client-side error rather than an opaque S3 400 when a `Known` plan ends short.

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
```diff
@@ -345,67 +379,209 @@ impl UploadTransfer {
...
+            }
+            Poll::Ready(Ok(Some(data))) => data,
+            Poll::Ready(Ok(None)) => {
                 tracing::trace!("part_reader exhausted");
-                self.maybe_transition_to_completing();
-                return WorkOutcome::Success { data: None };
+                return self.on_end_of_stream(&part_reader, timing).await;
             }
-            Err(e) => return self.fail(e.into()),
+            Poll::Ready(Err(error)) => return self.fail(error.into()),
```
ysaito1001, line 510: `PartStream::poll_part`'s contract: *"The transfer manager does not poll the stream again after end-of-stream or an error."* But `DynNextPartFuture::poll` calls `gate.finish(stream)` only when the result is `None`; for `Some(Err(_))` it calls `gate.release(stream)`, returning the gate to `GateState::Available`.
  
 Any other UploadPart execution already parked in `PartStreamGate::acquire()` is then handed the errored stream by `release`'s `notify_one()` and polls it again. The runtime's pre-execution
  `descriptor.is_terminal()` check only skips work that has not *started*. With `PartPlan::Unknown` this is the common case, since `poll_work` dispatches speculatively and several executions routinely sit in `acquire()`.
  
Consequence: user `poll_part` implementations that moved state out on error (e.g. the `Option::take` pattern) get an extra poll and can panic or return corrupt data.
  
**Fix:** in `io/part_reader.rs`, call `gate.finish(stream)` for `Some(Err(_))` as well as `None`, so later `start_part_read()` calls return `PartReadStart::Finished`.

aws-sdk-s3-transfer-manager/src/operation/upload/file_body.rs
```diff
@@ -5,104 +5,162 @@
...
+            io::Error::new(
+                error.kind(),
+                format!("failed to open {}: {error}", path.display()),
+            )
+        })?;
+        Ok(Self {
+            file: Arc::new(file),
+            pool,
+        })
+    }
```
ysaito1001, line 67: `FileBodySource::open` performs a synchronous `std::fs::File::open` and is reached from `InputStream::into_sdk_body` inside the async `UploadTransfer::execute_put_object` (`transfer.rs:740-744`).
  
The pre-change code deliberately offloaded exactly this call for the shared-runtime variant — the old `read_task` opened via `spawn_blocking(move || File::open(&path))` — and `RuntimeComponents::direct_io` documents the contract as *"When false, file I/O is offloaded via spawn_blocking (shared runtimes)"* (`runtime/mod.rs:73-75`). With `direct_io == false` the open now runs on a shared Tokio worker, where it can block for a long time on NFS/FUSE mounts — the Mountpoint-style workload this crate targets.
  
Every other blocking file operation in the new code is correctly offloaded (`read_task` wraps `read_exact_at` in `spawn_blocking`), so this open is the sole remaining blocking syscall on the async path.
  
  **Fix:** add an async constructor for the `!direct_io` case, e.g.
  `open_offloaded(path, pool) -> io::Result<Self>` wrapping the body in `spawn_blocking`, and await it from `into_sdk_body`'s non-direct branch (or hoist the open into `execute_put_object`).

s3-mock-server/src/storage/in_memory.rs
```diff
@@ -367,10 +377,14 @@ impl StorageBackend for InMemoryStorage {
         &self,
         request: crate::storage::CompleteMultipartUploadRequest<'_>,
     ) -> Result<crate::storage::CompleteMultipartUploadResponse> {
-        // Get the upload metadata
-        let (bucket, key, mut final_metadata, checksum_algorithm, checksum_type) = {
-            let mut uploads = self.multipart_uploads.write().await;
-            let upload = uploads
+        let (bucket, key, mut final_metadata, checksum_algorithm, checksum_type, upload_parts) = {
```
ysaito1001, line 380: The rewritten extraction block `remove`s both the upload metadata and the part data from `MultipartState` before any validation runs. If the part-ETag check (lines 400-410) returns `InvalidPart`/`NoSuchPart`, or the client-checksum check (lines 551-555) returns `ChecksumMismatch`, the `?` drops `upload_parts` and the upload ceases to exist.
  
Real S3 leaves the multipart upload intact on a failed CompleteMultipartUpload; the client may retry Complete or Abort it.
  
This is the backend the PR's upload tests assert against, so the divergence turns a precise failure into a misleading one: the transfer manager's follow-up `AbortMultipartUpload` (and any Complete retry) then fails with `NoSuchUpload`, which is the error a test author sees instead of the real `InvalidPart`/`ChecksumMismatch`. A genuine "we aborted an upload we should not have" bug in the new upload state machine would be indistinguishable from this artifact.
  
  **Fix:** read the upload and parts under the guard with `get`/clone, run all validation, and `remove` only after the object is committed to `buckets` — mirroring `abort_multipart_upload`, which removes only on success.

aws-sdk-s3-transfer-manager/src/scheduler/scheduler.rs
ysaito1001, line 540: This is pre-existing, not introduced by this PR, but `on_panic` never calls `generate_work()`, unlike `on_completion`. This might invalidate the documented retire-at-capacity guarantee: at capacity the runner retires trusting that an in-flight item will complete and re-drive generation. If that item panics instead, a transfer sitting in the ready set with generatable work and no pending wake hangs permanently.

ysaito1001 (commented):
Looks good over all, just leaving potential findings to share.

aws-sdk-s3-transfer-manager/src/io/fs.rs
aajtodd: e9eff4261c27c7da570cbe02a190a79802a81fd2

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
aajtodd: 31376f7 — semantic fix; be89c56 — final accounting hardening

aws-sdk-s3-transfer-manager/src/operation/upload/transfer.rs
aajtodd: bb5a94c

aws-sdk-s3-transfer-manager/src/operation/upload/file_body.rs
aajtodd: We have an existing issue for this, I added it to that. The default runtime is our managed runtime where this blocking work is expected. We'll have to move it async maybe for tokio-mt

s3-mock-server/src/storage/in_memory.rs
aajtodd: d53fb0c — state retention; 9b0385f — retry regression coverage

aws-sdk-s3-transfer-manager/src/scheduler/scheduler.rs
aajtodd: dce2760 — scheduler: refill work after task panic

aws-sdk-s3-transfer-manager/src/io/part_reader.rs
aajtodd: be89c56

ysaito1001 (approved):

PR #177, opened 2026-09-15
sync: key-ordered enumeration

ysaito1001:
## Summary
This PR adds the first layer of the `s3 sync` operation: a pull-based stream yielding `(relative key, metadata)` pairs in `ListObjectsV2` order from either a local directory or an S3 prefix, so the two sides of a run can be merge-joined (merging comes in a later PR). Key derivation moves out of `upload_objects` and `download_objects` into `io/key.rs` for both to share, `io/key/stream.rs` adds the entry shape and the stream trait both walkers implement, `io/key/filter.rs` adds include and exclude rules matched on the relative key, and `FsWalk` gains a `SortOrder` that selects the traversal a merge needs.

Two shapes carry information that had been riding the error channel. `FsWalk` now yields an `FsEntry` describing what the filesystem said is at a path — a regular file, a socket, a symlink it was told not to follow — so a name a transfer cannot move still occupies its key, which is what stops a later delete from treating it as absent. And `KeyStream` reports a `StreamError`, which separates a walk's failure from the two the stream itself has: a local name no S3 key could carry, and a listed object missing a field a comparison needs.

`docs/design/sync/enumeration.md` is the design ([rendered](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/6ad436f1343faa1edff44e783c80f8e267d2f3f1/docs/design/sync/enumeration.md)). It describes what this layer is and why that set of requirements forms one unit, the design decisions and what each would cost to change, and how the pieces fit. Requirement coverage and benchmark figures are in this description rather than in the doc.

 The existing directory operations keep their traversal and their entry set, though their code changes: 218 lines leave `upload_objects` and `download_objects` as key derivation moves to the shared module, and both take the renamed entry type.

Behaviour does change, in what a walk and an upload report rather than in what they transfer:
- A subdirectory that turns out not to be a directory between being listed and being read used to end the walk; it now costs that one name.
- A directory that cannot be opened was reported as `PermissionDenied`, which names one entry. It is now `DirectoryUnreadable`, which says a subtree went unenumerated.
- A walk used to hand over the entries it already held before reporting a failure. It now reports the failure first, so a failure arrives at the position it happened.
- A fatal failure used to be reported while the walk carried on yielding whatever it still held, then ended with a clean `None`. It now ends the walk.
- `upload_objects` checks the file type before building a stream, so it skips a socket or a FIFO. Nothing asks a walk for those today, so the check guards what the entry type now permits. It also runs before key derivation, so a name holding a custom delimiter no longer ends a transfer over an entry nothing was going to upload.
- An object a listing gave no key for is reported whether or not a filter is installed. A filter decides what is selected, never what is reported.
- An excluded name whose type could not be read reports the range it covers, since it may be a directory and a rule on a name says nothing about the keys under it.
- Building an upload body no longer re-reads metadata the caller supplied, so it no longer fails on a path removed after the walk.

## Review order
Start with `docs/design/sync/enumeration.md` — "Where this sits" and §1 give the shape and the scope. Then the code, in the order each piece depends on the one before it:

**1. The ordering rule.** `cmp_key_form` and `SortOrder` in `io/walk/fs.rs`. Everything else rests on this: a directory is compared as if its name ended in `/`, which is what places `a.txt` before `a/c`. "The sort trick" in §3 of the design explains why. `SortOrder` replaces a boolean that could not say which of three orders was wanted,

**2. What a walk yields.** `FsEntry` and `FileType` in `io/walk/fs.rs`. An entry now says what the filesystem found — `Regular`, `Symlink`, `Fifo`, `Socket`, `Device` — whether a symlink was resolved to reach it, and carries metadata as an `Option` for the case where the filesystem will not describe what is there. This is what let two error kinds go: a socket arrives at its own key instead of on the error channel, which is what holds back a later delete.

**3. What a walk reports as a failure.** `WalkErrorKind` in `io/walk/error.rs`. Only `DirectoryUnreadable` is new, and the distinction to check is against `SourceUnreadable`: the same failure at the root leaves nothing to walk, one level down the walk carries on but a subtree went unenumerated. `is_fatal` is unchanged from before this branch, byte for byte.

**4. The keyed stream.** `Entry`, `EntryMeta` and the `KeyStream` trait in `io/key/stream.rs`, plus the two impls on `FsWalk` and `S3Walk`. `EntryMeta` is deliberately small, an optional size and an optional time in whole seconds, because S3 reports last-modified at second granularity and finer local precision would make an identical pair differ every run. Either field is absent when the side could not describe the entry, and a comparison that reads absence as a value decides wrongly: `None` sorts below every `Some`, so a source with no time would read as the older of the pair.`Entry` carries the walker's own item alongside the key, since a comparison reading a checksum or a storage class needs fields the key cannot hold.

**5. What the stream reports as a failure.** `StreamError` in `io/key/stream.rs`. Three variants, because deriving a key and reading a listing are the stream's own work and a walk reports neither. Worth checking: an object missing a size used to become `WalkErrorKind::Service`, which is fatal, so one malformed object in a page stopped a whole run. `keys_lost` is where a consumer asks what a failure cost, and it answers in three: nothing, one key,
  or an unknown range.

**6. Root-relative keys.** `root_prefix` and `relative_key` in `io/key/stream.rs`, and `derive_object_key` in `io/key.rs`. `relative_key` refuses a key outside the root, which is what keeps the order intact once the prefix comes off. `strip_key_prefix` cannot serve here: with the prefix `data` it turns `data/z` into `z` and `datab/x` into `b/x`, and that reverses them — `data/z` sorts before `datab/x`, while `b/x` sorts before `z`.

**7. Filters.** `KeyFilter`, `Rule`, `Anchor` and `glob_match` in `io/key/filter.rs`, then `local_predicate` and `s3_predicate` in `io/key/stream.rs`. The property to check is that one rule set gives both sides the same answer for the same key, so an excluded key cannot look like a deleted one. FR-Filter-5 is the subtle one: an excluded entry stays silent even when it cannot be read. What it does not cover is a name whose type could not be read at all — that may be a directory, and excluding a name says nothing about the keys under it, so the failure is reported.

**8. That it composes.** `identical_sides_plan_no_transfers_or_deletes` in `io/key/stream.rs` points a merge-join at a directory and a prefix holding the same content and shows nothing is planned. That is the end-to-end claim this layer exists to support.

Two answers are in comments rather than in code. Fatality is a property of *where* a failure happened, and the walk already publishes that as `is_done`, so moving consumers onto it is a follow-up rather than a change to two shipped operations. And peak-memory benchmarking is dropped from this PR, since it duplicates work covered elsewhere.

## Requirement coverage
28 requirements across `FR-Root`, `FR-Enum` and `FR-Filter`.

<details>
<summary>Click to expand the coverage table</summary>

| Requirement | Status | Where it stands |
|---|---|---|
| FR-Root-1 Upload, Download and Copy | partly | `FsWalk` and `S3Walk` both implement `KeyStream`, so a direction is a choice of two streams; `Copy` pairs two `S3Walk`s and is backfill work |
| FR-Root-2 a root names a place, never one entry | later phase | root parsing and validation |
| FR-Root-3 relative key derivation | covered | `io/key.rs` |
| FR-Root-4 a trailing separator changes nothing | covered | `root_prefix` normalises it, pinned by test |
| FR-Root-5 any bucket identifier the API accepts | later phase | root parsing and validation |
| FR-Root-6 two S3 endpoints for `Copy` | later phase | root parsing and validation |
| FR-Root-7 reject a bad configuration first | later phase | root parsing and validation |
| FR-Root-8 a key resolving outside the root | kept open | settled elsewhere; nothing here forecloses it |
| FR-Enum-1 list both roots through | covered | only filters reduce what gets listed |
| FR-Enum-2 folder markers invisible | covered | dropped in the stream, so no filter can bring them back, pinned by test |
| FR-Enum-3 skip one entry, and say which kind | covered | an entry carries its `FileType`, so a skipped kind is named where the key is |
| FR-Enum-4 symlinks off by default | inherited | `follow_symlinks` unchanged; an unfollowed link arrives as an entry saying it is one |
| FR-Enum-5 a time the platform cannot represent | partly | `last_modified_secs` is an `Option`; never-skip is the comparison's |
| FR-Enum-6 keys compared byte for byte | covered | no folding, no normalizing, pinned by test |
| FR-Enum-7 start before listing finishes | partly | an entry arrives before the second listing page is fetched, pinned by test; nothing starts transfers yet |
| FR-Enum-8 files added or deleted mid-run | inherited | unchanged |
| FR-Enum-9 a symlink loop ends the descent | covered | `SymlinkCycle`; the ancestor-handle check survives depth-first |
| FR-Enum-10 markers become local directories | kept open | one check in `key_and_meta` is where the opt-in has to reach |
| FR-Enum-11 configurable page size | inherited | unchanged |
| FR-Enum-12 an unreadable directory is a range | covered | `DirectoryUnreadable`, positioned in key order |
| FR-Enum-13 listing sends no delimiter | inherited | unchanged |
| FR-Filter-1 ordered include and exclude rules | covered | `io/key/filter.rs` |
| FR-Filter-2 matched on the whole path, `*` crosses `/` | covered | `fnmatch` semantics reproduced |
| FR-Filter-3 the rule set applies to both sides | covered | one `KeyFilter`, read by `local_predicate` and `s3_predicate` |
| FR-Filter-4 order-independent, holds neither side | covered | matched per entry as it arrives, with the same answers in any order, pinned by test |
| FR-Filter-5 an excluded entry stays silent | covered | consulted before metadata; a name of unknown type still reports, since it may be a directory |
| FR-Filter-6 anchored or matched anywhere | covered | the API says which; the pattern text keeps its `fnmatch` meaning |
| FR-Filter-7 delete what the filters excluded | kept open | `s3_predicate` is the only place to invert |

</details>

## Benchmarks
`benches/walk.rs`, one host and one build: Amazon Linux 2023, x86_64, 6.12 kernel, Xeon 8488C, 63 GB, fixtures on xfs over an EBS volume (confirmed with `df -T`), machine 97% idle. Three arms over the same tree: the default order, which sorts nothing and is what this branch started from, and the two sorted orders. Caches stay warm throughout — `sudo` was unavailable, so nothing could be dropped — which makes every figure a floor.

**Requiring key order costs a few percent of throughput at most, and on a wide tree it costs nothing.**
```
   shape                   default     within   whole_walk
   wide_1dir_10k          20.55 ms   20.45 ms     18.20 ms
   deep_100dirs_100each   55.29 ms   57.97 ms     56.44 ms
   balanced_f10_d3        54.48 ms   55.05 ms     54.60 ms
```

Against the default, a whole-walk order adds about 2% on the deep tree and a fraction of a percent on the balanced one, and comes out 11% faster on the wide one — the shape where a per-directory sort of 10,000 entries should have cost the most. Why sorting is free there is unexplained: the depth-first pass appears to do less per entry than the unsorted one, by enough to pay for the sort.

Reversing the order the arms register in kept the sign of all nine comparisons and moved the magnitudes — the deep tree's 2% became 0.4%, the balanced tree's 0.2% became 5.8% on the widest confidence interval in the set, and the last arm to run on the wide tree is about 3% slower whichever arm it is. The direction is measured; the percentages are not precise to a decimal place.

**Time to the first entry does not grow with the tree, where the shape holds the leftmost path fixed.** That is what this sweep does, so a rising curve would mean the walk reads more than it needs to before emitting. It is evidence that nothing buffers the tree, not a latency bound.

```
   entries      within   whole_walk
     1,000    52.08 µs     52.19 µs
    10,000    52.19 µs     51.60 µs
    50,000    52.02 µs     51.81 µs
```

**What a whole-walk order pays for its first entry depends on where the first key sits.**

```
                             within   whole_walk
   front_loaded_100dirs     61.0 µs     102.7 µs
   chain_100deep             201 µs      56.1 ms
```

Where the first key in order is one directory away, the cost is about 40 µs. In a chain it is at the bottom — `d1/` sorts before `f0000.dat` — so the walk descends all 100 directories first, and each one it opens enumerates that directory's 100 files: 10,000 entries read before one is emitted. That is 56 ms, about what a full walk of a 10,000-entry tree costs, and 279× what ordering within a directory pays on the same tree.

This follows from emitting in key order rather than being a defect in the walk: nothing can name the first key without reading every directory that might hold it. D11 in the design gives the same product as the memory bound — the depth of the leftmost path times the width of each directory on it — and it bounds this latency too.

Most of that 56 ms is metadata the ordering never asked for. The walk stats a file as soon as it sees the name, so reaching the bottom of the chain stats all 10,000 files before handing over even one — and choosing the order needs only the name and whether it is a file or a directory. Waiting until a file is about to be handed over would make that descent cost the 100 directory reads and nothing else.
  
Waiting would also move where a failure appears. A file the walk cannot stat is found today while reading the directory that holds it, and reported there, which in key order can be well before the keys that sort ahead of it; waiting reports it where that file's own key belongs. That is a change to how failures are ordered, which is what this PR settles, so we track it with a [TODO(walker)](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/ea52f7833b837e0a990eb4df59831cd5a6210302/aws-sdk-s3-transfer-manager/src/io/walk/fs.rs#L1048-L1053) and address it in a subsequent PR.

The largest tree measured is 50,000 entries, where the requirements ask about a million. The sweep grows the total while holding root fanout and leftmost depth fixed, and those are what first-entry latency depends on, so the flat curve shows that the walk does not buffer before emitting rather than projecting to a million keys.

## What follows
This is the first of a series of 4 to 5 PRs. **Comparison** likely comes next: it consumes the two streams and decides per key whether to transfer, delete or leave, and it is where the comparison interface lands including the ability for a mode to answer "not yet, ask me again", so a checksum comparison can fetch what a listing does not carry. **Execution** follows, wiring those decisions into the scheduler as a composite transfer. **Deletes and failure handling** come last, since a delete is only safe once the stream can say which keys it could not see. Sync's `Copy` direction is backfill work, landing after the rest of sync.

## Testing
- Each of the functional requirements the design marks `covered` or `partly` is pinned by at least one unit test.
- The suite and the benchmarks ran on Linux, which is where the non-UTF-8 filename test can execute.


aws-sdk-s3-transfer-manager/src/io/walk/error.rs
```diff
@@ -72,6 +103,11 @@ impl WalkError {
         self.kind
     }
 
+    /// How this error bears on the run. Equivalent to `self.kind().severity()`.
+    pub fn severity(&self) -> Severity {
```
aajtodd: Isn't this more a property of the walk's configuration? I'm not sure this belongs on the error itself and might need to stay with the transfer state machine?

aws-sdk-s3-transfer-manager/src/io/walk/error.rs
```diff
@@ -5,49 +5,80 @@
...
     BrokenSymlink,
     /// Symlink whose target is a directory already on the current descent
     /// path (a cycle). Non-cyclic duplicate symlinks (two different symlinks
     /// to the same target) are not reported as cycles and are traversed
-    /// normally. Non-fatal.
+    /// normally.
     SymlinkCycle,
+    /// A socket, FIFO, or block or character device. A walk yields regular files,
+    /// so there is no entry this path could produce.
+    SpecialFile,
```
aajtodd: `SymlinkNotFollowed`/`SpecialFile`/`NonUtf8Name` aren't actually errors they're just being used to carry information riding the error channel. I'm wondering if these need to be handled differently

aws-sdk-s3-transfer-manager/src/io/walk/error.rs
```diff
@@ -5,49 +5,80 @@
...
-    /// Non-fatal.
     PermissionDenied,
-    /// Symlink encountered with no valid target. Non-fatal.
+    /// A directory could not be read: the `read_dir` itself failed, or opening it
+    /// for cycle detection did. The walk continues with the rest of the tree, but
+    /// that subtree was never enumerated, so a consumer that infers absence from
+    /// the stream must decide for itself whether to keep going. Distinct from an
+    /// entry-level failure so that decision is possible. The underlying
+    /// `io::Error` remains available via [`std::error::Error::source`].
+    DirectoryUnreadable,
```
aajtodd, line 31: How is this different from SourceUnreadable?

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
```diff
@@ -305,21 +403,51 @@ impl FsWalkerBuilder {
...
+    /// directory before descending, so it can never place a nested key between
+    /// two siblings.
+    ///
+    /// Implies [`sort`](Self::sort). Two costs come with it:
+    /// [`try_claim_subtree`](FsWalk::try_claim_subtree) is unavailable, since
+    /// ordered emission has to descend before it can know what comes next, and
+    /// time-to-first-entry grows because subtrees sorting ahead of a sibling file
+    /// must be read first.
+    #[must_use]
+    pub fn key_order(mut self, key_order: bool) -> Self {
```
aajtodd: This is what `sort` was meant for. The deleted doc comment even says it:

```rs
    /// Sort is required for `sync`-style operations that merge-join against
    /// `ListObjectsV2` results (which are UTF-8 binary sorted by key).
```

I don't love `key_order` for a few reasons outside of that though. It claims something about S3 ListObjectsV2 that isn't true for directory buckets which is confusing. 

---

Seems like we should just "fix" `sort()` or change it to an enum like

```rs
enum SortOrder { Native, Lexicographical}

pub fn sort(mut self, sort: SortOrder) -> Self { ...}
```

or something.

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
```diff
@@ -305,21 +403,51 @@ impl FsWalkerBuilder {
...
+
+    // Report a path the walk cannot yield an entry for: a socket, a FIFO, a device, a
+    // symlink left unfollowed. Off by default, because a caller that only uploads what
+    // it finds has no use for them, and a tree can hold thousands. A consumer that
+    // infers absence from the stream needs them, since a name it never hears about
+    // reads as a name that is free.
+    //
+    // Only tests call this until the comparison lands.
+    #[allow(dead_code)]
+    pub(crate) fn report_untransferable(mut self, report: bool) -> Self {
```
aajtodd: This comes back to the ErrorKind objection but also I don't love "untransferable" on the walk abstraction. The type just controls walking not transfer decision or policy. I'm ok keeping a flag on config to control it for the pathological cases listed above but maybe renamed

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
```diff
@@ -354,15 +482,34 @@ impl FsWalkerBuilder {
...
+    // Unlike `filter`, which sees a fully built `DirEntry`, this is consulted first,
+    // so a rejected file is never stat'd and anything that went wrong reading it is
+    // never reported. That is the difference between "excluded" and "excluded but it
+    // still warned about a file the caller said to ignore".
+    //
+    // Directories are not offered to it: skipping one would skip everything beneath,
+    // and a rule written for a file's key says nothing about the keys under a folder
+    // of a similar name.
+    #[allow(dead_code)]
+    pub(crate) fn path_filter(mut self, f: impl Fn(&Path) -> bool + Send + Sync + 'static) -> Self {
```
aajtodd, line 554: Fine for now. One forward/future thinking idea (like after GA) is to support `.gitignore` one day. Nothing to do now just thinking out loud.

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
```diff
@@ -460,9 +606,6 @@ impl FsWalk {
     /// - `None` when the walk is complete.
     pub async fn next(&mut self) -> Option<Result<DirEntry, WalkError>> {
```
aajtodd: As I poke around this a few things stick out and the biggest one is the error channel being used for propagating some information we need as well as making the walker do classification.

I'm wondering if we just modify the result type from `DirEntry` to something like:


```rs

pub struct FsEntry { /* path, root, ... */ }

impl FsEntry {
    pub fn path(&self) -> &Path;
    pub fn relative_path(&self) -> &Path;
    pub fn file_type(&self) -> FileType;          // regular | symlink | socket | fifo | device | dir
    pub fn metadata(&self) -> Option<&Metadata>;  // None when not stat'd or not stattable
    pub fn followed(&self) -> bool;               // symlink resolved, or left alone
    pub fn is_cycle(&self) -> bool;                  // descent stopped here
}
```

I think that would allow us to reduce/remove the added error kinds and to make these entries representable as normally yielded entries/items. It also buys additive evolution, a new file type or traversal outcome isn't a new variant that ripples. 

`NonUtf8Name` is constructed in `key_stream.rs` not in `fs.rs` so a walker already never produces that error kind, it's a sync/s3 issue.

aws-sdk-s3-transfer-manager/src/io/mod.rs
```diff
@@ -15,6 +15,13 @@ mod stream;
 /// Error types related to I/O abstractions
 pub mod error;
 pub(crate) mod fs;
+pub(crate) mod key;
```
aajtodd, line 18: I'm fine with extracting these, key.rs conslidation is an improvement, `DEFAULT_DELIMETER` was declared twice. I don't know that `io/` module is right placement for it or not though (not that I have a better one). It's maybe ok for now given it's `pub(crate)` only.

Wherever we put it I'd also maybe consider 

```
key.rs
key/
    stream.rs
    filter.rs
```

if these are all related modules/concepts.

aws-sdk-s3-transfer-manager/src/io/mod.rs
```diff
@@ -15,6 +15,13 @@ mod stream;
...
+// Reached only by its own tests until the comparison wires it up.
+#[allow(dead_code)]
```
aajtodd: Consider a TODO(sync): ... so we don't lose track of these allowances?

aws-sdk-s3-transfer-manager/src/io/mod.rs
aajtodd: Also many functions/types/etc are missing doc comments in some of these new modules (or in relocated but now pub(crate).

aws-sdk-s3-transfer-manager/src/io/key_stream.rs
```diff
@@ -0,0 +1,934 @@
...
+// sets, which would otherwise make an entry out of a key holding nothing.
+fn key_and_meta(
+    obj: &Object,
+    prefix: Option<&str>,
+) -> Result<Option<(String, EntryMeta)>, &'static str> {
+    if !exclude_s3_folder_markers(obj) {
+        return Ok(None);
+    }
+    let key = obj.key().ok_or("listing returned an object with no key")?;
+    let relative = strip_key_prefix(key, prefix, None);
```
aajtodd: `strip_key_prefix` isn't order preserving unless the prefix ends in the delimiter. We don't want to (can't) modify `strip_key_prefix` though as it's behavior is correct for download.

e.g.
```
Prefix `data`, bucket holds `data/a`, `data/z`, `datab/x`, `datafile`. S3 returns those in byte order
(`/` is 0x2F, `b` is 0x62, `f` is 0x66). Stripping gives:

data/a    -> a
data/z    -> z
datab/x   -> b/x        <- z > b/x, no longer sorted
datafile  -> datafile   <- early return at key.rs:67, no delimiter in the key
```

aws-sdk-s3-transfer-manager/src/io/key/stream.rs
```diff
@@ -0,0 +1,934 @@
...
+// An unread subtree makes its side look emptier than it is, which position-based
+// comparison cannot tell apart from deletion. A single failed entry is just one key
+// unaccounted for.
+pub(crate) fn view_incomplete(err: &WalkError) -> bool {
+    err.is_fatal() || err.kind() == WalkErrorKind::DirectoryUnreadable
+}
+
+// `None` when the time is missing, or is far enough outside this platform's range
+// that it cannot be expressed as seconds from the epoch.
+fn secs_since_epoch(modified: std::io::Result<SystemTime>) -> Option<i64> {
```
aajtodd, line 206: `duration_since` returns `Err` when the argument is later than the receiver, 

Any pre-1970 mtime becomes `None` silently then. Reachable in practice (bad timestamp, restored backup, manual mtime manipulation, etc).

Probably something like:

```rust
let t = modified.ok()?;
match t.duration_since(UNIX_EPOCH) {
    Ok(d)  => i64::try_from(d.as_secs()).ok(),
    Err(e) => i64::try_from(e.duration().as_secs()).ok().map(|s| -s),
}
```
Needs some tests probably.

benches/walk.rs
```diff
@@ -0,0 +1,419 @@
+/*
```
aajtodd, line 1: What are the benchmarks telling us on tmpfs? 

BFS and DFS have different access patterns that on a real device can change readahead and cache locality. May be worth running benchmark on a real disk/device mount rather than tmpfs.

benches/walk.rs
```diff
@@ -0,0 +1,419 @@
...
+            });
+        }
+    }
+    group.finish();
+}
+
+// Peak memory, measured by counting allocations in this bench crate. Criterion is not
+// involved: it runs many iterations and allocates as it goes, which would swamp the
+// figure being read.
+mod peak {
```
aajtodd: Do we actually need this? I would assume peak memory will be tested in our performance benchmarking. I'm not opposed necessarily but this may be unnecessary test code.

docs/design/sync/enumeration.md
```diff
@@ -0,0 +1,481 @@
+# Key-ordered enumeration
```
aajtodd, line 1: This doc is overall fine but it's carrying. alot of project state.

It's written for this moment rather than a description of what exists. I'd look at the other design docs which all describe what is, in present tense, and move the status to PR description if needed.

aws-sdk-s3-transfer-manager/src/io/key_stream.rs
```diff
@@ -0,0 +1,934 @@
...
+
+use aws_sdk_s3::types::Object;
+
+use crate::io::key::{derive_object_key, strip_key_prefix};
+use crate::io::key_filter::KeyFilter;
+use crate::io::walk::{
+    exclude_s3_folder_markers, DirEntry, FsWalk, S3Walk, WalkError, WalkErrorKind,
+};
+
+// Whole seconds, because that is the granularity S3 reports last-modified at.
```
aajtodd: Only commenting on this one but several doc comments in the PR could be improved. This one jumps right into "Whole seconds" and describing details which is fine but I would think we'd want to lead off with what the type is.

aws-sdk-s3-transfer-manager/src/io/walk/s3.rs
aajtodd: You didn't introduce this but reviewing has pointed out that this may not be cancel safe or that dropping the future may invalidate the walker. May need to consider restoring on error path or not calling `take()` until the call returns?

aajtodd (commented):
Good start. 

---

Another question I have which is maybe not for this PR but how we plan to consume this in sync transfer and how concurrency and keeping the pipeline full will work.

By construction now `try_claim_subtree` returns `None` so no parallel walk and `ListObjectsV3` pagination is serial per prefix. Are you imagining putting the walk and join behind an API/seam like:

```rust
// ---- config, mirrors FsWalker / S3Walker ----
pub struct SyncWalker { /* comparison mode, filter rules, delete on/off, concurrency knobs */ }
pub struct SyncWalkerBuilder { ... }

impl SyncWalkerBuilder {
    pub fn comparison(self, mode: impl ComparisonMode) -> Self;
    pub fn filter(self, rules: Vec<Rule>) -> Self;
    pub fn delete(self, enabled: bool) -> Self;
    pub fn build(self) -> SyncWalker;
    
    // not sure if this takes concrete walkers like `.source(FsWalk).destination(S3Walk) or if it is generic over KeyStream. Something has to normalize the join-level shape of entries though.
}

impl SyncWalker {
    pub fn builder() -> SyncWalkerBuilder;
    pub fn walk(self, ctx: SyncWalkContext) -> SyncWalk;   // same shape as FsWalker::walk
}

// ---- the two roots ----
pub struct SyncWalkContext { /* source, destination */ }

// ---- in-progress state ----
pub struct SyncWalk { ... }

impl SyncWalk {
    pub async fn next(&mut self) -> Option<Result<SyncDecision, Error>>;
    pub fn is_done(&self) -> bool;
}

and what comes out:

pub enum SyncDecision {
    Transfer { key: String, reason: TransferReason, source: Entry, destination: Option<Entry> },
    Delete   { key: String, destination: Entry },
    Skip     { key: String, reason: SkipReason },
}

pub enum TransferReason { SourceOnly, SizeDiffers, SourceNewer, ChecksumDiffers, PropertiesDiffer
pub enum SkipReason     { Unchanged, DestinationNewer, /* ... */ }

```

docs/design/sync/enumeration.md
ysaito1001: Point taken. Will rewrite it to describe what the code does in present tense, and move the requirement status table to the PR description. Ideally, it should be project state agnostic so someone reading it a year from now learns the design decisions without needing to know which PR introduced what.

Updated in 796ecf5.

aws-sdk-s3-transfer-manager/src/io/walk/error.rs
aajtodd: I'm still wondering if SourceUnreadable should just have it's doc comments relaxed so it can be used for either. Do we need to distinguish these (root vs directory below root)?

aws-sdk-s3-transfer-manager/src/io/walk/error.rs
```diff
@@ -39,46 +35,22 @@ pub enum WalkErrorKind {
...
-            | WalkErrorKind::SpecialFile
-            | WalkErrorKind::SymlinkNotFollowed
-            | WalkErrorKind::NonUtf8Name => Severity::EntryWarning,
-        }
-    }
-
-    /// Whether an error of this kind terminates the walk, which is
-    /// [`Severity::EndsRun`] seen from the walk's side.
+    /// Whether an error of this kind terminates the walk.
     pub fn is_fatal(&self) -> bool {
```
aajtodd: You didn't introduce this but related to `severity()` should we remove `is_fatal` or make it only `pub(crate)` for now. I think it's context/walk dependent

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
```diff
@@ -113,24 +134,53 @@ fn cmp_key_form(a: &Child, b: &Child) -> Ordering {
...
 
-/// A file entry discovered during a filesystem walk.
+/// What the filesystem said is at a path.
+///
+/// A walk yields an entry for whatever it found, so a consumer decides what to do with each
+/// one. Only a regular file holds bytes a transfer could move; the rest are named so a consumer
+/// can say why it left them alone.
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FileType {
```
aajtodd: Maybe move to `io/fs.rs` (same with fn special_file_type) since they are more generic FS types/functions?

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
```diff
@@ -401,50 +493,22 @@ impl FsWalkerBuilder {
...
-    /// per directory, not global: the traversal is breadth-first, so a
-    /// directory's files are emitted before its subdirectories are descended.
-    /// For a globally ordered walk see [`key_order`](Self::key_order).
-    ///
-    /// When `false` (default), entries are returned in OS-native order.
+    /// How to order the entries this walk produces. See [`SortOrder`].
     #[must_use]
-    pub fn sort(mut self, sort: bool) -> Self {
-        self.sort = sort;
+    pub fn sort_order(mut self, sort_order: SortOrder) -> Self {
```
aajtodd: (minor/nit/bikeshed): I prefer `sort(order: SortOrder)`over `sort_order(...)` personally 🤷‍♂️

aws-sdk-s3-transfer-manager/src/io/walk/error.rs
ysaito1001: Makes more sense. Removed in 02d9e0d. Severity and both severity() methods are gone, and is_fatal is back to exactly what it was before this branch.

aws-sdk-s3-transfer-manager/src/io/walk/error.rs
ysaito1001: Good call. 9db1d3d makes a socket, FIFO, device or unfollowed symlink come back as an entry carrying its `FileType`. f49a568 moves `NonUtf8Name` to the stream as `StreamError::UnkeyableName`.

aws-sdk-s3-transfer-manager/src/io/walk/error.rs
ysaito1001: `dir_error_kind` gives us `SourceUnreadable` at the root and `DirectoryUnreadable` below it (clarified in a3edade). The former terminates the walk, and the latter carries on from the next entry it can read. I introduced it for clarity, but we can consider collapsing them into one kind carrying a depth.

aws-sdk-s3-transfer-manager/src/io/mod.rs
ysaito1001: Added in 9a67d43

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
ysaito1001: Extra knob could be confusing. 8da49c0 replaced `sort` and `key_order` with `SortOrder { Native, WithinDirectory, WholeWalk }`, naming each value for the scope it orders. Needed three to cover per-directory ordering, without which 15 tests would fail; 13 of them assert which subtree `try_claim_subtree` hands over, and that is only deterministic when siblings are sorted. Claiming is unavailable under a whole-walk order (at least I haven't figured how to make them work nicely), hence three values for now.

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
ysaito1001: cfa0333 renamed it to `include_special_files`, which says what turning it on produces. It defaults off, so no changes to the directory operations as before.

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
ysaito1001: Good idea, I think this comment reinforces [this existing TODO](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/bd9ac62b51e6bea0a3e5940f606a16b1889827b4/aws-sdk-s3-transfer-manager/src/io/walk/fs.rs#L122)

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
ysaito1001: That's actually cleaner. `9db1d3d` incorporated it.

`NonUtf8Name` moved to the stream as `StreamError::UnkeyableName` in `f49a568`, which made it a sync-specific issue.

aws-sdk-s3-transfer-manager/src/io/mod.rs
ysaito1001: Updated in 9a67d43, and doc comments added in 013233d.

aws-sdk-s3-transfer-manager/src/io/key_stream.rs
ysaito1001: Ah, good catch. `strip_key_prefix` remains intact, and 22525e4 added `root_prefix` that functions as described above.

aws-sdk-s3-transfer-manager/src/io/key/stream.rs
ysaito1001: Fixed in 213e478

benches/walk.rs
ysaito1001: Re-ran them on xfs over an EBS volume, and 796ecf5 replaces the figures with those. A whole-walk order was faster than ordering within a directory on both tmpfs and xfs, but the margin shrank on xfs from as much as 14% to one or two percent.

benches/walk.rs
ysaito1001: Was more of a remnant from development. Removed in 796ecf5.

aws-sdk-s3-transfer-manager/src/io/key_stream.rs
ysaito1001: Updated in 4302bd7 to start with "what".

aws-sdk-s3-transfer-manager/src/io/walk/s3.rs
ysaito1001: Great find. Fixed in 9b571b4, and the token now stays until the page comes back. 32fe3de added more test that verifies the token is left in place when a future dropped mid-request.

ysaito1001:
> Another question I have which is maybe not for this PR but how we plan to consume this in sync transfer and how concurrency and keeping the pipeline full will work.

Yes, roughly that shape.
  
Some type (`SyncWalkContext` in your example) pairs the two key streams, so each key arrives with both sides in hand, where a side can be present, absent, or unknown. It makes a decision per key and hands back an action with  the reason for it. It is generic over the `KeyStream` trait rather than over concrete walkers, which is what lets the S3 -> S3 `Copy` operation just take a pair of two S3 listings.
  
Will be in a later PR, and `SyncWalker` / `SyncWalkerBuilder` / `SyncWalk` is the right seam and mirroring `FsWalker::walk` is worth it. I'll most likely build around that pattern, once comparison and execution are settled underneath.

On concurrency, I believe the merge stays sequential, since reading "absent from the other side" from position only works while both sides advance in one order. But to keep the merge off the critical path, merge decision is synchronous and cannot await, so a comparison mode can never be what limits transfers in flight.

`try_claim_subtree` returning `None` is deliberate and I haven't figured out a different way: emitting in key order means descending into a subtree before the sibling that sorts after it, so by the time entries come out there is little left to hand off. Serial pagination is the real constraint on the listing side. I have not measured a walk against a listing in one run, so that part is to be measured.

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
```diff
@@ -63,22 +190,67 @@ impl DirEntry {
...
+// Consulted with a path relative to the root, before any metadata is read.
+type PathFilterFn = Arc<dyn Fn(&Path) -> bool + Send + Sync>;
 
-/// Configuration for walking a local filesystem directory.
+/// How a walk orders the entries it produces.
 ///
+/// The default is [`Native`](SortOrder::Native).
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum SortOrder {
```
aajtodd, line 205: Why 3 variants here? Is WithinDirectory expected to be used?

aws-sdk-s3-transfer-manager/src/io/walk/error.rs
aajtodd: Ya maybe revisit in future PR, doesn't need to block this one.

aajtodd (approved):
Nothing blocking.

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
ysaito1001: Reads better, udpated in fe6fcf6.

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
ysaito1001: The previous boolean mapped to: `sort(false)` -> `Native` and `sort(true)` -> `WithinDirectory`. Turns out `sort(true)` was only used in tests for `try_claim_subtree` (making the test verification deterministic). `WholeWalk` is the new one that satisfies needs for merge stream.

143f141 gated `WithinDirectory` behind `test-util`.

aws-sdk-s3-transfer-manager/src/io/walk/error.rs
ysaito1001: `pub(crate)` sounds good, it's in 6624a1e

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
ysaito1001: Moved in 197b3c8

aws-sdk-s3-transfer-manager/src/io/walk/fs.rs
aajtodd: Ok as an FYI the previous boolean / sort(true) was meant to be used/address this functionality so we may not need to retain it's behavior if it was always wrong.

PR #178, opened 2026-09-16
buffer pool: hardening and optimization pass

aajtodd:
## Summary

This PR closes correctness and release qualification for the shared buffer pool after upload integration in [#176](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/176). It makes configured capacity a strict reservation ceiling, checks ownership and accounting through an independent generated-operation model, hardens native mapping and positioned file-I/O boundaries, and makes the supported platform, sanitizer, and fuzzing matrix required CI.

It also selects one allocator placement policy. Claims containing complete bitmap words first seek one contiguous run, then retain the existing fragmented fallback. Admission and physical preparation remain unchanged; the policy does not reserve memory headroom.

This is the hardening and qualification layer in the buffer-pool series:

- **This PR:** correctness hardening, file integrity, allocator placement, and platform qualification

The authoritative memory, ownership, and platform contracts are documented in `docs/design/memory.md`.

## Why

The preceding integration PRs put download and upload staging under one memory lifecycle. A reservation represents planned work, acquired carriers represent physical ownership, and mutable or immutable byte owners may retain those carriers after the reservation closes. The pool therefore has to preserve its capacity and accounting contracts across queueing, cancellation, close, retry, publication, partial consumption, and final owner release.

The existing admission exception allowed an otherwise idle reservation to exceed configured capacity. That made the configured value a pressure target rather than a reservation ceiling and allowed one request to reserve an envelope the pool could never admit under ordinary contention. This PR rejects impossible envelopes before FIFO insertion or physical preparation. Explicit unreserved acquisition remains the caller-controlled path for work that intentionally operates outside reservation envelopes.

The existing tests covered individual operations and bounded concurrency, but they did not generate enough compositions of aliases, queued demand, reservation close, growth, partial consumption, and final owner return. Native mapping faults also sit outside Miri and Loom's execution models. This PR adds one typed operation vocabulary shared by property tests, checked-in regression replay, and libFuzzer; compares each operation with an independent ownership model; audits the complete pool after every step; and adds native guard-page probes around test block reservations.

The transfer boundary had separate integrity gaps. Positioned reads and writes did not apply one exact interruption, progress, and offset contract across Unix and Windows. A successful download into a caller-provided file could also leave a stale pre-existing tail. This PR makes exact transfer progress explicit and moves destination resizing into successful finalization.

Finally, whole-part churn returned complete bitmap-word extents and the general lowest-fit claimant immediately split them again. Overlapping 8 MiB and 16 MiB replacements therefore became segmented even though the returned extent already had the desired shape. The selected placement policy recycles complete word runs before falling back to arbitrary free carriers.

### Admission and retained ownership

Reservation eligibility is now:

```text
active planned demand + uncovered charges + requested envelope
    <= configured capacity
```

An envelope larger than configured capacity returns `ReserveError::ExceedsCapacity` before it can enter the FIFO. Grant publication still waits until physical preparation covers the complete post-grant admission floor. `BufferPool::acquire_unreserved` remains separate and may create an uncovered charge above the normal reservation ceiling.

One reservation and its owners move through this lifecycle:

```text
grant reservation
  active planned demand += envelope
  available coverage += envelope
                  |
                  v
          acquire carriers
  available coverage -= acquired
  live covered ownership += acquired
                  |
          +-------+-------+
          |               |
          v               v
drop while open      close reservation
return bitmap        remove envelope
restore coverage     covered ownership becomes uncovered
                          |
                          v
                   final owner drop
                   return bitmap first
                   retire direct authority
                   remove aggregate charge last
                          |
                          v
                   reconsider FIFO
```

Closing a reservation prevents new direct acquisition but does not invalidate existing owners. Coverage close is derived from conserved live charges and remaining active demand, so closing one reservation cannot consume coverage required by another. Physical carrier availability is published before accounting pressure is reduced; a newly eligible waiter is woken only after pool locks are released.

### Generated ownership validation

Property tests, checked-in corpus replay, and libFuzzer exercise the same typed operation semantics:

```text
proptest strategies                    corpus or libFuzzer bytes
        |                                         |
        |                                fixed-width decoder
        |                                         |
        +-----------> typed Operation sequence <--+
                              |
                       sequence runner
                     /                 \
                    v                   v
        independent PoolModel     production BufferPool
        expected result/state             |
                    |             private whole-pool audit
                    |                     |
                    +--------- compare ---+
                              |
                  repeat after every operation
                              |
                    teardown: quiescent zero
```

The model owns plain reservation, coverage, byte, alias, queue, and owner-lifetime state. It does not read pool internals to compute expected outcomes. The production side uses public operations, then reconstructs physical bitmap ownership, prepared capacity, accounting, queue state, and cleanup state through the private audit. Every generated sequence compares operation results and complete audited state, then requires all authority and ownership to return to zero.

The fixed-width decoder covers reservation, reserved and unreserved acquisition, growth, publication, freeze, clone, slice, cursor advance, cancellation, and drop. Checked-in seeds use semantic length selectors so their intended transitions remain meaningful across supported page sizes. Ordinary package tests replay promoted seeds under the root lock; cargo-fuzz writes generated inputs to its own ignored corpus.

The sequential runner cannot create an interleaving inside one production operation. Exhaustive coverage-state tests and production-path Loom models therefore retain responsibility for close racing carrier return, immutable cursor movement or reads racing sibling drop, batched return racing trim, and one-time final return. Strict-provenance Miri checks pointer derivation, initialization, publication, and aliasing on the deterministic paths.

### Native mapping boundaries

Test-only block reservations retain an inaccessible page before and after the complete managed range. A positive lifecycle test proves ordinary pooled storage remains usable. Child-process probes then write one byte before and after a carrier range, prove the target address lies inside the guarded reservation, and require a native memory fault or an anchored sanitizer report.

The guard mechanism is absent from normal construction. Production block layout, preparation, discard, and reclamation remain unchanged.

### File transfer integrity

Positioned I/O now uses explicit exact-progress loops. Interrupted operations retry, zero-progress writes return `WriteZero`, partial reads expose only the initialized prefix before reporting `UnexpectedEof`, and Unix vectored writes reject offsets that cannot be represented by the signed `pwritev` interface. Upload file bodies publish no partial frame when an exact read fails.

Download sinks separate three operations. `prepare` reserves manager-owned storage where the platform supports it, `terminal_drain` releases queued payload after failure or cancellation without claiming success, and `finalize` applies success-only destination state. Caller-provided files are not eagerly resized; successful finalization sets the file to `expected_download_len`, so a larger pre-existing file cannot retain a stale tail.

`write_to_file` remains replace-from-zero. This PR does not define append, nonzero-offset, preserve-tail, `fsync`, or power-loss durability behavior.

### Whole-part placement

The default 16 KiB carrier makes an 8 MiB part exactly eight bitmap words and a 16 MiB part exactly sixteen. Whole-value return can restore those words as complete free extents. Eligible claims now preserve that shape:

```text
whole-word claim
      |
      v
bounded optimistic run scan
      | hit                         | miss: publish no partial claim
      v                             v
one complete run          serialized all-slot run scan
                                   | hit                 | miss
                                   v                     v
                           one complete run      existing fragmented
                                                 fallback and growth
```

`BlockSlot::try_claim_word_run` wins each word with compare-and-exchange. The attempt publishes ownership only after winning the complete run; a collision drops the attempt and rolls back every word already won. Requests that are not whole-word sized, exceed one block, or cannot find a run retain the general allocator.

The optimistic cursor keeps its fixed-budget stride so concurrent claimers receive disjoint search windows. Serialized run search begins from the cursor-derived slot instead of always preferring slot zero. Admission, configured capacity, and preparation remain unchanged; no eager headroom is added.

At four prepared and live blocks, the churn fixture remains at 100% occupancy and reports 100% contiguous replacement for 8 MiB and 16 MiB parts in every released-first and overlapping, ordered and shuffled cell. Dedicated arena and block tests cover preference and fallback, and a Loom model covers collision after a partial run win. The compact model/property/fuzz geometry cannot reach a 64-carrier run, so larger-geometry sequence coverage remains separate qualification work.

### CI and platform qualification

Required CI groups the transfer-manager support contract under one `check-transfer-manager` result:

| Child check | Required evidence |
| --- | --- |
| `default-targets` | Full workspace and doctests on GNU Linux x86_64/arm64, Windows x86_64 MSVC, and macOS x86_64/arm64 |
| `musl` | Full workspace and doctests in native Alpine containers on Linux x86_64/arm64 |
| `freebsd` | Transfer-manager library tests in the native FreeBSD 14.4 VM |
| `android` | API 35 arm64 test-binary linking and x86_64 library execution on asserted 4 KiB and 16 KiB emulator page sizes |
| `e2e` | Real-S3 transfer tests on Linux, Windows, and macOS, with credentials isolated to this child |

Android arm64 remains build-qualified rather than runtime-qualified. Scheduled workflows remain responsible for long fuzz campaigns and destructive pressure or crash testing; ordinary supported-platform evidence runs on each PR.

The required fuzz check captures and validates a nonempty target list, replays promoted seeds, builds every target, runs a deterministic libFuzzer smoke campaign, and uploads crash artifacts. The scheduled workflow restores an evolving writable corpus for longer exploration. ASan, TSan, Miri, Loom, release tests, and native platform tests remain distinct because they cover different failure classes.

The dependency and toolchain update moves the workspace to the Rust 1.94.1 SDK line and updates the pinned nightly and external-type tooling used by these gates. Normal builds and rustdoc do not expose the fuzz entry point or depend on libFuzzer.

## How to review

1. Start with reservation eligibility and retained ownership in `admission.rs`, `admission/coverage.rs`, `admission/waiter.rs`, and `acquisition.rs`. Follow one envelope through grant, carrier acquisition, reservation close while owners remain live, final physical return, aggregate repayment, and FIFO reconsideration.
2. Read `test_util/operation_sequence.rs`, `test_util/model_harness.rs`, and `test_util/audit.rs` as one validation mechanism. The decoder defines the operation vocabulary, the model computes expected state independently, and the audit reconstructs actual pool state after each operation. Then inspect `tests/prop.rs`, `tests/fuzz_replay.rs`, and the promoted corpus as input sources for that mechanism.
3. Review native memory safety through `virtual_memory.rs`, `block.rs`, and the guarded-buffer tests in `pooled_buf.rs`. The important boundary is that guard probes prove their faulting address belongs to the reservation and cannot pass through an unrelated process failure.
4. Review positioned I/O in `io/fs.rs`, upload publication in `operation/upload/file_body.rs`, and download sink transitions in `operation/download/body.rs`, `transfer.rs`, `handle.rs`, and `builders.rs`. Separate preparation, terminal drain, and successful finalization while checking that partial progress never becomes a complete frame or successful file result.
5. Review placement through `Arena::claim_optimistic`, `Arena::complete_claim_serialized`, `BlockSlot::try_claim_word_run`, and `BlockClaimAttempt::take_word_run`. A miss must publish no partial run, collision must roll back every word, and general fallback must retain admission progress.
6. Finish with `.github/scripts/run-fuzz-targets.sh`, the required fuzz and sanitizer jobs, the transfer-manager platform children, E2E credential scope, and required-check aggregation. Target discovery and job aggregation must fail closed rather than report success after executing no target or omitting a required child.

## Future Work

The committed placement policy still needs production qualification across mixed whole-part and tail traffic, repeated cohorts, capacity scales, larger-geometry generated sequences, and end-to-end upload plus reused-pool download. Preparation headroom, freed-extent handoff, and broader search indexes remain evidence-driven alternatives rather than required follow-ups.

Before GA, caller-provided download destinations need an explicit layout and durability decision. Replacement may remain the only mode, or the API may add append and nonzero-offset writes with defined tail handling, concurrent-mutation behavior, offset overflow, append-mode handle behavior, and optional durability.

Remaining performance work covers download drain serialization and receive-path copies, upload checksum and request-body architecture, direct file I/O, small-object startup, and diagnostics overhead. Final series cleanup owns module and terminology reconciliation, temporary allowances and probes, private-rustdoc debt, blanket dead-code suppression, and stacked replay or collapse.


aajtodd:
Closing in favor of splitting this into multiple PRs:

* https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/179
* https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/180
* https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/181

PR #179, opened 2026-09-19
buffer pool: harden admission and ownership accounting

aajtodd:
## Summary

This PR hardens the shared buffer pool after its download and upload integrations. It makes configured capacity a strict reservation ceiling, preserves accounting while reservations close with live byte owners, and restores the original coverage classification when an acquisition fails before publishing ownership.

It also adds the verification infrastructure used to qualify those contracts: generated operation sequences checked against an independent model, checked-in fuzz regression replay, libFuzzer targets, production-path Loom models, strict-provenance Miri coverage, and native guard-page probes.

This branch starts from `bp-main` after upload integration in #176.

This is the correctness and verification layer in the buffer-pool series:

- **This PR:** admission, ownership, and verification hardening
- Next: native file integrity and platform qualification
- Final: contiguous whole-part placement

The authoritative memory and ownership contracts are documented in `docs/design/memory.md`.

## Why

A reservation represents planned demand, while acquired carriers represent physical ownership. Mutable and immutable byte owners can retain those carriers after the reservation closes. Admission therefore cannot treat an idle queue or a closed reservation as proof that retained memory is available.

The previous idle-only exception could admit a reservation whose envelope exceeded configured capacity, or admit new planned demand while earlier owners still retained uncovered charges. Repeated close-retain-reserve cycles could consequently exceed the configured ceiling without a bounded limit.

Failure rollback had a separate classification problem. An acquisition provisionally debits coverage before claiming physical carriers. Reusing the normal return transition when that claim failed could repay unrelated uncovered ownership and change admission classification even though the failed operation published no buffer.

Focused unit tests covered individual transitions, but they did not compose reservation close, queued demand, unreserved acquisition, growth, mutable publication, immutable aliases, partial consumption, cancellation, and final owner return in generated sequences. Loom and Miri also cannot prove that a native mapping faults at the intended page boundary.

## Admission and retained ownership

Normal reservation eligibility is now:

```text
active planned demand + retained uncovered charges + requested envelope
    <= configured capacity
```

An oversized envelope returns `ReserveError::ExceedsCapacity` before entering the FIFO or triggering physical preparation. Explicit unreserved acquisition remains available for work that intentionally operates without a reservation envelope, but that ownership stays charged and delays later managed work.

One reservation and its owners move through these accounting states:

```text
grant reservation
        |
        v
available coverage ---- acquire ----> live carrier charge
        |                                      |
        |                               close reservation
        |                                      |
        |                           retained charge may become
        |                                  uncovered
        |                                      |
        +<----------------------------- final owner drop
                                       return bitmap
                                       retire authority
                                       retire charge
```

Closing a reservation prevents new direct acquisition but does not invalidate existing owners. Carrier availability is published before accounting pressure is reduced, and newly eligible waiters are woken only after pool locks are released.

A failed acquisition carries the contribution made by its provisional debit until ownership conversion succeeds. Rollback restores that exact covered or uncovered contribution while accounting for a reservation that may have closed concurrently.

## Property tests and fuzzing

Property tests, checked-in corpus replay, and libFuzzer use one typed operation vocabulary and one sequence runner:

```text
proptest strategies                    corpus or libFuzzer bytes
        |                                         |
        |                                fixed-width decoder
        |                                         |
        +-----------> typed Operation sequence <--+
                              |
                       sequence runner
                     /                 \
                    v                   v
        independent PoolModel     production BufferPool
        expected result/state             |
                    |             private whole-pool audit
                    |                     |
                    +--------- compare ---+
                              |
                  repeat after every operation
                              |
                    teardown: quiescent zero
```

The operation set covers reservation creation and close, reserved and unreserved acquisition, buffer growth, initialized writes, prefix publication, freeze, clone, slice, cursor advancement, cancellation, and drop. The model uses plain reservation, coverage, byte, alias, queue, and owner-lifetime state. It does not read production internals when calculating expected results.

After every operation, the runner compares the expected result with the real result and compares the complete model state with a private audit of reservation accounting, queue state, prepared capacity, live bitmap ownership, cleanup state, and retained owners. Sequence teardown drops all remaining handles and requires every source of authority and ownership to return to zero.

Checked-in seeds provide stable regression cases under ordinary package tests and Miri. Semantic length selectors keep their intended transitions meaningful across supported page sizes. libFuzzer uses the same decoder and runner, writes generated inputs to an ignored working corpus, and promotes only reviewed reproductions into package-owned seeds.

The fuzz entry point and its dependencies are compiled only under the private `s3_tm_fuzz` configuration. They are absent from normal builds and rustdoc.

## Concurrency and pointer safety

The sequence runner is deliberately sequential. It cannot create an interleaving inside one production operation, so concurrency obligations remain with production-path Loom models.

The new Loom coverage races immutable cursor movement and reads against sibling-owner drop, batched carrier publication against trim, and coverage-restoring return against reservation close. Exhaustive coverage-state tests pin arithmetic branches that require a return between production snapshots and compare-and-exchange.

Strict-provenance Miri tests cover pointer derivation through the live virtual-range owner, mutable initialization, prefix publication, disjoint retained suffix writes, immutable aliases, and final owner return. Publication exposes only ranges reported initialized.

## Native guard pages

Test-only block reservations retain one inaccessible page before and after the complete managed carrier range:

```text
+------------+-------------------------------+------------+
| guard page | managed carriers              | guard page |
| no access  | normal pooled storage         | no access  |
+------------+-------------------------------+------------+
```

A positive lifecycle test first proves that ordinary reads, writes, publication, and return still work inside the managed range. Child-process probes then target one byte immediately before and after that range. Before writing, the child proves that the target address lies inside the guarded reservation, emits a readiness marker, and must terminate through a native memory fault or an anchored ASan deadly-signal report. A panic, unrelated sanitizer report, missing guard reservation, or surviving write fails the parent test.

The guard mechanism exists only in test construction. Production reservation geometry, preparation, discard, and reclamation are unchanged.

## CI qualification in this PR

Required CI replays the checked-in corpus, verifies that target discovery is nonempty, builds every fuzz target, runs deterministic libFuzzer smoke campaigns, and uploads crash artifacts. A scheduled workflow restores an evolving writable corpus for longer exploration.

ASan, TSan, Miri, Loom, release tests, and ordinary package tests remain separate because they cover different failure classes. The workspace MSRV, SDK dependencies, pinned nightly, and external-types checker move together so every required shape resolves against the supported dependency graph.

## How to review

1. Start with reservation eligibility and coverage transitions in `admission.rs`, `admission/coverage.rs`, `admission/waiter.rs`, and `acquisition.rs`.
2. Follow one owner through grant, acquisition, reservation close while ownership remains live, final bitmap return, authority retirement, aggregate repayment, and FIFO reconsideration.
3. Read `test_util/operation_sequence.rs`, `test_util/model_harness.rs`, and `test_util/audit.rs` as one verification mechanism.
4. Review the deterministic property cases and checked-in replay seeds before reviewing the fuzz wrapper.
5. Review immutable-owner concurrency in `segmented_bytes.rs` and the coverage return-versus-close Loom models.
6. Finish with the guarded virtual-memory construction and child-process probes in `virtual_memory.rs`, `block.rs`, and `pooled_buf.rs`.

## Scope

This PR does not select allocator placement policy, change download destination semantics, or expand the supported-platform matrix. Those are owned by the next two PRs in this replacement stack.


aws-sdk-s3-transfer-manager/src/runtime/buffer_pool/admission.rs
```diff
@@ -438,18 +448,20 @@ impl AdmissionLedger {
...
             .ok()
             .and_then(|used| used.checked_add(envelope))
-            .is_some_and(|next| next <= self.configured_capacity);
-        normal || self.active_planned_demand == CarrierCount::ZERO
+            .is_some_and(|next| next <= self.configured_capacity)
+    }
+
+    /// Rejects an envelope that can never fit under normal admission.
+    fn validate_envelope(&self, envelope: CarrierCount) -> Result<(), ReserveError> {
+        if envelope > self.configured_capacity {
```
ysaito1001: The reject is correct and intended. The potential issue is that part size against the memory budget isn't validated at construction time, so a misconfiguration surfaces as a per-transfer runtime error, e.g. capacity is floored at pool construction:
```  
BufferPool::builder().capacity(MemoryBudgetConfig::Limit(n))
└─ config.rs:143   PoolConfig::resolve_for_page
   └─ config.rs:154   let carriers = bytes / carrier_size;        ◄── FLOOR
        └─ configured_capacity
```
Envelope is ceiled on every part:
```
io/adapters.rs:80        part_buffer.poll_acquire(cx, stream_cx.part_size())
└─ io/part_buffer.rs:128   if envelope == 0 || envelope > remaining      → passes
   └─ io/part_buffer.rs:131   self.start_reservation(envelope)
      └─ io/part_buffer.rs:208   fn start_reservation
         └─ io/part_buffer.rs:211   self.pool.reserve(envelope_remaining)
            └─ buffer_pool.rs:186     BufferPool::reserve → ReserveFuture::new
               └─ buffer_pool.rs:250     reservation_envelope(bytes)
                  └─ geometry.rs:205       bytes.div_ceil(carrier_size)   ◄── CEIL
               └─ buffer_pool.rs:468     PoolInner::reserve_or_enqueue
                  └─ buffer_pool.rs:485     admission.validate_envelope(envelope)?
                     └─ admission.rs:366      AdmissionGuard::validate_envelope
                        └─ admission.rs:460      if envelope > configured_capacity
                                                    → Err(ReserveError::ExceedsCapacity)
```
And unwinds into a terminal transfer error:
```
                        ◄─ admission.rs:461      Err(ExceedsCapacity)
      ◄─ io/part_buffer.rs:142   Poll::Ready(Err(io::Error::other(error)))
◄─ io/adapters.rs:83-86    *this.buffer = None
                           Poll::Ready(Some(Err(error)))
   └─ part stream yields Err → transfer fails
```
`try_reserve` reaches the same reject via `buffer_pool.rs:443 → :455`.
  
Concrete failures, carrier = 16 KiB on both 4 KiB and 16 KiB page hosts, defaults 8 MiB upload / 5 MiB download (client.rs:80,88):
```
Limit(4 MiB)      cap = 4194304/16384  = 256   env = ceil(8388608/16384)  = 512   → FAILS
Limit(5 MiB)      cap = 5242880/16384  = 320   env = ceil(8388608/16384)  = 512   → FAILS
Limit(8 MiB)      cap = 8388608/16384  = 512   env = ceil(8388608/16384)  = 512   → ok
Limit(10_000_000) cap = 10000000/16384 = 610   env = ceil(10000000/16384) = 611   → FAILS
```  
That last line is nominally equal budget and part size but still fails, purely from floor-vs-ceil.
  
This worked at bp-main, proven by the test deleted in this PR; `test_pool(block_carriers, configured)`, so a one-carrier pool granting a three-carrier envelope:
```
fn test_idle_only_grant_can_exceed_configured_capacity() {
    let (pool, carrier_size) = test_pool(2, 1);
    let reservation = pool.try_reserve(carrier_size * 3).unwrap().unwrap();
```  
types.rs:24-25 still promises the old behaviour: "At the limit transfers backpressure rather than fail."
  
  Fix: validate at client/pool build time and return a configuration error, or clamp effective part size — comparing in carriers, not bytes. Then update the MemoryBudgetConfig rustdoc.

ysaito1001 (approved):
The hardening is the right move and the accounting work is done carefully. One potential bug worth looking into before merge.

PR #181, opened 2026-09-19
buffer pool: prefer contiguous whole-part placement

aajtodd:
## Summary

This PR adds best-effort contiguous placement for claims containing whole bitmap words. The arena prefers one complete run when available and retains fragmented allocation as the guaranteed fallback.

Whole-part traffic commonly returns complete free bitmap-word extents. The existing lowest-fit claimant immediately split those extents under overlapping churn, producing segmented buffers even when a correctly sized run was already available. Eligible claims now seek one complete run before falling back to the existing fragmented allocator.

The claim-order change affects physical placement only. It does not change admission, configured capacity, preparation, reclamation, or the guarantee that fragmented free storage remains usable.

This is the final PR in the three-part replacement stack for aggregate draft [#178](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/178).

This branch is stacked on [#180](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/180); its diff contains only the allocator claim-order change and its qualification.

This is the placement and final qualification layer in the buffer-pool series:

- [#179](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/179): admission, ownership, and verification hardening
- [#180](https://github.com/awslabs/aws-s3-transfer-manager-rs/pull/180): native file integrity and platform qualification
- **This PR:** prefer contiguous runs for whole-part claims and complete final qualification

## Why

The default 16 KiB carrier makes one 64-bit bitmap word represent 1 MiB of storage. An 8 MiB upload or download part occupies eight complete words, and a 16 MiB part occupies sixteen.

Returning a whole immutable value can restore those complete words as one reusable physical extent. The general prefix claimant instead selected the lowest free carriers and split the extent again. Under overlapping replacement, the pool therefore returned segmented values even though a complete part-sized run existed.

Segmentation is correct, but it increases presentation segments and downstream vectored-I/O work. The allocator should prefer an already available whole-part shape without spending additional memory or preventing fragmented fallback.

## Best-effort contiguous placement

For eligible whole-word claims, the buffer pool begins with a bounded optimistic search for one complete run. A miss enters serialized search, and the request still succeeds through fragmented fallback when no complete run exists:

```text
whole-word claim
        |
        v
bounded optimistic run scan
        | hit                         | miss
        v                             v
claim complete run          serialized complete run scan
                                       | hit        | miss
                                       v            v
                               claim complete run   existing fragmented
                                                    fallback and growth
```

A request is eligible only when its carrier count is a complete bitmap-word multiple and it fits within one block. Smaller requests, cross-block requests, and arbitrary tail sizes continue directly to the general allocator.

The optimistic pass retains the configured scan budget. A miss publishes no partial run. Serialized completion checks the complete registry for a run and then falls back to the existing prefix claimant before considering growth.

## All-or-nothing word ownership

`BlockClaimAttempt::take_word_run` wins each complete free word with compare-and-exchange. Loads used during search are only a filter; the compare-and-exchange is authoritative.

Ownership is published only after every word in the run has been won. If another claimant wins a later word first, dropping the attempt rolls back every word already acquired. A Loom model forces that partial-win collision and proves that one claimant succeeds while the loser returns all provisional ownership.

The trailing partial bitmap word is never eligible as a complete run. Fragmented fallback retains its existing padding-bit and incarnation checks.

## Search distribution

Optimistic claimants advance the arena cursor by one charged search window so concurrent callers start in different regions. When the configured budget covers the complete search space, advancing by one full budget would make a complete lap and return every caller to the same word. Full-space scans now advance by one word instead, while partial scans retain budget-sized strides.

Serialized run search also starts from the cursor-derived slot instead of always preferring slot zero.

## Large-geometry property tests and fuzzing

The compact accounting model uses only a few carriers and cannot reach a 64-carrier bitmap-word claim. This PR adds a second sequence profile with block and pool geometry large enough to exercise production placement while retaining the same typed operation decoder, reference model, and whole-pool audit.

The placement profile records whether an eligible acquisition began with a pre-existing free run, whether the returned value was contiguous or segmented, and whether a partial-word or cross-word request actually executed. Deterministic tests and checked-in seeds assert those milestones so a geometry or decoder change cannot silently turn a placement case into a no-op.

The checked-in `partial-word-before-part` seed covers a small owner followed by reusable whole-word claims. The `fragmented-fallback` seed proves that an eligible request remains segmented when no complete run exists, preventing the free-run oracle from passing vacuously.

## Mixed-size qualification

Homogeneous churn replaces 8 MiB and 16 MiB cohorts under released-first and overlapping, ordered and shuffled schedules. At four prepared and live blocks, every measured cell reports 100 percent contiguous replacement at full occupancy.

A second fixture keeps three small owners live while repeatedly replacing normal part-sized claims. The selected sizes damage three bitmap words and leave exactly three physical free extents for the final part, exercising the interaction between small long-lived claims and whole-part traffic. Local runs observe approximately 93–98 percent contiguous replacement for 8 MiB parts and 90–95 percent for 16 MiB parts. These are qualification baselines rather than CI percentage thresholds; per-cell ordering is not stable.

The benchmark preflight reports prepared and charged occupancy, contiguous replacement, physical runs, part count, and unallocated tail bytes. It requires at least 95 percent charged occupancy so fixture truncation cannot create misleading placement headroom.

## No preparation headroom

The implementation does not prepare an extra block in anticipation of demand. Preferential whole-word claiming alone kept homogeneous replacement contiguous at full occupancy. Optional headroom would add committed capacity on Windows, a fallible preparation path on grant, and slack to the prepared-capacity invariant without improving the measured result.

Preparation headroom, direct freed-extent handoff, and broader free-run indexes remain evidence-driven alternatives if production measurements show a placement or demand-spike problem.

## How to review

1. Start with `Arena::claim_optimistic` and `Arena::complete_claim_serialized` to verify the search order and fallback boundary.
2. Review `BlockClaimAttempt::take_word_run` and `BlockSlot::try_claim_word_run` for complete-word eligibility, compare-and-exchange ownership, and rollback.
3. Review cursor advancement separately; full-budget scans must rotate without changing partial-budget search reach.
4. Read the large-geometry profile in `operation_sequence.rs` with the placement milestones in `model_harness.rs`.
5. Review deterministic placement tests and checked-in replay seeds before the benchmark preflight.
6. Confirm that requests outside the whole-word fast path retain the existing fragmented allocator and growth behavior.

## Scope

This PR does not add placement configuration or expose a topology policy. The arena unconditionally tries a contiguous run for an eligible request before using fragmented fallback, which preserves progress and utilization.


docs/design/memory.md
ysaito1001: I believe the doc still documents the the old acquisition path, i.e. 
[3. Keep every bit won from each atomic bitmap operation.](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/ab582e3228920688983b86999de428d44ebe40ed/docs/design/memory.md#carrier-claiming-and-fallback)?

- `take_word_run` rolls back partial wins and the contiguous branch is all-or-nothing. The steps in the link above don't mention the new rotating-start run pre-phase.
-  the eligibility rule, `contiguous_word_count` (nonzero, exact multiple of 64 carriers, within one block) splits allocator behaviour into two paths, but there seems no mention in the doc.

Could we update the above somewhere in the doc for consistency?

aws-sdk-s3-transfer-manager/src/runtime/buffer_pool/arena.rs
```diff
@@ -325,7 +339,36 @@ impl Arena {
...
+                    invariant_violation("contiguous block claim reported invalid scan work");
+                }
+                batch.inspected_words += inspected;
+                if let Some(provisional) = claim.into_provisional() {
+                    batch.push(provisional)?;
+                }
+            }
+            self.diagnostics
+                .record_optimistic_scan(batch.inspected_words, !batch.is_complete());
+            return Ok(batch);
```
ysaito1001: A question about eligible requests now potentially doing more work when no run exists
  
This returns before the bounded `try_claim_words` loop, so an eligible request that misses the run scan comes back empty. At the base branch, everything fell through and claimed whatever it found lock-free so for eligible sizes the optimistic pass went from partial progress to all-or-nothing.
  
When a complete run exists, that's the clear win. When it doesn't, each acquisition potentially scans ~32 words (in vain), takes the mutex, sweeps every slot across its full bitmap with a Vec alloc per slot (:415-434), then runs the serialized `try_claim_exhaustive` scan at :436 for the full request instead of the remainder. An 8 MiB part is eligible by default, so that's the common path once the pool is fragmented, which is motivation behind the PR.
  
Separately, `optimistic_scan_words` is 32 by default, so any eligible request over 32 MiB doesn't hit the run scan, but still pays the full budget allocating per window on the way (skipping the branch when budget < word_count would cover it).
  
Maybe it's worth profiling, outside the PR, if we haven't done so?

ysaito1001 (approved):
LGTM, a few minor comments.

PR #182, opened 2026-09-22
per-object transfer events on all four operations

vcjana:
## Summary
Adds per-object lifecycle events to `upload`, `download`, `upload_objects` and `download_objects`. Right now the only way to watch a transfer is to poll `metrics()` in a loop; this lets a caller be told instead.

Lifecycle is pushed, quantities are pulled. `Decided` and `Settled` go out on a bounded channel that drops under backpressure rather than stalling the transfer, and each `Decided` carries a read-only [`TransferView`](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/0d5ad2732815367729625b33efa455f54e0231a6/aws-sdk-s3-transfer-manager/src/types.rs) you read the counters off. That's the part worth arguing about: a total you accumulate from a lossy stream is wrong forever once one event drops, but a value you re-read heals itself. `dropped()` is how a consumer finds out its own stream lost something.

There's no callback module — a test in [`download_objects.rs`](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/0d5ad2732815367729625b33efa455f54e0231a6/integration-tests/src/download_objects.rs) rebuilds callback-style reporting from the stream in one function. Sinks registered at client level and request level merge instead of the second one winning. `Direction::Copy`, `Decision::Delete` and `Decision::Skip` are `doc(hidden)` until `sync` produces them, since these names freeze at 1.0.

Two numbers are weaker than they look and say so in their docs. `bytes_streamed` only goes up; I tried add-on-read/subtract-on-confirm and it leaks, because a hedged attempt gets dropped mid-read so no release path ever runs. `StallReason` is downloads-only — an upload's park sites aren't labelled, so it reports `None` instead of guessing.

Behaviour changes in two places:
- `download_objects` now drops 0-byte folder markers no matter which walker found them. Before, only the default walker filtered them, so supplying your own walker or filter let them back in and the transfer died: `sub/` derives the local path `<dest>/sub`, which is the directory `sub/b.bin` already made, so the write fails with `EISDIR` and `Abort` takes the whole operation down. Verified on a real bucket, exit 1 with four of five objects on disk.
- [`S3Walker`](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/0d5ad2732815367729625b33efa455f54e0231a6/aws-sdk-s3-transfer-manager/src/io/walk/s3.rs) doesn't filter anything on its own now, so `filter` means what it says.

## Review order
1. [`src/events.rs`](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/0d5ad2732815367729625b33efa455f54e0231a6/aws-sdk-s3-transfer-manager/src/events.rs) — the events, the view, the channel. Two events instead of four callbacks; everything a progress callback would have carried is read off the view.
2. Any one of the four `builders.rs` — [`download_objects`](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/0d5ad2732815367729625b33efa455f54e0231a6/aws-sdk-s3-transfer-manager/src/operation/download_objects/builders.rs) will do — plus `resolve_sink`. Same shape in all four.
3. `claim_one` and `announce_child` in [`download_objects/transfer.rs`](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/0d5ad2732815367729625b33efa455f54e0231a6/aws-sdk-s3-transfer-manager/src/operation/download_objects/transfer.rs), then the same pair in [`upload_objects`](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/0d5ad2732815367729625b33efa455f54e0231a6/aws-sdk-s3-transfer-manager/src/operation/upload_objects/transfer.rs). This is the only subtle part. `on_terminal` isn't gated on `children_reserved == 0` the way `check_terminal` is, so the root can be taken while an entry is off `pending_entries` and not yet in `child_lifecycles` — neither sweep reaches it there. So the sink is cloned inside `claim_one`, under the same lock that pops the entry.
4. The `record_entry_settled` calls, six in `download_objects` and seven in `upload_objects`. An entry whose event was lost is still counted, which is what keeps a progress bar from sticking below 100% forever.

## Testing
- The claim-to-announce window is pinned deterministically — racing it under load only reproduced about one run in five. Both window tests are mutation-verified, in both composites.
- fmt, clippy `-Dwarnings` on all targets, the workspace suite three times over, 41 loom tests. miri and check-external-types run on Linux in CI.


aws-sdk-s3-transfer-manager/src/operation/download/builders.rs
```diff
@@ -42,11 +68,22 @@ impl DownloadFluentBuilder {
...
     ) -> Result<ManagedDownloadHandle, crate::error::Error> {
         let input = self.inner.build()?;
+        let path = path.into();
+        let events =
+            crate::events::resolve_sink(self.handle.config.events(), self.events).map(|sink| {
+                crate::operation::download::EventRegistration {
+                    sink,
+                    destination: crate::events::Endpoint::Local {
+                        path: std::sync::Arc::from(path.as_path()),
+                    },
```
ysaito1001: `write_to_path` doesn't download straight to the path the caller asked for. It writes to a temporary file alongside it and then renames that file into place, so a reader never sees a half-written file at the destination without rename.

This sets the event's destination to the caller's path, but the `Settled { Succeeded }` event carrying it is emitted from the transfer's status change; `on_terminal` maps `TransferStatus::Completed` -> `Outcome::Succeeded {}` (`download/transfer.rs:1162`), and the rename happens somewhere else entirely, in `join()` (`handle.rs:395` -> `finalize()` -> `tokio::fs::rename` at
  `:445`). There is a gap between the two, so the event announces a finished file at a path the rename has not created yet, which never will in the following cases:
- The rename fails. `join()` returns an error, the temp file is deleted, and `Settled { Succeeded }` naming the destination still stands. The caller's event stream says the download succeeded; nothing exists at that path, and the event simply cannot see the failure.
- The caller trusts the event and drops the handle instead of calling `join()`. `Drop` deletes the temp file unconditionally (`handle.rs:467-469`), so the rename never runs and a successful download is thrown away after being reported as succeeded.
  
`write_to_file` sidesteps this by reporting `Endpoint::Stream {}` instead of a path, and the directory-download path gets the order right; it joins each child at `download_objects/transfer.rs:1329` and only then claims the child's terminal event at `:1338`. `

Fix: emit the terminal after `finalize()` succeeds. If `on_terminal`'s placement makes that awkward, the fallback is to put the temp path in the event until the rename lands, so the event never states something false.

integration-tests/src/progress_chaos.rs
```diff
@@ -0,0 +1,888 @@
...
+        assert!(d.double_finished.is_empty(), "exactly-once terminal");
+        assert!(d.unfinished.is_empty(), "no announced transfer left owing");
+        assert!(
+            d.orphan_finished.is_empty(),
+            "no terminal without a decision"
+        );
+        assert_eq!(d.dropped, 0, "nothing lost at capacity 256");
+        // 18 children succeed + the root; 2 children fail.
+        assert_eq!(d.failed, 2, "two children failed");
+        assert_eq!(d.succeeded, FILE_COUNT - 2 + 1, "18 children plus the root");
```
ysaito1001: The test decides up front how many uploads should fail, then asserts that exactly that many did. It only controls which files are unsuccessful, but it doesn't have control over files failing forunrelated reasons.

`doomed_keys` injects a fault into every 100/percent-th of the 20 files, so 2 files at 10% and 4 at 15%. `d.failed` is not that number: it is a tally of the `Settled { Failed }` events the run actually produced, and a file nobody sabotaged can fail for its own reasons on a loaded machine. `assert_eq!(d.failed, 2)` then fails at 3 even though the code behaved correctly, and `assert_eq!(d.succeeded, FILE_COUNT - 2 + 1)` fails alongside it at 18.

 Fix: take both numbers from what the run reported `(objects_uploaded() /  failed_transfers())`, the way the corrected assertion now does. The `success_bytes` doc comment a few lines above still explains the old pattern-derived reasoning and should be updated with them.

aws-sdk-s3-transfer-manager/src/operation/download/builders.rs
ysaito1001: `write_to_path` doesn't download straight to the path the caller asked for. It writes to a temporary file alongside it and then renames that file into place, so a reader never sees a half-written file at the destination without rename.

This sets the event's destination to the caller's path, but the `Settled { Succeeded }` event carrying it is emitted from the transfer's status change; `on_terminal` maps `TransferStatus::Completed` -> `Outcome::Succeeded {}` (`download/transfer.rs:1162`), and the rename happens somewhere else entirely, in `join()` (`handle.rs:395` -> `finalize()` -> `tokio::fs::rename` at
  `:445`). There is a gap between the two, so the event announces a finished file at a path the rename has not created yet, which never will in the following cases:
- The rename fails. `join()` returns an error, the temp file is deleted, and `Settled { Succeeded }` naming the destination still stands. The caller's event stream says the download succeeded; nothing exists at that path, and the event simply cannot see the failure.
- The caller trusts the event and drops the handle instead of calling `join()`. `Drop` deletes the temp file unconditionally (`handle.rs:467-469`), so the rename never runs and a successful download is thrown away after being reported as succeeded.
  
`write_to_file` sidesteps this by reporting `Endpoint::Stream {}` instead of a path, and the directory-download path gets the order right; it joins each child at `download_objects/transfer.rs:1329` and only then claims the child's terminal event at `:1338`. `

Fix: emit the terminal after `finalize()` succeeds. If `on_terminal`'s placement makes that awkward, the fallback is to put the temp path in the event until the rename lands, so the event never states something false.

aws-sdk-s3-transfer-manager/src/events.rs
```diff
@@ -0,0 +1,1410 @@
...
+/// the source directory and the key prefix, not a file and a key — and every child carries
+/// `parent: Some(root_id)`.
+///
+/// So a consumer acting per entry — deleting each source once its upload succeeds — skips the
+/// `parent: None` event **only for the directory operations**, where acting on it would act on
+/// the whole tree. Skipping it unconditionally is wrong: it drops every single-file transfer,
+/// whose sole event is the one with no parent.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub enum TransferEvent {
```
ysaito1001: Two things:
1. Bike-shedding the variant names, since they'll be public items. I can see that `Decided` and `Settled` mean different things in the code, but when I read out loud they sound alike. Could we pick names that a first-time reader can tell apart without reading source code comments?

2. This is not an error type per se, but if a variant carries a payload, we should consider following [the actionable pattern](https://smithy-lang.github.io/smithy-rs/design/rfcs/rfc0022_error_context_and_compatibility.html#actionable-error-pattern) for extensibility by moving the payload fields into private fields on a struct.

